Issue #85, opened 2014-08-18
Regression: SerialManager is not working


AlexDaniel:
There is just no button to create serial port.

![screenshot](http://files.progarm.org/2014-08-18-053738_638x429_scrot.png)

Using version 1.23-git201407171232-2 on debian.
SerialManager plugin is enabled.

Issue #13, opened 2013-10-13
BlueZ 5 issues


cschramm:
- [x] Port audio device support (PulseAudioProfile)

- [x] Port serial device support (Services["serial"] in applet plugins)

- [x] Port services support in blueman-assistant

- [x] Port services support in RecentConns

- [ ] Connection to generic services (HID and audio) does only seem to work right after discovering / pairing; throws "Input/output error" (mouse) or "No such file or directory" (smartphone; not connectable?)

- [x] Custom passkey not working

- [x] No pairing notifications depending on device?!

- [x] Re-Enabling AuthAgent plugin tries to re-register GlobalAgent

PR #160, opened 2014-12-26
Services

cschramm:
This branch introduces service classes as an abstraction away from the BlueZ dbus interface (which is problematic for the BlueZ 5 transition).

It makes Audio, RecentConns, and blueman-assistant work for BlueZ 5 (https://github.com/blueman-project/blueman/issues/13#issuecomment-59015346). Serial support is unfinished and this change currently also breaks it for BlueZ 4!

@AlexDaniel [provided](https://github.com/blueman-project/blueman/issues/85#issuecomment-67269716) this traceback I can work on while waiting for my HC-05:

````
Function on_request_menu_items on Services Failed 
  File "/usr/local/lib/python2.7/dist-packages/blueman/main/PluginManager.py", line 206, in Run
    ret = getattr(inst, function)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/blueman/plugins/manager/Services.py", line 66, in on_request_menu_items
    add_menu_item(manager_menu, service)
  File "/usr/local/lib/python2.7/dist-packages/blueman/plugins/manager/Services.py", line 37, in add_menu_item
    if service.connected:
  File "/usr/local/lib/python2.7/dist-packages/blueman/Service.py", line 74, in connected
    return self.serial_port_id(args[0])
IndexError: tuple index out of range
````

### CODE REVIEW

**Comment by cschramm** [comment]
Another one from @rcath (#117); looks like SerialManager tries to work on an already removed device:

````
  File "/usr/lib/python2.7/site-packages/dbus/connection.py", line 230, in maybe_handle_message
    self._handler(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/blueman/bluez/PropertiesBlueZInterface.py", line 51, in wrapper
    handler(name, value, **kwargs)
  File "/usr/lib/python2.7/site-packages/blueman/plugins/applet/SerialManager.py", line 48, in on_device_property_changed
    d = Device(path)
  File "/usr/lib/python2.7/site-packages/blueman/main/Device.py", line 39, in __init__
    self.Properties = self.Device.get_properties()
  File "/usr/lib/python2.7/site-packages/blueman/bluez/errors.py", line 143, in warp
    raise parse_dbus_error(exception)
DBusException: org.freedesktop.DBus.Error.UnknownObject: Method "GetAll" with signature "s" on interface "org.freedesktop.DBus.Properties" doesn't exist
````

**Comment by cschramm** [comment]
@AlexDaniel: I've fixed the above error and played a little bit with my HC-05, but unfortunately I couldn't achieve anything. When I try to connect the serial port BlueZ just reports 'Operation currently not available'. This is a low level issue and independent of the connection method I use. Do I have to do something with the device to get the service running?

**Comment by AlexDaniel** [comment]
@cschramm Hi! Once you power up HC-05, everything should be ready. Can you see that it is up and running? If yes, can you connect to it with downgraded bluez and blueman? Or can you connect with this perl script?

```
use strict;
use warnings;
use v5.10;

use Net::Bluetooth;
use IO::Handle;

my $BluetoothAddress = '00:13:12:25:47:81';
my $BluetoothPort = 1;

sub InitConnection {
  my $obj = Net::Bluetooth->newsocket("RFCOMM");
  return unless defined $obj;
  return if $obj->connect($BluetoothAddress, $BluetoothPort) != 0;
  *SERVER = $obj->perlfh();
}

InitConnection();

while (1) {
  say 'Waiting for data...';
  my $buf = '';
  my $amount = read(SERVER, $buf, 1);
  say "Read $amount bytes ($buf)";
}
```
This script requires libnet-bluetooth-perl from debian repos (or Net::Bluetooth from CPAN). Also, you should change the address according to your device.

cschramm:
@AlexDaniel Thanks! The script works for me. So the device is fine. Unfortunately the way I implemented serial support cannot yet work... looks like we need to use libbluetooth like you do in the Perl script as BlueZ removed the serial handling from their dbus API. :disappointed: There is a wrapper with python-bluez, but I'm not sure if I want to pull that in...

cschramm:
I'm going for a c implementation I'll add to libblueman to create and remove RFCOMM devices, a little bit like this:

````c
#include <bluetooth/bluetooth.h>
#include <bluetooth/rfcomm.h>
#include <sys/ioctl.h>
#include <stdio.h>

int create_rfcomm_device(char *laddrstr, char *raddrstr) {
    int sk;
    struct sockaddr_rc laddr, raddr;
    struct rfcomm_dev_req req;
    int dev;

    sk = socket(AF_BLUETOOTH, SOCK_STREAM, BTPROTO_RFCOMM);
    if (sk < 0) {
        perror("Can't create RFCOMM socket");
        return -1;
    }

    laddr.rc_family = AF_BLUETOOTH;
    str2ba(laddrstr, &laddr.rc_bdaddr);
    laddr.rc_channel = 0;

    if (bind(sk, (struct sockaddr *) &laddr, sizeof(laddr)) < 0) {
        perror("Can't bind RFCOMM socket");
        close(sk);
        return -1;
    }

    raddr.rc_family = AF_BLUETOOTH;
    str2ba(raddrstr, &raddr.rc_bdaddr);
    raddr.rc_channel = 1;

    // TODO: Alternate channel
    // channel = strtol(pattern, &endptr, 10);
    // if ((endptr && *endptr != '\0') || channel < 1 || channel > 30)
    //     return btd_error_does_not_exist(msg);

    if (connect(sk, (struct sockaddr *) &raddr, sizeof(raddr)) < 0) {
        perror("Can't connect RFCOMM socket");
        close(sk);
        return -1;
    }

    memset(&req, 0, sizeof(req));
    req.flags = (1 << RFCOMM_REUSE_DLC) | (1 << RFCOMM_RELEASE_ONHUP);
    bacpy(&req.src, &laddr.rc_bdaddr);
    bacpy(&req.dst, &raddr.rc_bdaddr);
    req.channel = raddr.rc_channel;
    req.dev_id = -1;

    dev = ioctl(sk, RFCOMMCREATEDEV, &req);
    if (dev < 0) {
        perror("Can't create RFCOMM TTY");
        close(sk);
        return -1;
    }

    printf("/dev/rfcomm%d", dev);

    return 0;
}

int main(int argc, char **argv) {
    return create_rfcomm_device(argv[1], argv[2]);
}
````

(This already works, I just need to alternate the channel to open multiple. Looks like the BlueZ implementation derived the channel from the UUID)

cschramm:
I can successfully connect with the current draft. Unfortunately it does not yet recognize the existing connection, so you currently can neither see nor disconnect it.

@AlexDaniel Can you confirm connecting works for you too?

AlexDaniel:
@cschramm Well, good job! It works.
What do you mean by "neither see nor disconnect it"?
I can definitely see it:
blueman screenshot with connected serial port
One thing that looks a bit weird is that after you connect it, the connection disappears after a little while (like 10 seconds?). When you touch the /dev/rfcomm0 file with anything (like ``cat /dev/rfcomm0``) it pops up again. I don't remember such behavior in previous version, but it works nicely.

Now, here is the right-click menu:
right click menu with two disconnect buttons
There are two disconnect buttons, none of them works.

cschramm:
Interesting. Actually it looks like I implemented two ways to iterate over the open ports and add disconnect entries for them. One of those methods does only work with the "Refresh Services" function of BlueZ 4 and I just removed all the related stuff (feels so good to throw out ballast :smiley:). The other method should work independent of the dbus API and in fact shows an item for you. I should get that too, will try again.

Yes, I can confirm that the BlueZ device goes into disconnected state when the channel is idle. If that's an issue, I think we could keep the the created device file open to avoid that, but I guess it's fine once the user notices what's going on.

cschramm:
@AlexDaniel I fixed the disconnect and it works for me, although it throws an error as the ioctl call to release gives a negative return code. Another small issue I encountered is that when restarting blueman (not sure if manager or applet) the existing connections re-appear in the device's menu although they are not released. I'll look into both and can then finally get this branch ready for master. :smiley:

cschramm:
Hmm... It's not about restarting blueman, but idle devices are just not listed by the RFCOMMGETDEVLIST call. There's a pretty easy way to handle this which both @AlexDaniel's script and the BlueZ rfcomm tool take: Opening the file and keeping it open so it does not become idle. Unfortunately blueman-applet may not be able to that due to the file permissions and in fact I need to run the rfcomm tool as root for this to work. Although we have a privileged process that could handle that for us - blueman-mechanism - it is currently not expected to keep running, but terminate after idling for 30 seconds, so it cannot keep a file open for us. We either need to change this or find another way.

Anyway, I really want to get this into master now, with or without imperfect serial support. I rebased it and will tackle the issue in blueman.services.meta.NetworkService to merge it soon. If anybody sees any other issues in this branch, please intervene.

AlexDaniel:
Hello, Christopher!
I agree that it should be merged, it is better than nothing. And it works OK.
* You said that you need root privileges to access /dev/rfcommX, right? Well, first of all your user must be in the **dialout** group. Then, if it still doesn't work, run ``sudo killall ModemManager``. I have no idea why, but ModemManager blocks it. If that does not work, ``sudo /etc/init.d/bluetooth restart``, then kill ModemManager again because it might have restarted, then try again. It is so weird that you have to perform all these actions to make it work on debian, but that's how it been even before any serial port changes in blueman.
* Today I have faced another problem. My 3d printer has HC-05 attached to it, so that I can control it over bluetooth. This was working fine since forever, but there are some problems in the most recent version of blueman (I have rebuilt blueman with the latest changes, same services branch). I connect to Serial Port and it kinda works (connects and disappears), then I try to connect to /dev/rfcomm1 in [Pronterface](https://github.com/kliment/Printrun) (you might want to take a look, it is written in python as well) but it says "No such file or directory". All surprised, I connect Serial Port again and try ``cat /dev/rfcomm1`` in the terminal, it works. So I cancel the ``cat`` command, reconnect the Serial Port, click "Connect" in Pronterface again, but I get the same thing, "No such file or directory". What? It feels like Pronterface has to touch /dev/rfcomm1 twice... Otherwise I can't see any reason for it to disconnect.  But then I've found a workaround: run ``cat /dev/rfcomm1``, keep it running and then click Connect in Pronterface. This way it works. Weird.

cschramm:
Urgs... That considerably raises the importance of keeping the files open.

I'll look into how to do it. The solution will probably be to keep blueman-agent (or some child process) running when there are rfcomm device files.

cschramm:
I've drafted a system to handle this. blueman-mechanism (root) opens a subprocess (root) which opens a file handle and runs until terminated by blueman-mechanism again.

There are obviously situations where the file could vanish without SerialManager initiating a disconnect, so I'm going to add an GIO file monitor or do some polling to kill the process if the file is gone.

AlexDaniel:
Do you want me to test it?

cschramm:
Sure, go ahead. It worked fine for my limited testing. I also added the GIO monitor in the mean time, so the helper process will recognize and quit if the file is gone.

I think this should tackle all problems mentioned, except the generic disconnect entry at the bottom of the device menu which is not exactly serial related. I guess I should remove that one in general.

AlexDaniel:
Hm, it works. It feels weird though. I am using bluetooth every day so I'll let you know if I stumble upon any other problems. All is going well so far.

AlexDaniel:
Sometimes I can't even get it to work:
traceback

cschramm:
Strange. This error means the blueman-rfcomm-watcher binary cannot be found (in your PATH). No idea how that could happen "sometimes"...

AlexDaniel:
$ type blueman-rfcomm-watcher 
    blueman-rfcomm-watcher is /usr/local/bin/blueman-rfcomm-watcher

cschramm:
Any chance to reproduce the behavior or does it appear random?

AlexDaniel:
I don't know yet.

cschramm:
OK, this is now in master. If your problem persists, please open a new issue for it.

Issue #161, opened 2014-12-28
dbus failure in fedora 21


raveit65:
Hi,
first i wanna say a big thank you for re-starting development.
I wanna bring back blueman in fedora.
Unfortunately, the blueman-applet doesnt start in fedora 21, from a terminal:

```
[rave@mother ~]$ blueman-applet 

** (blueman-applet:1581): WARNING **: Couldn't register with accessibility bus: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
Loading configuration plugins
blueman-applet version 1.99.alpha1 starting
Stale PID, overwriting
Using GConf config backend
_________
Load (/usr/lib/python2.7/site-packages/blueman/main/PluginManager.py:54)
['TransferService', 'NMPANSupport', 'RecentConns', 'DhcpClient', 'SerialManager', 'ExitItem', 'AppIndicator', 'StandardItems', 'DBusService', 'Menu', 'StatusIcon', 'NetUsage', 'PPPSupport', 'ShowConnected', 'AuthAgent', 'Networking', 'NMDUNSupport', 'PowerManager', 'DiscvManager', 'KillSwitch', 'Headset'] 
Using GConf config backend
_________
Load (/usr/lib/python2.7/site-packages/blueman/main/PluginManager.py:54)
Unable to load plugin module AppIndicator
cannot import name AppIndicator3 
_________
get_interface_version (/usr/lib/python2.7/site-packages/blueman/bluez/BlueZInterface.py:8)
Detected BlueZ 5 
_________
__load_plugin (/usr/lib/python2.7/site-packages/blueman/main/PluginManager.py:127)
loading <class 'blueman.plugins.applet.StatusIcon.StatusIcon'> 
_________
__load_plugin (/usr/lib/python2.7/site-packages/blueman/main/PluginManager.py:127)
loading <class 'blueman.plugins.applet.DBusService.DBusService'> 
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/blueman/plugins/BasePlugin.py", line 65, in _load
    self.on_load(parent)
  File "/usr/lib/python2.7/site-packages/blueman/plugins/applet/DBusService.py", line 35, in on_load
    self.add_dbus_method(self.ServiceProxy, in_signature="sosas", async_callbacks=("ok", "err"))
[...]

PR #168, opened 2015-01-17
Python3 fixes

infirit:
All these changes are backwards compatible with python 2.7 but I like to see someone test this to be sure. If this is working fine for others I'll push this to master.

Note, after this it does __not__ run under python3. It fixes all errors up to metaclass errors and then my eyes glazed over. For reference the error:
```
  File "/usr/bin/blueman-applet", line 114, in <module>
    BluemanApplet()
  File "/usr/bin/blueman-applet", line 57, in __init__
    self.Plugins.Load()
  File "/usr/lib64/python3.3/site-packages/blueman/main/PluginManager.py", line 85, in Load
    __import__(self.module_path.__name__ + ".%s" % plugin, None, None, [])
  File "/usr/lib64/python3.3/site-packages/blueman/plugins/applet/AuthAgent.py", line 3, in <module>
    import blueman.main.applet.BluezAgent as BluezAgent
  File "/usr/lib64/python3.3/site-packages/blueman/main/applet/BluezAgent.py", line 33, in <module>
    class CommonAgent(GObject.GObject, Agent):
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```


cschramm:
Great!

What my IDE still detects is:

* [ ] `__builtin__` in Constants.py
* [ ] u prefix in blueman-sendto due to the ∞ character
* [ ] u prefix in iniparse/ini.py due to some \ufeff character (dito)
* [ ] basestring type in iniparse/compat.py

I'm fine with keeping the u prefix and not supporting version 3.2 if things would get ugly.

Concerning the `basestring`, I think in our case we can replace it with `str` or even just remove the checks, since we're always calling `read` and `set` with `str` objects, so the condition in `read` is always True and the condition in `set` is always False.

Concerning your traceback: `CommonAgent` is derived from `GObject` and `Agent` and tries to use `DummyGObjectMeta` as its type which is a subclass of both `GObject`'s type `gi.types.GObjectMeta` and `Agent`'s type `dbus.service.InterfaceType`, so it should be correct. Maybe due to your dbus changes, `Agent` now has a different type? Then we just need to update `DummyGObjectMeta`'s parent accordingly. (Just to note: I have no idea what that type is there for. :wink: )

cschramm:
> Maybe due to your dbus changes, `Agent` now has a different type?

No, not for me:

````
>>> from blueman.bluez.Agent import Agent
>>> type(Agent)
<class 'dbus.service.InterfaceType'>
>>> from gi.repository import GObject
>>> type(GObject.GObject)
<class 'gi.types.GObjectMeta'>
````

So this error is really strange as what it demands is obviously satisfied... :confused:

**Comment by cschramm** [comment]
Using `class CommonAgent(GObject.GObject, Agent, metaclass=DummyGObjectMeta)` instead works in python 3 :smiley:, but gives a SyntaxError in 2.7 :disappointed:.

**Comment by infirit** [comment]
>Concerning the basestring, I think in our case we can replace it with str or even just remove the checks, since we're always calling read and set with str objects, so the condition in read is always True and the condition in set is always False.

I pushed a [commit](https://github.com/infirit/blueman/commit/212ce5c1e3c62291ba479f70deb9d32e5f94657b) that gets rid of the basestring usage. It is a bit hacky but solves the problem.

The ide is probably doing a simple search and missing the logic to detect [try/except](https://github.com/infirit/blueman/commit/1b526ab4d5b583702c2de5437a59e43271c52c00#diff-08a0dea1fcdad3a2d92d8ff168bdef0dL15) around the import.

On the ``u`` prefix, I do not know of a way to work around this. I think all distributions will ship with 2.7 and a 3.x version of python. So for distro's that do not have 3.3 yet they can use 2.7.

metaclass stuff always makes my head spin so wish my luck :wink:

**Comment by infirit** [comment]
And when I hit comment your other comments showed up so I did not read them :smile:

**Comment by cschramm** [comment]
> On the u prefix, I do not know of a way to work around this. I think all distributions will ship with 2.7 and a 3.x version of python. So for distro's that do not have 3.3 yet they can use 2.7.

Agree, changed the goal in #19.

How did people actually *use* Python 3.1 and 3.2? :astonished:

**Comment by infirit** [comment]
So what I found is that ``__metaclass__`` is ignored and it should be passed as an argument (as you found out). And there is no compatible way to do this in both versions.

The only thing I found that does this in a cross version way is the ``with_metaclass`` function from the [six](https://pythonhosted.org/six/#six.with_metaclass) module.

**Comment by cschramm** [comment]
Or this: https://wiki.python.org/moin/PortingToPy3k/BilingualQuickRef#metaclasses

Should be

````python
CommonAgent = DummyGObjectMeta(str('CommonAgent'), (GObject.GObject, Agent), {
    ...
    })
````

for us. :smiley: A bit ugly, but avoids the six dependency.

**Comment by cschramm** [comment]
OK, it's not that easy. We'd have to change the complete class to string names then...

**Comment by infirit** [comment]
:sob: I was confused how this was supposed to work. This is the part of classes that I do not  understand well as I never used it like this before.

What we could do is add a configure option to switch between python version and then do some Makefile voodoo to remove the metaclass argument when python2 is selected.

**Comment by cschramm** [comment]
Well, it would work, but it's really ugly. The class would start like this:

````python
CommonAgent = DummyGObjectMeta(str('CommonAgent'), (GObject.GObject, Agent), {
    '__gsignals__': {
        'released': (GObject.SignalFlags.NO_HOOKS, None, ()),
    },

    '__init__': lambda (self, status_icon, path, time_func, notifications):
````

but since there are no multiline lambdas and the class has 150 lines of code, that's not very useful.

I like the Makefile idea. Seems nicer than adding the six dependency for this.

**Comment by cschramm** [comment]
So, how would you proceed? Should we rename the file to BluezAgent.py.in and let the py file be generated? Or did you think of shipping the py file and modifying it on make?

**Comment by infirit** [comment]
Added a new commit to handle this case. It works all ok for me on Python2 but more problems still for Python3 :cry:

**Comment by cschramm** [comment]
Any tracebacks? Works fine for me so far.

**Comment by infirit** [comment]
```python
get_interface_version (/usr/lib64/python3.3/site-packages/blueman/bluez/BlueZInterface.py:8)
Detected BlueZ 5 
  File "/usr/bin/blueman-applet", line 114, in <module>
    BluemanApplet()
  File "/usr/bin/blueman-applet", line 57, in __init__
    self.Plugins.Load()
  File "/usr/lib64/python3.3/site-packages/blueman/main/PluginManager.py", line 85, in Load
    __import__(self.module_path.__name__ + ".%s" % plugin, None, None, [])
  File "/usr/lib64/python3.3/site-packages/blueman/plugins/applet/SerialManager.py", line 74
    def on_script_closed(self, pid, cond, (address, node)):
                                          ^
SyntaxError: invalid syntax
```

infirit:
I pushed some more commits but still more to do (will it ever end :grinning: )
```python
__load_plugin (/usr/lib64/python3.3/site-packages/blueman/main/PluginManager.py:128)
loading <class 'blueman.plugins.applet.DhcpClient.DhcpClient'> 
  File "/usr/lib64/python3.3/site-packages/blueman/plugins/BasePlugin.py", line 53, in _load
    self.on_load(parent)
  File "/usr/lib64/python3.3/site-packages/blueman/plugins/applet/DhcpClient.py", line 17, in on_load
    self.add_dbus_method(self.DhcpClient, in_signature="s")
  File "/usr/lib64/python3.3/site-packages/blueman/plugins/AppletPlugin.py", line 60, in add_dbus_method
    self.Applet.DbusSvc.add_method(func, *args, **kwargs)
  File "/usr/lib64/python3.3/site-packages/blueman/main/DbusService.py", line 44, in add_method
    setattr(self.__class__, name, dec)
NameError: global name 'dec' is not defined
```

infirit:
The problem is ``exec()``, https://stackoverflow.com/questions/15086040/behaviour-of-exec-function-in-python-2-and-python-3/15087355#15087355

We could pass our own namespace dict [1] but I much rather like to see the removal of ``exec``. The thing is I have no clue what is going on here and why it is done like this.  @cschramm, you have much more knowledge of this part and it would be nice if you can try and remove ``exec()`` from the code :wink: 

[1] This seems to work but then still further problems come up [2].
```diff
diff --git a/blueman/main/DbusService.py b/blueman/main/DbusService.py
index 0a5050d..28c5f3e 100644
--- a/blueman/main/DbusService.py
+++ b/blueman/main/DbusService.py
@@ -33,6 +33,16 @@ class DbusService(dbus.service.Object):
         a = ",".join(a[1:])
 
         # print name, a
+        dec = ""
+        namespace = {
+            "sec": dec,
+            "func": func,
+            "dbus":dbus,
+            "dbus_interface":dbus_interface,
+            "in_signature":in_signature,
+            "args":args,
+            "kwargs":kwargs}
+
         exec(\
 """def %(0)s(self, %(1)s):
     return self.%(0)s._orig_func(%(1)s)

```

[2] After [1] this repeats indefinitely.. :sob: 
```python
Destroy (/usr/lib64/python3.3/site-packages/blueman/main/Device.py:98)
invalidating device None 
Exception RuntimeError: 'maximum recursion depth exceeded while calling a Python object' in <bound method Device.__del__ of <Device object at 0x7f6d8e0746e0 (blueman+main+Device+Device at 0x15c8ba0)>> ignored
_________
__del__ (/usr/lib64/python3.3/site-packages/blueman/main/Device.py:60)
deleting device None 
```

cschramm:
Yes, `blueman.main.DbusService` is a pain, see #161 and [RedHat #1177659](https://bugzilla.redhat.com/show_bug.cgi?id=1177659). I'll try to refactor that.

cschramm:
I've refactored the dbus service in #170. It seems to work fine, but I'd like somebody else to have a look at it before merging into master. You can rebase the python3 branch, just skip the commit "Use the exec function instead of the statement".

To get a chance of a reviewer, I'll try to describe shortly what's going on in the DbusService class:

Basically the class needs a method with the same name that gets called on the bus and some dbus-specific attributes, e.g. `_dbus_is_method`. Those attributes are set by the `dbus.service.method` (or `dbus.service.signal`) decorator which uses inspection to extract, check, and store the signature of the function that's being decorated.

The exec call previously defined a function with the same name and same signature as the method that shall be added (from a different class). This function just calls the original method providing the arguments it got. The new function then gets decorated (within exec) and added to the class (after exec). I think `inspect.getargspec` plus `exec` is the only way to "copy" a signature.

How I do it now: I decorate the methods at their original location (some plugin class) as usual, so they do have the required dbus attributes with the correct name and signature. When the plugin they belong to gets loaded (`dbus.plugins.AppletPlugin._load`) I search the class for decorated methods (`dbus.main.DbusService._definitions`) and create wrappers for them (`dbus.main.DbusService._add_wrapper`). Instead of defining them with the same name, I just copy `__name__` after they are created with the name 'wrapper'. Instead of decorating them, I just copy the dbus attributes from the already decorated method. The signature does not matter then so I can use `*args, **kwarg` and just pass on all arguments to the original method (skipping the `dbus.main.DbusService` instance, it is bound to the original `self` anyway; maybe I could even make the wrapper method static somehow, so that it does not even get the instance?).

cschramm:
Btw. if anybody wants do some own testing: You can easily inspect what gets extracted to the bus with

    dbus-send --dest=org.blueman.Applet --print-reply / org.freedesktop.DBus.Introspectable.Introspect

and test a sample method call with

    dbus-send --dest=org.blueman.Applet --print-reply / org.blueman.Applet.QueryPlugins

(Apart from org.blueman.Applet on the session bus, we also have org.blueman.Mechanism on the system bus.)

cschramm:
One more thing I found: We need to replace the tuple unpacking here:

````
blueman/gui/manager/ManagerDeviceMenu.py:    def on_device_property_changed(self, List, device, iter, (key, value)):
blueman/gui/manager/ManagerToolbar.py:    def on_adapter_property_changed(self, List, adapter, (key, value)):
blueman/gui/manager/ManagerToolbar.py:    def on_device_propery_changed(self, dev_list, device, iter, (key, value)):
blueman/gui/DeviceSelectorWidget.py:    def on_adapter_prop_changed(self, devlist, adapter, (key, value)):
apps/blueman-manager:        def prop_changed(List, adapter, (key, value)):
````

infirit:
Thanks, I thought I grepped for more occurrences..

I fixed that up, took out the exec commit and rebased against master.

cschramm:
I'll merge in #170 soon, if nobody raises a concern.

cschramm:
OK, dbus refactoring is merged.  I think we can get this in as well. History looks nice to me, no need to rebase. I'm not sure if we should add Python 3 compatibility to CHANGELOG and README already. Don't know anything better than just letting users of the master branch test it out, though. :smirk_cat:

infirit:
I'll do a final rebase against master and make sure nothing broke python2. Then I'll push it into master.

All Exceptions should be gone (at least for me) now but there is still more work to be done unfortunately.

PR #170, opened 2015-01-29
Refactor DbusService

cschramm:
* Use the standard dbus-python decorators to add decorated methods to the plugin classes.
* When initializing the plugin class create wrappers for all dbus methods and copy the dbus attributes from the original methods.
* No more `exec` and similar magic is involved anymore. (References #168)
* Re-run `dbus.service.InterfaceType.__init__` to refresh the internal dbus class table instead of modifying it directly. (References #161)

### DIFF
--- a/blueman/main/DbusService.py
+++ b/blueman/main/DbusService.py
@@ -1,7 +1,6 @@
 import dbus
 import dbus.glib
 import dbus.service
-import inspect
 
 
 class MethodAlreadyExists(Exception):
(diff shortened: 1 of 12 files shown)


cschramm:
See https://github.com/blueman-project/blueman/pull/168#issuecomment-72161892 for some detailed description of the changes.

infirit:
Sorry I forgot to mention the things I found when I tried this. The applet and manager work ok, pairing works and other actions work. It is just in the services part I see some problems.

On start-up I get a pop-up with:
```
UnknownMethodException: org.freedesktop.DBus.Error.UnknownMethod: Unknown method: ReloadNetwork is not a valid method of interface org.blueman.Mechanism

You might not be able to connect to the Bluetooth network via this machine
```

Then with ``on_apply`` in the services (which I suspect has to do with the popup). The below I get when I change an option in the audio tab.
```python
on_apply (/usr/lib64/python2.7/site-packages/blueman/plugins/services/Audio.py:55)
org.freedesktop.DBus.Error.UnknownMethod: Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/dbus/service.py", line 654, in _message_cb
    (candidate_method, parent_method) = _method_lookup(self, method_name, interface_name)
  File "/usr/lib64/python2.7/site-packages/dbus/service.py", line 246, in _method_lookup
    raise UnknownMethodException('%s is not a valid method of interface %s' % (method_name, dbus_interface))
UnknownMethodException: org.freedesktop.DBus.Error.UnknownMethod: Unknown method: SetBluezConfig is not a valid method of interface org.blueman.Mechanism
```

cschramm:
Damn. I failed to both implement _and_ test blueman-mechanism correctly. :stuck_out_tongue_closed_eyes: 

Fixed that.

infirit:
Much better :smile_cat:. When I try this with latest master (after the merges in services branch) I see the below when changing the audio preferences.
```python
on_apply (/usr/lib64/python2.7/site-packages/blueman/plugins/services/Audio.py:55)
org.freedesktop.DBus.Python.ConfigParser.NoSectionError: Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/dbus/service.py", line 707, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/usr/lib64/python2.7/site-packages/blueman/main/DbusService.py", line 28, in wrapper
    return method(*args[1:], **kwargs)
  File "/usr/lib64/python2.7/site-packages/blueman/plugins/mechanism/Config.py", line 21, in SetBluezConfig
    c.set(section, key, value)
  File "/usr/lib64/python2.7/site-packages/blueman/iniparse/compat.py", line 305, in set
    ConfigParser.set(self, section, option, value)
  File "/usr/lib64/python2.7/site-packages/blueman/iniparse/compat.py", line 169, in set
    raise NoSectionError(section)
NoSectionError: No section: dbus.String(u'General')
```

cschramm:
This is #174. I think you should also get it with plain master (and every other version), no?

infirit:
Ah, yes it is so ignore that last bit then. Then there is nothing major holding this up I guess.

PR #173, opened 2015-01-31
Add blueman-report

cschramm:
As I always get to the same process and problems when trying to get behind reporters' issues, I've hacked together a small report script that

* starts a new blueman-applet,
* asks the user what actions he takes,
* uploads the log to GitHub,
* and provides a template with essential data for the bug report.

Sample output:

> Describe your next action (keep empty if done): Open manager window
> Perform the described action
> Describe your next action (keep empty if done): Close manager window
> Perform the described action
> Describe your next action (keep empty if done): 
> 
> =====================================
> 
> Use this data in your report:
> 
> =====================================
> 
> blueman: 1.99.alpha1
> BlueZ: 5.23
> Distribution: 
> Desktop: XFCE
> 
> https://gist.github.com/6c4fbdd2435520bb846f#file-log-L141 Open manager window
> https://gist.github.com/6c4fbdd2435520bb846f#file-log-L251 Close manager window

It currently only targets python 2 due to the use of urllib2 (vs. urllib.request) and raw_input (vs. input).

### DIFF
--- a/apps/blueman-report
+++ b/apps/blueman-report
@@ -6,9 +6,15 @@ import signal
 import subprocess
 import time
 from threading import Thread, Lock
-import urllib2
 from blueman.Constants import VERSION
 from blueman.Functions import get_lockfile, get_pid, is_running
+
+import sys
+if sys.version_info.major == 2:
+    input = raw_input
+    import urllib2
+else:
+    import urllib.request as urllib2
 
 for program in 'adapters', 'applet', 'manager', 'services':
     name = 'blueman-' + program
@@ -33,7 +39,7 @@ def copy_output():
     for line in iter(applet.stdout.readline, b''):
         lock.acquire()
         lines += 1
-        log += line
+        log += line.decode('UTF-8')
         lock.release()
 
 t = Thread(target=copy_output)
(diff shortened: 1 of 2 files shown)


cschramm:
@infirit 

> It currently only targets python 2 due to the use of urllib2 (vs. urllib.request) and raw_input (vs. input).

Anything better I can do about that than try, except?

infirit:
For input/raw_input I would replace all occurrences of ``raw_input`` with ``input`` and add the following to the top. Normally for urllib2 a try/except would do but now that we already check the major version there is no need for that.

```python
import sys
if sys.version_info.major == 2:
	input = raw_input
	import urllib2
else:
	import urllib.request as urllib2
```

infirit:
Also, the assumption that cache dir is in ``~/.cache`` __may__ not be true (it is in most cases though). GLib provides the ``GLib.get_user_cache_dir()`` [function](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-user-cache-dir) but we can delay using this until someone reports a bug :wink: .

cschramm:
Good point. I just copied that code from [check_single_instance](https://github.com/blueman-project/blueman/blob/75b169e51dcbd3b88643aafb8691c6016b55c1c2/blueman/Functions.py#L292). We should probably change it there.

cschramm:
I've refactored check_single_instance to use the GLib function and to avoid duplication.

PR #177, opened 2015-02-04
Game controller wakelock

bwRavencl:
Added a plugin that temporarily suspends the screensaver when a bluetooth game controller (either a gamepad or a joystick) is connected.

### DIFF
--- a/blueman/plugins/applet/GameControllerWakelock.py
+++ b/blueman/plugins/applet/GameControllerWakelock.py
@@ -1,0 +1,63 @@
+import blueman.bluez as bluez
+from blueman.Functions import *
+from blueman.main.SignalTracker import SignalTracker
+from blueman.main.Device import Device
+from blueman.plugins.AppletPlugin import AppletPlugin
+from gi.repository import Gdk, GdkX11
+import subprocess
+
+class GameControllerWakelock(AppletPlugin):
+    __description__ = _("Temporarily suspends the screensaver when a bluetooth game controller is connected.")
+    __author__ = "bwRavencl"
+    __icon__ = "input-gaming"
+
+    def on_load(self, applet):
+        self.wake_lock = 0
+        self.root_window_id = "0x%x" % Gdk.Screen.get_default().get_root_window().get_xid()
+
+        self.signals = SignalTracker()
+        self.signals.Handle("bluez", bluez.Device(), self.on_device_property_changed, "PropertyChanged", path_keyword="path")
+        
+    def on_unload(self):
+        if self.wake_lock:
+            self.wake_lock = 1
+            self.suspend_screensaver("resume")
+        self.signals.DisconnectAll()
+
+    def on_device_property_changed(self, key, value, path):
+        if key == "Connected":
+            klass = Device(path).get_properties()["Class"] & 0x1fff
+
+            if klass == 0x504 or klass == 0x508:
+                if value:
+                    self.xdg_screensaver("suspend")
+                else:
+                    self.xdg_screensaver("resume")
+
+    def xdg_screensaver(self, action):
+        if action == "resume":
+            if self.wake_lock == 0:
+                pass
+            elif self.wake_lock > 1:
+                self.wake_lock -= 1
+                pass
+        elif action == "suspend" and self.wake_lock >= 1:
+             self.wake_lock += 1
+             pass
+
+        command = ["xdg-screensaver", action, self.root_window_id]
+
+        try:
+            process = subprocess.Popen(command, stderr=subprocess.PIPE)
+            _, stderr = process.communicate()
+
+            if process.returncode != 0:
+                dprint(" ".join(command) + " failed with: " + stderr)
+                pass
+
+            if action == "suspend":
+                self.wake_lock += 1
+            elif action == "resume":
+                self.wake_lock = 0
+        except:
+            dprint(traceback.format_exc())
--- a/blueman/plugins/applet/Makefile.am
+++ b/blueman/plugins/applet/Makefile.am
@@ -1,6 +1,6 @@
 bluemandir = $(pythondir)/blueman/plugins/applet
 
-blueman_PYTHON = AppIndicator.py ExitItem.py ShowConnected.py AuthAgent.py DhcpClient.py NMPANSupport.py DBusService.py DiscvManager.py NetUsage.py Headset.py __init__.py KillSwitch.py Menu.py Networking.py PowerManager.py PPPSupport.py NMDUNSupport.py RecentConns.py SerialManager.py StandardItems.py StatusIcon.py TransferService.py
+blueman_PYTHON = AppIndicator.py ExitItem.py ShowConnected.py AuthAgent.py DhcpClient.py NMPANSupport.py DBusService.py DiscvManager.py NetUsage.py Headset.py __init__.py KillSwitch.py Menu.py Networking.py PowerManager.py PPPSupport.py NMDUNSupport.py RecentConns.py SerialManager.py StandardItems.py StatusIcon.py TransferService.py GameControllerWakelock.py
 
 CLEANFILES =		\
 	$(BUILT_SOURCES)


cschramm:
I have no idea what's that good for, but well, why not. :wink: 

Did you test it with BlueZ 4 and / or 5?

bwRavencl:
Most games on Linux do not prevent the screensaver from becoming active, when there's no keyboard or mouse input.
The plugin is useful for players of games that can be solely played with game controllers and still want to enable their screensaver.
Mac OS X provides a similar feature as my plugin - it disables the screensaver for as long as a bluetooth gamepad is connected.
I recently switched back from OS X to Linux and I got used to this behaviour, so I decided to write the said plugin.

Since I'm running Ubuntu the code is only tested with BluZ 4 so far.

cschramm:
Ah, got it. I thought it only disables the screensaver for a while after the controller is "plugged in". Disabling it for as long as it's connected makes much more sense, of course.

I'll need to adapt it to work with BlueZ 5 due to the event handler.

cschramm:
Please try to use `import blueman.bluez as bluez` and then `self.signals.Handle('bluez', bluez.Device(), self.on_device_property_changed, 'PropertyChanged', path_keyword='path')` instead of specifying the dbus event directly. This should make it work with both BlueZ 4 and 5.

bwRavencl:
@cschramm I made your proposed change in SHA: 7737ce1
However I still could only test it with BlueZ 4.

cschramm:
Works with BlueZ 5 as well. :smiley: 

I think the check `klass & 0x50C == klass` is wrong as it would also match e.g. an uncategorized peripheral device (0x504) and everything else with a subset of the 0x50c bitmap (16 different classes overall). What you meant is probably just `klass == 0x508 or klass == 0x504` (gamepad or joystick). A better way is to check `klass & 0x1fff` to ignore the three irrelevant most significant bits (they are usually zero, but I think that's not mandatory). We could also let BlueZ do the logic (see [class_to_icon](http://git.kernel.org/cgit/bluetooth/bluez.git/tree/src/dbus-common.c#n131)) and just check that the Icon property is `input-gaming`.

infirit:
Do not use ``str()`` as this does not get you bytes (codepoints) in python3. Please use ``.encode("UTF-8")`` instead which is compatible between two and three.

bwRavencl:
@cschramm Yes you are completely right about the class check, I went with your first proposal as a fix for this.

@infirit Thanks for the hint. ;-) Made your proposed change right away. I'm actually very new to Python and have never worked with Python 3 yet...

cschramm:
:+1 LGTM

infirit:
@bwRavencl then you should watch [this](http://nedbatchelder.com/text/unipain.html) talk to understand unicode/bytes in python better :wink:

bwRavencl:
@infirit So what you actually meant was that I should do it as follows:
`str(my_int).encode("UTF_8")`
As I see it this should produce a 'str' in Python 2 and 'bytes' in Python 3 - correct?

infirit:
@bwRavencl, sorry, ignore me.. I was assuming you needed bytes but you are actually converting to a string from a number, which is what is required in this case and how you had it originally is correct for both versions. My mind was in python3 porting mode where this can cause problems :smile_cat: 

ps: in python2 ``str()`` is a byte string and the python3 equivalent is ``bytes()``. The reverse is ``str()`` in python3 is codepoints (unicode) and the python2 equivalent is ``unicode()``. The ``encode`` method encodes from codepoints to bytes and the ``decode`` method decodes from bytes to codepoints. The big confusion is that python2 decodes/encodes implicitly in the background which was a mistake and python3 does not do this any-more. The talk I references discusses this confusion.

cschramm:
So, the current version in this PR is fine, isn't it? Then I'll add it to master.

bwRavencl:
Yup should be fine now. :)

infirit:
A couple of thing came to mind.

We should add some debug for when xdg_screensaver fails (non zero return code). Perhaps (totally untested) something like?

```python
def block_screensaver(self):
    if not self.blocking:
        self.window = Gtk.Window()
        self.window.realize()
        xid = str(self.window.get_property("window").get_xid())
        p = subprocess.Popen(command["xdg-screensaver", "suspend", xid],stderr=subprocess.PIPE)
        stdout, stderr = p.communicate()
        if p.returncode != 0:
            dprint("xdg-screensaver failed with: " + stderr)
        self.blocking = True
```
 And what if xdg-screensaver is not available? Then subprocess will fail, we should probably catch the ``OSError`` and handle it gracefully or check it is in $path in ``on_load``.

ps: I would prefer to figure out a way to get the root window of the applet instead of faking a window. But not sure if this is possible from within the plugin..

bwRavencl:
@infirit Yes your concerns about the error handling are valid.
I'm going to add some handling for both the cases you mentioned.

Concerning the window ID im not sure if using the ID of the main window would work.
The xdg-screensaver documentation says:
>suspend WindowID:
>Suspends the screensaver and monitor power management. WindowID must be the X Window ID of an existing window of the calling application. The window must remain in existance for the duration of the suspension.

I'm not an expert conerning X11 and GTK, but are you sure that the main window still 'exists' after if it is closed / was never opened in the first place?

infirit:
> I'm not an expert concerning X11 and GTK, but are you sure that the main window still 'exists' after if it is closed / was never opened in the first place?

I was hoping the GtkStatusIcon had a GtkWindow which would then keep running for as long as the applet is. However looking at the docs for the GtkStatusIcon it does not derive from GtkWidget so it does not create an actual window, oh well :disappointed:. Which means your dummy window solution is probably the only way.

cschramm:
An alternative to handling those error cases would be to not enable the plugin by default. Then it would just be up to the user to care about the prerequisites when enabling it explicitly.

bwRavencl:
After adding the proposed error handling I noticed that my method of using a fake GTK window didn't work reliably.

Consequently I removed the fake window completely and I am now using the ID of the root X-Window for screensaver blocking.
In order to retrieve the ID I am now using the following command:
`"xwininfo -root | grep xwininfo | cut -d' ' -f4"`

Unfortunately I wasn't able to find a nicer way to do this, however due to the error handling it shouldn't be harmful.

Also, in order to avoid duplicate code I did some reorganizing...

infirit:
Nicer way :smile: 

```python
from gi.repository import Gdk, GdkX11
rootwin = Gdk.Screen.get_default().get_root_window()
xid = rootwin.get_xid()
print(hex(xid))
```

Some more comments.. 

An option to simplify things is to use ``self.num_connections`` to determine if we are suspended. In python zero evaluates to False so the following works if num_connections is zero, ``if  self.num_connections:``. Then in run_process do ``arg = "suspend" if self.num_connections else "resume". Just make sure that when something went wrong to reset num_connections to zero.

Another thing, there is no need to initialize variables :wink:, just make sure to assign a value before using it. Also you return in run_process but never do anything with what is returned, I see no need to return anything :smile: 

Why use a ``shell=True`` for subprocess? It should normally not be needed and could be dangerous.

Why do ``returncode = process.returncode``, you are just making a reference to the same object?

infirit:
Untested version of what I would do, https://gist.github.com/infirit/7a74142dffe3480d3ca5

bwRavencl:
@infirit I adopted your proposed code for the most part as it seems to be the cleanest solution so far

I made a minor correction to also handle the case when a controller was already connected before the plugin is started and then gets disconnected after the plugin is initialized.

infirit:
You picked my initial version which I updated later to account for multiple devices being connected, see the revision diff https://gist.github.com/infirit/7a74142dffe3480d3ca5/revisions

The reason I did this was to avoid running the command unnecessarily as running it again while already suspended does not do anything useful :smile: 

Other than that it look good now.

infirit:
Oh, and unless @cschramm disagrees please squash this into a single commit.

bwRavencl:
*Done* :smile:

cschramm:
Thanks, it's in master.

PR #182, opened 2015-02-07
Remove KillSwitchNG

### LINKED ISSUE #176: Strange chmod /dev/rfkill
https://github.com/blueman-project/blueman/blob/master/blueman/plugins/mechanism/RfKill.py#L18

    os.chmod("/dev/rfkill", 0o655)

This chmod sets 655 (rw-r-xr-x) on /dev/rfkill. Why do we need X bit on this device? It looks like something strange.

Also blueman should not touch /dev/rfkill permissions at all. They are managed by udev and system administrator. If system administrator allows users to read or write rfkill state, he will give them permission. If users are not allowed to do it, blueman must not try to get this permission, it should use some fallback. Also bluez D-Bus interface allows to get and set power state of adapters (org.bluez.Adapter1.Powered property).

cschramm:
Closes #176
Closes #182


infirit:
Is there a reason to have two methods to do this? It falls back to the privileged process anyway so perhaps it is an idea to always do it through mechanism and simplify things?

cschramm:
No, this does not make sense. I will just remove the class and put the respective functionality into the applet and mechanism plugins. There is also some dead code (the do_... methods).

cschramm:
OK, this is weird. I removed the class completely and moved the functionality into the plugins. Thereby I noticed that due to the dead code the state was never changed and the only effect of all the code now in the `on_load`, `on_unload`, and `io_event` methods of the KillSwitch plugin is that its `on_query_status_icon_visibility` method will force the status icon to show if there is a bluetooth kill switch registered (it would not be in case there is no bluetooth adapter).

`on_power_state_change_requested` is the only method of the plugin that relates to the main purpose of the plugin as written in its description.

I removed all of the ineffective code in this commit.

cschramm:
@infirit Or may it be that the do_ method definitions trigger some glib signals magic?

infirit:
The ``do_`` functions are run when the corresponding signals as defined by ``__gsignals__``.

cschramm:
Then forget all my nonsense. :wink: I'll clean this up.

cschramm:
OK, next try. Now the KillSwitch plugin has all the logic that was previously contributed by KillSwitchNG. It should now be pretty clear what's going on.

There is one difference I purposely made, though: In the previous version the `on_query_status_icon_visibility` method also forces the status icon to show if bluetooth is *not* blocked but there is a bluetooth killswitch registered and the BlueZ daemon is running. If an adapter is present, the icon is shown in any case, no need to force it. If that's not the case, I cannot think of a reason to show the icon as blueman is useless then anyway (there is neither a soft killswitch to unlock, nor a supported adapter to use).

PR #198, opened 2015-02-25
blueman-report got left off the /usr/bin/env changes

rworkman:
I noticed this while looking into how trivial it might be to allow specifying python2 or python3 at build time (and use the autoconf tools to sub that into the relevant places). Is there any interest in that or would you prefer to use whatever the system python is?


rworkman:
Eek. This included the rebased commits from master, and I only wanted that one for blueman-report. Dammit.

PR #204, opened 2015-03-02
Gsettings port

infirit:
This implements the use of gsettings in Blueman and removes the old config plugin system.

Everything looks to be working properly but needs testing, especially Transfer and Network services which are broken currently on my machine.

One thing I'll do later is add a gconf -convert file for the options that can be converted.

I also have a bugfix and variable rename in here I stumbled upon while working on this.

Todo items for the future:
 * Bind more widgets to keys with the bind_to_widget function.
 * Look at the apply code in the services plugins. For example most of the Transfer ui options are already instant apply but still make the button active. Use use instant apply/checks plus move some code to the close button to remove the apply button?


cschramm:
Looks good from a quick code review that I did yesterday.

I'll play around with the network and transfer parts.

The FIXME in blueman-manager sounds like the signal will not be emitted, right?

What's left to do is a changelog entry for this. I think the transfer plugin fix deserves it's own bug fix entry. It looks like we do not need to change the dependencies as glib includes GSettings and gconf isn't mentioned in the README anyway.

infirit:
>The FIXME in blueman-manager sounds like the signal will not be emitted

We can only bind a key to a single widget but in the menu we have two, one for each sorting checkbox. 

Then there is the way the TreeView resort is done, calling ``self.List.DisplayKnownDevices(autoselect=True)`` instead of using Gtk directly on the TreeView. I do not understand why it was done like this so I left it alone. I did rename the function name to reflect that the re-sorting was the last user. 

The FIXME is something to figure out why this is done as it is and perhaps use Gtk to sort the TreeView at some point in the future but I can see it is nor clear enough so I'll reword it :smile:. 

>It looks like we do not need to change the dependencies as glib includes GSettings

Yep, You get it for "free".

I'll add entires to the Changelog now.

cschramm:
I found some issues in the network settings:

First, `on_query_apply_state` should not have been ported. `self.get_options()` has nothing to do with `self.Config` and also the logic got changed (previously `True` was returned if props contained anything, now it's only `True` if `props` contains `"ip"`).

Second, I cannot disable the nap-enable option as line 53 flips `value` from `False` to `True`. :cry:

infirit:
I pushed a new commit what I'll squash later with the original. I hope It fixes the problems you reported.

It should also remove the race condition and the need to ignore keys in ``on_property_changed``. 

Can you confirm something please. I added an "ip" key in ``org.blueman.network" (don't remember so don't ask why :grin:) which looking at it now is unnecessary. I do not see any references in the code to store the ip address as an configuration (Config) option. Where is it stored, in Bluez somewhere? If in Bluez I need to remove it from the schema :smile:

infirit:
Another question, I set the default width for blueman-manager to 750 pixels which I think is a bit too wide. What about making it 100/150 pixels less (or another option to your liking :smile:).

cschramm:
> I do not see any references in the code to store the ip address as an configuration (Config) option. Where is it stored, in Bluez somewhere?

The Network service plugin uses `NetConf.get_default()` which will read from `/var/lib/blueman/network.state`. On apply, mechanism's `EnableNetwork` is used to set the address and also write the data back to the state file. I'm not sure if the file makes sense or if the NetConf class could and should be ported to the GSettings configuration as well.

> Another question, I set the default width for blueman-manager to 750 pixels which I think is a bit too wide.

Yes, noticed this when opening it the first time. The default_width from the manager-main.ui file was 500. I think we should just keep this.

cschramm:
> I pushed a new commit that I'll squash later with the original. I hope It fixes the problems you reported.

Works great :smiley:

infirit:
>I'm not sure if the file makes sense or if the NetConf class could and should be ported to the GSettings configuration as well.

Ah, I see it is a pickle of the whole NetConf object and contains much more than the ip. Looking through it, ip address and ipt rules should be fairly easy to store as strings. Is the purpose of this pickle solely for the ip or am I missing something?

The ipt rules could be a list of tuples (order int and rule string?) with placeholders in the strings for the variable parts like the interface, ip ranges etc. It also comes with the ability for power users to modify them (if this is something the project is willing to support).

Some other candidates for this are the command's + args, pid file name + path and resolve.conf file + path.

cschramm:
I think the relevant parts of the pickle are the members initialized in `__init__`:

````
        self.version = class_id
        self.dhcp_handler = None
        self.ipt_rules = []

        self.ip4_address = None
        self.ip4_mask = None
        self.ip4_changed = False
````

`version` is specific to pickle. Storing `ip4_changed` seems useless (there is a situation where it may be stored as True, when you disable NAP, but it will just be reset to False when the Networking applet plugin is loaded). `ip4_address`, `ip4_mask`, and `ipt_rules` are easy to migrate, as you said. `dhcp_handler` is an object, but I think it can be simplified: It's either a `blueman.main.NetConf.DnsMasqHandler` or a `blueman.main.NetConf.DhcpdHandler` and it looks like all state they have are a pid which does not need to be persisted and a reference to the netconf object which can be initialized when started. So we would just have to store a string or something to indicate which of those two classes shall be used (that's just the `r_dnsmasq` element in the Network service plugin).

Anyway, let's leave this as a second step and get this PR into master. I think all that's left is the squashing, right?

infirit:
> I think all that's left is the squashing, right?

Pretty much with the exception of a little clean-up of the keys in the schema which I just did. I pushed the re-based commits just now.

PR #212, opened 2015-03-06
data: Do not translate polkit vendor

pwithnall:
polkit does not support this and outputs warnings about unrecognised
<vendor> elements in the .policy file.


cschramm:
Thank you!

PR #215, opened 2015-03-06
Blueman now runs under Python 3.3

infirit:
With this blueman runs under python3 but it needs some proper testing in the usual places.

I'll try earlier python3 versions and if they work ok we can relax the requirement.

### DIFF
--- a/apps/blueman-adapters
+++ b/apps/blueman-adapters
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
 import dbus
 from dbus.mainloop.glib import DBusGMainLoop
 from gi.repository import Pango
--- a/apps/blueman-applet
+++ b/apps/blueman-applet
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
 import sys
 import os.path
 import os
--- a/apps/blueman-assistant
+++ b/apps/blueman-assistant
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
 PAGE_INTRO = 0
 PAGE_DEVLIST = 1
 PAGE_PASSKEY = 2
--- a/apps/blueman-browse
+++ b/apps/blueman-browse
@@ -59,7 +59,7 @@ class Browse:
                 d.props.text = _("Failed to launch \"%s\"") % conf["browse-command"].split(' ')[0]
             else:
                 d.props.text = _("Failed to launch default file browser")
-            d.props.secondary_text = "%s\n\n" % unicode(e.message, 'UTF-8') + \
+            d.props.secondary_text = "%s\n\n" % e.message + \
                                      _("You can enter an alternate browser in service settings")
             d.run()
             d.destroy()
--- a/apps/blueman-browse
+++ b/apps/blueman-browse
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
 import os, sys
 
 #support running uninstalled
--- a/apps/blueman-manager
+++ b/apps/blueman-manager
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
 import os.path
 import sys
 import signal
--- a/apps/blueman-mechanism
+++ b/apps/blueman-mechanism
@@ -1,4 +1,10 @@
 #!/usr/bin/env python
+
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
 import sys
 import os
 
--- a/apps/blueman-mechanism
+++ b/apps/blueman-mechanism
@@ -52,8 +52,8 @@ class Tee(object):
     def flush(self):
         pass
 
-
-syslog.openlog("blueman-mechanism", 0, syslog.LOG_DAEMON)
+# Python 2, call str() to convert to a byte string.
+syslog.openlog(str("blueman-mechanism"), 0, syslog.LOG_DAEMON)
 a = Tee()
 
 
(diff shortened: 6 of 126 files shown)


infirit:
Hmm, gentoo does not support earlier versions than 3.3...

cschramm:
Sooo cool. I think we need at least 3.3 as we use the u prefix.

I'm off to the Caribbean now. :grin:

hobarrera:
Thanks, @infirit :beer: 

> I think we need at least 3.3 as we use the u prefix.

Can't we avoid the `u'` prefix by importing `unicode_literals`?

infirit:
>Can't we avoid the u' prefix by importing unicode_literals

All modules import all of ``__future__`` still relevant to Python 2.7 including ``unicode_literals`` :wink:

hobarrera:
Oh, my bad, I thought that `unicode_literals` made the `u'` prefix redundant. I must have misunderstood something! 😅

infirit:
Confusion there is. There is no usage of the `u` prefix because I told the compiler to use ``unicode_literals``. If there is that is a mistake that should be corrected :smile:

cschramm:
The u prefix is in `apps/blueman-sendto` (only).

infirit:
I just pushed an updated branch. But I have problems with the network service (where else).

With 2.7 on applet start-up  I immediately get a dbus error wrt to loading of network settings. It does not give me much to work with and have trouble figuring out where it comes from and why. But, I do not see this with 3.3 :confused: 

````python
on_adapter_property_changed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/DiscvManager.py:96)
prop UUIDs dbus.Array([dbus.String(u'00001200-0000-1000-8000-00805f9b34fb'), dbus.String(u'00001800-0000-1000-8000-00805f9b34fb'), dbus.String(u'00001801-0000-1000-8000-00805f9b34fb'), dbus.String(u'0000110e-0000-1000-8000-00805f9b34fb'), dbus.String(u'0000110c-0000-1000-8000-00805f9b34fb'), dbus.String(u'00001116-0000-1000-8000-00805f9b34fb')], signature=dbus.Signature('s'), variant_level=1) 
ERROR:dbus.proxies:Introspect error on org.blueman.Mechanism:/: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.Spawn.ChildExited: Launch helper exited with unknown return code 1
```

I also have errors in ```on_apply``` but they may very well be related to the above.

cschramm:
The error means mechanism cannot be started.

This seems to be the cause:

````
% python2 apps/blueman-mechanism
  File "apps/blueman-mechanism", line 62, in <module>
    syslog.openlog("blueman-mechanism", 0, syslog.LOG_DAEMON)
TypeError: openlog() argument 1 must be string, not unicode
````

infirit:
Ah, got it thanks. Pushed the latest.

Still a couple of things to check and verify in ``NetConf``, almost got it all sorted :wink:.

jod35fan:
I tried to compile latest blueman sources with python 3.4 that I have in Jessie 64-bit. There is no more "Python.h" header file in dev package, instead it says that:
man python3.4-config:
To build the singe-file c program prog against the python library, use

              gcc $(python-config --cflags --ldflags) progr.cpp -o progr.cpp

       The same in a makefile:

              CFLAGS+=$(shell python-config --cflags)
              LDFLAGS+=$(shell python-config --ldflags)
              all: progr

I did set CF&LDFLAGS like export CFLAGS=$(python3.4-config --cflags) but C code compilation  failed.

Recent Connections plugin does not store anything (the menu item is grayed) when compiled with python 2.7. Can your code  changes affect like that?

If you want bluez5 bluetooth to be on after reboot automatically with bluez5, here are instructions:
https://bbs.archlinux.org/viewtopic.php?id=166362
Create new file /etc/udev/rules.d/10-local.rules
# Set bluetooth power up
ACTION=="add", KERNEL=="hci0", RUN+="/usr/bin/hciconfig hci0 up"

Bluez 5.30 installs some udev rules but when booting you can see that them fails with kernel  
uname -r
3.16.0-4-amd64

cschramm:
What's the status on this? Any ToDos?

infirit:
Yes, in the network service. Mostly the ip bytes(arrays) being passed around. Other than that everything works from my testing but there may be little bytes vs codepoints issues in unexpected places so more testing is needed.

Unfortunately I am not able to do much work on this at the moment.

infirit:
Just re-based against master and fixed a new bytes/codepoints issue in BluezAgent.

infirit:
Ok, I think I got the NetConf and Network stuff sorted now. It now needs lots of testing!

I need some UI changes feedback on tow error dialogs. I ran into the problem that the error dialogs did not contain much information from the exception (only the first line) so I reworked them to have an expander that has the whole exception. I really like to get some feedback on these changes as this is a bigg(ish) UI change. The other option is to ``dprint`` the exception which is less intrusive.

cschramm:
> It now needs lots of testing!

Sound like we should merge it into master and see what happens. :wink: 

Apart from code duplication the code for the exception dialogs looks good to me. :+1:

infirit:
>Apart from code duplication the code for the exception dialogs looks good to me

I put the dialog in a new file/class (see https://github.com/infirit/blueman/commit/fc0744e278ac06a3d09b4e54dbdf8d1e06399594) and used it in the appropriate places. Let me know if this look ok.

If ok, then the scary thing is to merge this into master :fearful:

cschramm:
:smile: Go ahead. I don't expect much issues with Python 2. And the Arch / Gentoo guys will be happy to test the Python 3 support in their automatic packages from git, I guess. :wink:

hobarrera:
Some Arch users follow the latest release (eg: alpha1), so if another alpha is released after this merge, you'll probably increase the amount of python3 users testing it (AFAIK, most other distros still have py2 as the default).

infirit:
Ok, things are merged. 

I am asking @ZeroChaos- to enable python3 in __at least__ the live git ebuild but hoping he will also in the next snapshot :smile:

hobarrera:
> I am asking @ZeroChaos- to enable python3 in at least the live git ebuild but hoping he will also in the next snapshot :smile: 

I have done the equivalent for Arch's `blueman-git`.

cschramm:
Give me two years and the Debian build will use Python 3 too :stuck_out_tongue:

infirit:
Jessie is due (if they make it of course) at the end of the month and it will come with python 3.4 :smiling_imp:

hobarrera:
> Jessie is due (if they make it of course) at the end of the month and it will come with python 3.4 :smiling_imp:

From what I can tell, python2 will *still* be the default python.

cschramm:
Absolutely, yes. Python 3 is available in squeeze and wheezy as well, but it won't be the default anytime soon.

PR #222, opened 2015-03-15
Fix Audio modules import and Recent Connections labels

Plaque-fcc:
After 0aa9344099f0031aef6eba0a059c9cd8fe9303d9 there's no sense in settings props.label.
There are more, I guess, to be fixed, but you guys seems to know more about that than I do.

### DIFF
--- a/blueman/plugins/applet/RecentConns.py
+++ b/blueman/plugins/applet/RecentConns.py
@@ -308,7 +308,7 @@ class RecentConns(AppletPlugin, Gtk.Menu):
             mitem.props.sensitive = True
             mitem.props.tooltip_text = None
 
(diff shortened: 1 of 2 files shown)


infirit:
Thanks, I reworked the label code a bit as that line was getting ridiculously long. Commit pushed to master as 31c129067d9a75d69597529e1e02908d6b093c21.

I'll let @cschramm look at the other bit as he just reworked this whole thing and it is not an area I know much about.

Plaque-fcc:
Thanks! I know I was too lazy to report/fix it ASAP.

Plaque-fcc:
Oh, and I guess there could be a function to set that label, since it was used across the code in many times.

Plaque-fcc:
Oh I first close this request.

PR #224, opened 2015-03-20
Allow to reset alias to the device name

infirit:
After renaming the device there is no way to reset it back to it's default. I opted to add a reset button which sets the ``Alias`` to the ``Name`` of the device.

Another option is to have a button that sets ``device.Name`` in the ``GtkEntry`` text. Or something else you may prefer over either of these two.


cschramm:
It's actually possible but not obvious to the user. Setting an empty alias removes it. I think that's what we should do in response to the reset button. Can you adapt that?

infirit:
Yep, now it uses an empty string to set the ``Alias``.

Issue #220, opened 2015-03-09
missing sendto plugin for caja


raveit65:
sendto plugin for caja is missing.

NiceandGently:
Ok, first i wanna say that bueman is now working in fedora with 1.99.alpha2, so the chances are huge that it will get into fedora 22.
But nemo hasn't a sento plugin , so i changed the title here.
In a test build we renamed the nautilus sendto plugin to caja, but bluemen is build with GTK3 and caja use GTK2.
So can this work with differrent gtk+ versions.
Personaly i don't have a device for testing only a tongle and i don't see anything about bluetooth in caja sendto menu. Is this because of the missing bluetooth device?

cschramm:
Fixed by https://github.com/blueman-project/blueman/commit/768b0d1f138f9ba22d7900044b7f24bcac15925d?

cschramm:
Oh sorry, that one got reverted. @infirit can you comment on this please?

infirit:
It was a hack and did not work, it unfortunately got into my master branch by accident.

I have even tried to see if it worked in Nautilus but it does not show up. Does it work at all with recent Nautilus?

infirit:
The current nautilus sendto extension does not allow for plugins which explains it not showing up.

cschramm:
Hm. Is there another way then? Maybe similar to thunar?

NiceandGently:
Why not renaming the plugin to caja?

cschramm:
Looks like support for nautilus-sendto is not possible anymore. empathy also dropped their plugin:

> The nautilus-sendto plugin has been removed as latest nautilus-sendo
   versions don't support plugins any more.

It is possible to add entries to the context menu by creating scripts in `~/.local/share/nautilus/scripts`, but this a) is user-specific and not possible globally and b) creates the entry in an odd sub menu called `Scripts`.

Renaming and using it for caja-sendto makes sense to me, but we have to make it work. :laughing:

cschramm:
@stefano-k @infirit Should we add the plugin to blueman or caja-sento itself?

cschramm:
Basically, caja-sendto needs to be built against GTK 3, but I think there are more issues. You can see the output of what's going on by running caja-sendto directly in a terminal.

monsta:
Hi @cschramm,

[...]

PR #230, opened 2015-03-28
Rename nautilus-sendto to caja-sendto

cschramm:
Nautilus 3 does not support sendto plugins anymore while caja does.

Also use the external header instead of shipping it.

References #220 

Note that the dependency on the external header is currently not reflected by configure.ac; not sure if it needs to.


infirit:
I pushed a commit makes the build correct. But there is, afaik, an unsolvable problem, see below the traceback.

My thought was instead of using python in the plugin just invoke ```blueman-sendto``` with ```GAppInfo```.

````python
  File "/usr/lib64/python2.7/site-packages/blueman/gui/NstBluetooth.py", line 1, in <module>
    from gi.repository import Gtk, GObject, Pango
  File "/usr/lib64/python2.7/site-packages/gi/__init__.py", line 39, in <module>
    raise ImportError(_static_binding_error)
ImportError: When using gi.repository you must not import static modules like "gobject". Please change all occurrences of "import gobject" to "from gi.repository import GObject". See: https://bugzilla.gnome.org/show_bug.cgi?id=709183
````

cschramm:
The reason for the Python code seems to be to use the `DeviceSelectorWidget` within sendto.

I'm seeing an issue when using the plugin in caja:

````
  File "/usr/lib/python2.7/dist-packages/blueman/gui/NstBluetooth.py", line 1, in <module>
    from gi.repository import Gtk, GObject, Pango
  File "/usr/lib/python2.7/dist-packages/gi/importer.py", line 67, in load_module
    dynamic_module._load()
  File "/usr/lib/python2.7/dist-packages/gi/module.py", line 296, in _load
    self._overrides_module = importlib.import_module('gi.overrides.' + self._namespace)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/usr/lib/python2.7/dist-packages/gi/overrides/Gtk.py", line 1554, in <module>
    initialized, argv = Gtk.init_check(sys.argv)
AttributeError: 'module' object has no attribute 'argv'
````

Cannot make any sense of that... :cry: 

Concerning the gobject issue: Can't we just remove the call to `pygobject_init`? I don't think it's necessary as we're not really using the API, just the PyGObject type.

cschramm:
The reason for the Python code seems to be to use the `DeviceSelectorWidget` within sendto.

I'm seeing an issue when using the plugin in caja:

````
  File "/usr/lib/python2.7/dist-packages/blueman/gui/NstBluetooth.py", line 1, in <module>
    from gi.repository import Gtk, GObject, Pango
  File "/usr/lib/python2.7/dist-packages/gi/importer.py", line 67, in load_module
    dynamic_module._load()
  File "/usr/lib/python2.7/dist-packages/gi/module.py", line 296, in _load
    self._overrides_module = importlib.import_module('gi.overrides.' + self._namespace)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/usr/lib/python2.7/dist-packages/gi/overrides/Gtk.py", line 1554, in <module>
    initialized, argv = Gtk.init_check(sys.argv)
AttributeError: 'module' object has no attribute 'argv'
````

Cannot make any sense of that... :cry: 

Concerning the gobject issue: Can't we just remove the call to `pygobject_init`? I don't think it's necessary as we're not really using the API, just the PyGObject type.

infirit:
I have no clue why you see that error :confused:, was this before or after my commit to fix the build?

I removed the lines for pygobject but it makes no difference.

Reading NstBluetooth, it basically only needs the devicelist to add the spawning of ```blueman-sendto```, but we already get this when we launch ```blueman-sendto``` without the device id anyway. So back to my original suggestion, why not use ```GappInfo``` to launch ```blueman-sendto``` directly :smile:

infirit:
Or just query bluez directly for the devicelist? Perhaps expose this over dbus from blueman?

cschramm:
> I have no clue why you see that error :confused:, was this before or after my commit to fix the build?

Both.

> I removed the lines for pygobject but it makes no difference.

I wonder where pygobject gets imported then. It definitely gets imported in pygobject_init, but if you removed that there must be another location...

> Reading NstBluetooth, it basically only needs the devicelist to add the spawning of blueman-sendto, but we already get this when we launch blueman-sendto without the device id anyway. So back to my original suggestion, why not use GappInfo to launch blueman-sendto directly :smile:

You're right. As blueman-sendto can provide the device list it's probably no big win to have it integrated into the caja-sendto GUI as well. So lets remove the python stuff from the plugin and keep the integration simple.

@stefano-k @infirit Do you want to have the plugin in blueman or caja-extensions? It looks like your preferred way is to ship the plugins with caja-extensions, so I'd just remove the old nautilus-sendto plugin on the blueman side then.

PR #233, opened 2015-03-30
Change agent capability to KeyboardDisplay

cschramm:
@kemnade-uni discovered in [Debian #780776](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=780776#15) that agents get registered with capability DisplayYesNo while it should be KeyboardDisplay.

I wonder why we never had issues with any device reported before because of this and I'm unsure about the implications here, so a couple of testers would be nice. Just comment if pairing your devices behaves as before or if it changes with this and include your BlueZ major version (4 or 5 or both).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 * [ui] Button to reset device alias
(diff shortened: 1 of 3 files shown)


kemnade-uni:
now it works for me with bluez 5.23, passkey dialogs pop up as expected

Issue #235, opened 2015-04-03
Auto-Conecting  Debian Wheezy/Jessie 


rsm-imap-cc:
I'm not sure if this is a bug or not but I wanted to ask the question...

I use this program to connect my speakers. On my system blueman is started by default so it is always running..
When I had Debian Wheezy if I turned on the speakers the connection was made automatically. On Debian Jessie I have to do:

1) Turn on the speaker and set it in paring mode
2) Select the speaker over blueman and set a connection

The speakers are marked as trusted device.

Is there any way of making an auto-conection as in Debian Wheezy?

Thanks!!

cschramm:
You should at least not have to re-pair the device. If you start your system and open blueman, the speaker is not included in the device list and you have to discover it first?

rsm-imap-cc:
The device is in the device list. But i think that the problem is that it doesn't detects that the device is turned on.
When I try to connect it says "Connection Failed: Host is down", Then if I set my speaker in paring mode and I connect, everything works good

cschramm:
Unfortunately that sounds like the change is actually in BlueZ and not in blueman. You could try downgrading to wheezy's bluez package and see what happens.

rsm-imap-cc:
I just found the half of the fix, the problem is that I do not have a "Public" directory.

line 57 of /usr/lib/python2.7/dist-packages/blueman/main/applet/Transfer.py

I sujest to create the directory in case if it doesn't exists. Do you want me to submit the patch?

This only fixes the fact that I can conect without using the paring button. The auto-conect stils not working!

cschramm:
OK, so 1) :white_check_mark:.

Concerning 2) I'd actually expect that behavior. An auto-connection must have been initialized either by the device, which should still work if that's the case, or by some component on your system - especially BlueZ itself or pulseaudio.

I think the best way to find out what changed would be to downgrade blueman, bluez, and pulseaudio to the wheezy versions, check that auto-connecting works, and then upgrade one by one to see which change breaks it. Note that bluez should be upgraded last.

rsm-imap-cc:
Good idea!, I'll check that in a live version, I really miss the auto-connect feature

rsm-imap-cc:
The auto-connection problem is being caused by bluez, if I downgrade everthing works as before :)

cschramm:
Good to know, although unsatisfying for the users... :disappointed: 

As auto-connecting comes up once in a while, this may be a future feature of blueman, but most likely not in 2.0.

[...]

PR #236, opened 2015-04-03
This fixes the half of #235

rsm-imap-cc:
The auto connect stills not working, but the connection success with the "connect button".

### DIFF
--- a/blueman/main/applet/Transfer.py
+++ b/blueman/main/applet/Transfer.py
@@ -46,8 +46,10 @@ class Transfer(OdsManager):
                     self.Config["shared-path"] = os.path.expanduser("~")
                 else:
                     self.Config["shared-path"] = d
-
-            if not os.path.isdir(self.Config["shared-path"]):
+                    
+            if not os.path.exists(self.Config["shared-path"]):
+                os.makedirs(self.Config["shared-path"])
+            elif not os.path.isdir(self.Config["shared-path"]):
                 raise Exception("Configured share directory %s does not exist" % self.Config["shared-path"])
 
             if pattern == "opp":


cschramm:
Thanks for the fix. I think we can remove the elif as it only handles a very unusual edge case now (path exists but is not a directory) and is probably not shown to the user anyway (see below).

Am I right that you - as a user - did not see the exception that is raised if the directory does not exist and only saw it when looking at the applet's output? Then I think we should make sure that the user does get a message if the directory cannot be created.

rsm-imap-cc:
I'm not 100% of having understood your question but, I found the error by launching the applet by the terminal.

I think that you could let the elif. I really daub that some one have a file called "Public" but some times exceptions happens. For example I never use extensions, and I renamed my global dirs to avoid writing caps over the terminal.. !

infirit:
This is not right, the code first ask xdg for the public folder and we should not need to second guess if it exists. If xdg gives a folder that is not created you have a broken setup.

I do think we should show an error dialog if later on for some reason the initially selected folder has gone away.

rsm-imap-cc <notifications@github.com> schreef op 3 april 2015 18:09:22 CEST:
>The auto connect stills not working, but the connection success with
>the "connect button".
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/blueman-project/blueman/pull/236
>
>-- Commit Summary --
>
>  * This fixes the half of #235
>
>-- File Changes --
>
>    M blueman/main/applet/Transfer.py (6)
>
>-- Patch Links --
>
>https://github.com/blueman-project/blueman/pull/236.patch
>https://github.com/blueman-project/blueman/pull/236.diff
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/blueman-project/blueman/pull/236

-- 

cschramm:
I agree. I think it's best to just display a localized version of the message that is currently implemented as an exception to the user.

infirit:
I have a branch on my fork that adds a message dialog that warn the user, see https://github.com/infirit/blueman/commit/7e00f5eb7ace1fdd4bb7d0226229f5faa916b1a1. Important is that I still raise the Exception as we do not want to continue with the broken shared-path.

Let me know if this is acceptable and I'll push it into master.

cschramm:
Looks good to me.

infirit:
Thx, pushed and closing this one.

Thanks @rsm-imap-cc for reporting this problem.

PR #238, opened 2015-04-04
Implement udhcpd support

### LINKED ISSUE #237: Could blueman be made to work with udhcpc
Just a suggestion, but could blueman be made to work with udhcpc (busybox) for network connections?

cschramm:
Also: Fix bug where dhcpd handler would remove all content of the configuration file after the blueman section.
Also: Remove obsolete NetworkSetup mechanism method.

~~Closes #237~~

The udhcpd support is not yet tested. @juanitotc Please give it a try.

I'm having a small issue with checking if udhcpd ran successfully: I could either start it in foreground using the -f option, which would return with a non-zero exit code on failure, but not return at all if started successfully. Or I could start it in background, but then I always get exit code 0 and cannot detect if there was an error. In the current state, udhcp would fail silently. The only trace are the error messages in syslog.


juanitotc:
Which branch do I apply the patch to?

cschramm:
The commit for this PR is in the udhcpd branch (see the line below the title: *cschramm wants to merge 1 commit into master from udhcpd*).

juanitotc:
I compiled the uphcpc branch and gave it a go, but things didn't work so far.

There are two instances of udhcpc already running on my system:

`$ ps aux | grep udhcpc
 2788 root     /sbin/udhcpc -b -i eth0 -x hostname boxdell -p /var/run/udhcpc.eth0.pid
 3300 root     udhcpc -b -i wlan0 -x hostname boxdell -p /var/run/udhcpc.wlan0.pid`

..I wonder if I need to create a third instance for bnep0 before trying to make the connection?

`$ blueman-applet
...
notify (/usr/local/lib/python2.7/site-packages/blueman/plugins/applet/RecentConns.py:235)
<Device object at 0x7f1edca6c7d0 (blueman+main+Device+Device at 0xe45b80)> <blueman.services.NetworkAccessPoint.NetworkAccessPoint object at 0x7f1edcaee0d0> 00001116-0000-1000-8000-00805f9b34fb 
_________
initialize (/usr/local/lib/python2.7/site-packages/blueman/plugins/applet/RecentConns.py:140)
rebuilding menu 
_________
__init__ (/usr/local/lib/python2.7/site-packages/blueman/main/Device.py:21)
caching initial properties 
_________
__del__ (/usr/local/lib/python2.7/site-packages/blueman/main/Device.py:65)
deleting device /org/bluez/hci0/dev_54_E4_3A_22_48_E9 
_________
Destroy (/usr/local/lib/python2.7/site-packages/blueman/main/Device.py:87)
invalidating device /org/bluez/hci0/dev_54_E4_3A_22_48_E9 
_________
err (/usr/local/lib/python2.7/site-packages/blueman/plugins/applet/DhcpClient.py:49)
org.freedesktop.DBus.Python.Exception: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/blueman/plugins/mechanism/Network.py", line 31, in DhcpClient
    dh.Connect()
  File "/usr/local/lib/python2.7/site-packages/blueman/main/DhcpClient.py", line 39, in Connect
    raise Exception("no DHCP client found, please install dhcpcd or dhclient")
Exception: no DHCP client found, please install dhcpcd or dhclient`

juanitotc:
I'm unable to start udhcpc before attempting to set the iPhone as a network access point:

`$ sudo udhcpc -b -i bnep0 -x hostname boxdell -p /var/run/udhcpc.bnep0.pid
udhcpc: SIOCGIFINDEX: No such device`

..but if I set the iPhone as a network access point and then start udhcpc, the error above has already occurred in blueman-applet and:

`$ sudo udhcpc -b -i bnep0 -x hostname boxdell -p /var/run/udhcpc.bnep0.pid
udhcpc (v1.23.1) started
Sending discover...
Sending discover...
Read error: Network is down, reopening socket
udhcpc: bind: No such device`

Does this require wpa_supplicant?

cschramm:
This PR actually brings udhcpD support when offering a PAN and does of course NOT close #237. udhcpc support is in #244. Works for me.

infirit:
This conflicts with #215 so one will have breakage and needs to rebase.

cschramm:
Done.

infirit:
The ui now recommends both dnsmasq and udhcpd, is this correct? 

ps: I pushed a commit to solve a missing import of UdhcpdHandler.

cschramm:
Recommends in what way?

infirit:
The [tooltip](https://github.com/blueman-project/blueman/pull/238/files#diff-bd2ffa186398195e8dfdc4d0a00d977bR150) :smile:

cschramm:
No, that was not intended. :wink:

infirit:
Does this work or does it need more testing? I am poking at this atm and I have the busybox binary (I only need to add the relevant symlink I think) installed so let me know :smile:.

cschramm:
The issue in my initial comment still applies: I cannot really check if it was started successfully or not...

infirit:
We can use ``Popen.pid`` to get the pid from the child process and check it with ``is_running``. I think that should work. The returncode is indeed always 0 which is not helpful :disappointed:

infirit:
Just did a quick test script and it works. ``Popen.pid`` will point to the running udhcpd process so this will sort out that issue :smile:

infirit:
I rebased against master and pushed it to my fork [1]. The problem is that subprocess seems to stall/hang, not sure why yet.

[1] https://github.com/blueman-project/blueman/compare/master...infirit:udhcpd?expand=1

infirit:
Ok, figured out what is going on. The simple reason is that the ``-f`` option is added which means run in the foreground so never forking in the background. Was there a specific reason for running in the foreground?

One thing I noticed is that my udhcpd does not handle ipv6 dns entries (my router is fully running ipv4 + ipv6).

But, after filtering ipv6 bits from dns and removing the ``-f`` option it works for me. I updated my fork with these changes.

cschramm:
As mentioned in my initial post `-f` provided me with a non-zero return code on failures, but on the other hand stalled, so it was not useful. :wink:

infirit:
>As mentioned in my initial post

Oops, I am sure I got glasses last year, oh well :smile_cat: 

I'll clean things up and make sure everything works properly and report back here. Just a quick question, should I squash the whole thing into a single commit with you as author or keep it as it is in my fork?

cschramm:
No need to keep me as the author.

infirit:
Here goes, https://github.com/blueman-project/blueman/pull/402

PR #246, opened 2015-04-08
Add configuration option to force notification fallback

### LINKED ISSUE #245: Add build configuration to force blueman integrated notification fallback
Ref;

`...using flwm the pairing pin request pop-up buttons (confirm, deny) are almost invisible unless I roll the mouse pointer over them.`

and:

`That is an unfortunate theming issue, but for convenience we could offer a build configuration option to force blueman to use its own GTK+ window for notifications with actions even if the notification daemon is capable of showing actions. If you're interested, please open a new issue for that. The implementation shouldn't be much more than two lines.`

If you could do that, it would be much appreciated - as well as being able to see the buttons, it would avoid having notification-daemon as a dependency and having to start it with the X-session.

cschramm:
~~Using --disable-notification-daemon~~ the user can now choose not to use a notification-daemon at all. Previously the fallback GTK+ notifications were only used if we had to show actions, but the daemon did not report to support them.
Additional changes:

* Fix error message when libnotify cannot be initialized (it is no longer pynotify)
* ~~Fix blueman.Constants.POLKIT (was always false since 1954e2d3ffbc79a0cf696c7fe12bca431d34d368)~~
* Fix issue from GTK+ 3 transition in NotificationDialog (window vs. get_window())

Closes #245
Closes #246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [ui] Button to reset device alias
 * [bluez] Agent capability KeyboardDisplay (@kemnade-uni / Andreas Kemnade)
+* New configuration option to disable the use of a notification daemon
 
 ### Bugs fixed
 
--- a/apps/blueman-applet
+++ b/apps/blueman-applet
@@ -45,7 +45,7 @@ class BluemanApplet(object):
     def __init__(self):
         setup_icon_path()
         if not Notify.init("Blueman"):
-            dprint("Error: Failed to init pynotify")
+            dprint("Error: Failed to initialize libnotify")
 
         check_single_instance("blueman-applet")
 
--- a/blueman/gui/Notification.py
+++ b/blueman/gui/Notification.py
@@ -1,3 +1,4 @@
+from blueman.main.Config import Config
 from gi.repository import Notify
 from gi.repository import Gtk
 from gi.repository import Gdk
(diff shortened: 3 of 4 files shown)


infirit:
Why not make it a config option in org.blueman.general schema? That way folks can set it once per user or install a gsettings override file to set it for all users.

cschramm:
I like the idea. Didn't know of override files. :smiley:

cschramm:
Hacked the gsettings part together but did not have the time to test it yet.

infirit:
Had some time to test this and it works nicely.

Not related to this change but, the button icons are both starts when pairing my phone. I am not sure how this was before but it looked a bit weird.

cschramm:
> the button icons are both starts

Is it just me vor does this sentence not make sense? :wink:

infirit:
Oopsy, __stars__ :grin:

cschramm:
I think that's help-about as defined in blueman/main/applet/BluezAgent.py. Looks like that icon is everywhere. :laughing:

infirit:
I see this was done in the big icon update. They where gtk-yes and gtk-no before which do not have a direct replacement according to the Gtk3 docs. I would suggest we find some red and green icons that could replace these. Not sure which would fit this but perhaps dialog-error for the decline?

infirit:
``window-close`` seems a nice fit for gtk-no, but I have a hard time finding something for gtk-yes :cry:

cschramm:
We can also just remove the icons from those notification actions. They are not used for notifcation daemons as well, so I do not see a reason to have them in the fallback.

infirit:
Removed in https://github.com/blueman-project/blueman/commit/495018ae3e7441daf011437924f4787909456839.

cschramm:
:+1:

Removed support and the usage on incoming transfers in f214a05.

PR #256, opened 2015-04-14
Use GAppInfo to launch (gui) application

infirit:
My goal is to get rid of all the custom startup notification code in the cython module. Startup notification is now well integrated in GLib and Gtk so there should be no need to call into X libs directly. Also at some point we may not run under X anymore but wayland.

Currently all places where gui applications are launched the new GAppInfo ``launch`` function is used. There is still the ```SerialManager``` plugin which does a little more than just spawning.

Then we have the ```RecenConns``` plugin where startup notification is used outside of ```spawn``` to signal the connection attempt. I'll need to figure something out for this.

Work in progress still but like some feedback :smile:.


cschramm:
Sounds great. :smiley_cat:

cschramm:
I'm not sure if we need to be a little more cautious with the command arguments especially in ManagerDeviceList and Transfer where we include arbitrary file paths. We probably need to add quotes and escape correctly, don't we?

cschramm:
I removed NstBluetooth in 25c4d12311152e8cddb047a8dd29b173ee1105b8, so SerialManager is the single left usage of the spawn method. I think in RecentConns startup_notification is completely misused and should just be replaced with Notification(). That eliminates the spawn and the startup_notification functions completely. :smiley:

infirit:
Ok, I have been poking GAppInfo and you are correct in that we need to worry about paths with spaces. I'll rework this so it uses ``GFile`` objects which on creation will handle this transparently.

On ``RecentConns``, I agree it is misused. It shows a notification if the connection failed but nothing on success which I'll add.

infirit:
Ok, fixed up the handling of paths and removed the custom startup notification code from the cython module. Just SerialManager left now.

RecentConns already has a success notification but I never saw it as it locks up the applet. Whenever I try to connect to my wii u controller it locks the applet (no console output from then on) and I have to restart it. It still connects it though and as long as it is connected it does show my the notification when I click on the menu entry. Do you see the same when using the RecentConns plugin?

cschramm:
I'll test what I see with RecentConns.

I think in SerialManager we can just replace the line with `p = Popen(args, preexec_fn=lambda: os.setpgid(0, 0))` to get rid of the spawn method.

But, in every case where we replaced `spawn` with `launch` or `Popen` we lost some functionality that was present in `spawn`: Depending on the `system` parameter it either prepended `BIN_DIR` or expanded the user directory in the command. I think we should add this to launch as well to stay compatible.

Additionally, `spawn` set the environment variable BLUEMAN_EVENT_TIME. The value is passed to `present_with_time` on the window that displays the information that an instance is already running (for adapters, manager, and services). As I don't know what the present_with_time timestamp actually does, I do not know if that matters or not, though. :wink:

cschramm:
When done, the startup-notification changes need to be reflected in configure.ac, README.md, and CHANGELOG.md.

infirit:
I have no clue why the event time is written to the lock file. It does not look like it is used anywhere, but as we both are not sure I'll add this back.

I understand BIN_DIR is used to run blueman from within git and I'll look into it. But, can you explain how BIN_DIR is supposed to work when ```system=True```. For example: ```spawn(["xdg-open", path], True)``` in ```Transfer.py```.

If I follow this in spawn ```xdg-open``` is replaced with ```os.path.expanduser("xdg-open")``` which does absolutely nothing :confused:.

cschramm:
You're right, it does not make sense for xdg-open. Adding BIN_DIR makes sense for blueman-* commands and expanduser makes sense for the user supplied commands in blueman-browse and Headset. I would not mind losing that expand feature, though, but BIN_DIR is great. Instead of an additional parameter you could also just check if the command starts with blueman-.

> I have no clue why the event time is written to the lock file. It does not look like it is used anywhere

It is. What happens when `check_single_instance` detects a running instance is it writes the value of BLUEMAN_EVENT_TIME into the lockfile, then signals SIGUSR1 to the running instance. The handler for that signal is defined in `check_single_instance` itself. It reads the value from the lockfile and passes it to the unhide_func provided to `check_single_instance`. This is use by blueman-adapters, blueman-manager, and blueman-services to present a window with a timestamp (the given value). But at the end of the day I have no idea what all that fuss is for as I do not know the meaning of the timestamp to the GtkWindow. Do you?

infirit:
I have added the system argument already, not much work and it is more clear. But I am very confused why ```os.path.expanduser``` was used in spawn. It does not work on __any__ command and will only retrieve the path to a user's homedir when given ``~``.

Below is what the docs say about ```gtk_window_present ``` which is what actually uses the timestamp in the end. The ``with_time`` addition I suspect deals with focus handling/stealing. I have added the timestamp handling to the launch function already.

>Presents a window to the user. This may mean raising the window in the stacking order, deiconifying it, moving it to the current desktop, and/or giving it the keyboard focus, possibly dependent on the user’s platform, window manager, and preferences

cschramm:
> It does not work on any command and will only retrieve the path to a user's homedir when given ~.

Right. `spawn` is used for three types of commands: blueman apps, xdg-open, and user supplied commands. The later ones can be specified in the settings for the Headset and SerialManager plugins and for the local Transfer service. The user may currently use ~ or ~someuser there. But as I doubt that anybody ever used this, we can drop it and require him to expand it himself. The blueman settings are user-specific anyway, so ~ always expands to the same user directory (unless multiple user accounts share the same .config/dconf, but, yeah...).

infirit:
>we can drop it and require him to expand it himself

I was not arguing against it, I just did not understand the use case. I only ever used expanduser to get the user homedir so I never imaged it could be used like this :smile:. I have added this now and pushed the latest changes. It should be ok now.

cschramm:
Looks good to me. :smiley:

infirit:
Merged

Issue #251, opened 2015-04-13
Code quality according to pylint


infirit:
I ran pylint (no options) on the blueman and it came with a **very** large list of things it is not happy about.

Some we can ignore like the gettext underscore but it does come up with some real problems.

Now, my question is, can we come up with some rules on code quality? For example, enforce the line lenght of 100 or enforce the requirement of docstrings?

cschramm:
I don't have much Python experience, so I don't know the tools, but love such things (see [my rails-audit ruby gem](https://github.com/cschramm/rails-audit)). I'm also working on some CI for blueman and will be happy to integrate them there. :smiley: 

> can we come up with some rules on code quality?

Of course!

> enforce the line lenght of 100

I'm generally using 120, but there are situations with long (interpolated) strings where I actually violate that due to no good solution. :unamused: 

> enforce the requirement of docstrings?

Not sure. I'm an advocate of the theory that extensive documentation leads to bad naming and overly complex code and that it's better to try to write "self-documenting code". Excessive documentation can also contribute to unreadable code and unmaintained documentation can even lead to confusion (left over comments, but the code changed significantly). Anyway, if an enforcement of docstrings is the consensus, I'm on.

Is there something like the [Ruby style guide](https://github.com/bbatsov/ruby-style-guide) for Python? I know there are PEPs on code style, but do they suffice as a "rule book"? What rules is pylint using (without options)? Are they from PEPs or some other document?

infirit:
>Is there something like the Ruby style guide for Python?

[PEP 8](https://www.python.org/dev/peps/pep-0008/) is the go to place for most of these things.

[Pylint](http://www.pylint.org/) is the standard code analysis tool and integrates well in several ide's. It checks many of the points included in PEP 8 but can be modified to check for 120 characters. It checks style and looks for bad practices in the code. Run it and you'll get a sense of it.

As for docstrings, it is not meant to be a replacement for proper documentation. It is a short description of what the function or class __does__. It is then used by tools like ide's to show in the popup when using a function. It also shows up when using the built-in help function. See [this](http://www.pythonforbeginners.com/basics/python-docstrings) for a short introduction and examples. Another use of docstrings is for test ([doctest](https://docs.python.org/3.3/library/doctest.html)). It is very useful and this is what ppl mean when they say Python is self documenting :smile_cat: 

[...]

Issue #43, opened 2014-04-19
Tests


cschramm:
Long way to go, but sooo needed...

cschramm:
I'd really love to add some regression tests when fixing some stuff, but the fundamental thing we need to get started with this is some way to simulate bluetooth devices. Maybe there's some way to fake them, maybe even some official BlueZ tool. Otherwise we'll need to mock the org.bluez bus, but that will avoid testing the actual interaction of course...

Any ideas?

infirit:
disclaimer, I never done anything with tests, but I always found the topic interesting. 

I found [this](https://www.youtube.com/watch?v=FxSsnHeWQBY) very interesting. A follow up I have not watched yet is [this](https://www.youtube.com/watch?v=ArLf03-RLwU) one.

Basically they suggest to rewrite the functions so they can operate without the needing external input like real bluetooth devices.

leigh123linux:
@cschramm 

I think you could use python-dbusmock for testing, it already has bluez4 and 5 templates

http://www.piware.de/2012/11/python-dbusmock-templates/

cschramm:
Looks promising. Project home is [here](https://github.com/martinpitt/python-dbusmock). Not sure if the BlueZ templates are complete and up-to-date, but we could send some PRs if we need something else. :smiley:

cschramm:
OK, I experimented with dbusmock a little bit and created the [tests branch](https://github.com/blueman-project/blueman/tree/tests). I'm experiencing strange issues with the dbus socket. Could someone please try running `PYTHONPATH=. python blueman/test/plugins/applet/SerialManagerTest.py` on the branch to see if he gets "Connection refused" errors from the applet child process as well? It looks like dbus-python creates an abstract socket which dbus fails to connect to. I can connect to the socket from another python process though...

infirit:
I do see ``Connection Refused`` and also noticed ``blueman-applet`` fails to start with ``maximum recursion depth exceeded``.

PR #261, opened 2015-04-14
Travis

cschramm:
I've set up a travis configuration for blueman. It currently basically just ensures that the build succeeds, but I want to add analysis as discussed in #251 and tests (#43; a very first step for testing would be to just start our apps and see if the keep running for a couple if seconds before killing them).

I did not manage to get Python 3.3 and 3.4 passing on travis yet. When setting the PYTHON variable it gives `checking whether python3.3 version >= 2.7... configure: error: too old` which I guess means the name python3.3 is not available (or just some dummy script). Using `language: python` and the python instead of the env matrix I couldn't get ./configure to find python-dbus, neither for Python 2, nor 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 script:
   - ./autogen.sh
-  - make check
   - make
+  - make check
   - cd po && intltool-update -m && [ ! -f missing ]
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+script:
+  - ./autogen.sh
+  - make check
+  - make
+
+env:
+  - PYTHON=python2.7
+  - PYTHON=python3.2
+
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq automake autoconf libglib2.0-dev libtool libgtk-3-dev libstartup-notification0-dev python-gi-dev libbluetooth-dev intltool python-dev python3-dev cython python-dbus python3-dbus dh-autoreconf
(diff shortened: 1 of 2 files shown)


infirit:
Looking at the docs you are supposed to add the below to indicate which python versions should be available.

Also, we should run make check after make I think or it will not be able to find certain files.

```
python:
  - "2.7"
  - "3.2"
  - "3.3"
  - "3.4"
  - "nightly"
```

cschramm:
Using the python matrix switches our language to python (it's c otherwise) which means travis will use virtualenvs. And in those virtualenvs no dbus-python binding is available. Or at least I could not find out how I can get it.

Sample:
https://travis-ci.org/blueman-project/blueman/builds/58515534
https://github.com/blueman-project/blueman/commit/588057e07303f4eb8d2d62b8f5e09d0a88997dc2

infirit:
What about setting language to c and then define ``python``? Also try to install the deps in ``before_script``, I suspect it is too late whith ``before_install``.

cschramm:
> What about setting language to c and then define python?

I think the python matrix has no effect in that case.

> Also try to install the deps in before_script, I suspect it is too late whith before_install.

No, that works. In the above sample apt-get starts on line 84 of the output and autogen - that's what fails as it does not find dbus - on line 210.

cschramm:
Merged. Python 2.7 and 3.2 is better than nothing. We can figure out how to get 3.3 and 3.4 running later.

Green :white_check_mark: s everywhere from now! :laughing:

infirit:
I just saw you added an intltool missing check after make check. However make check already does this, look at the end of make check's output :smile:

Issue #69, opened 2014-07-05
feature request: accept all files (obex push)


treaki:
Hi,

i would appreciate if you could add an option to accept all files send to the computer by obex push, not only them send from trusted devices. Therefore the files should be saved in a folder structure like this:
bluman_recived_files
- recive.log
- XX:XX:XX:XX:XX:XX (bluetooth id)
  - date_filename.ext

thanks

infirit:
I understand the need to unconditionally accept files. And I think doing a dir/file formatting language that can be customized by the user is needed. But..

Can you explain the use case for the receive log?

treaki:
hi infirit, to integrate it into  an "kind of public accessible interactive kiosk system" where people can drop files using bluetooth. Another daemon is reading the log file created by the obex push service, like tail -f, and is doing some stuff if events take place...

but in general i would like that every pice of software running on my box is generating logs at any relevant event that i can use for analyzing the system history. Nothing as important as receiving files using bluetooth should be unnoticeable and if we have a unintended receive mode we need a log file (in my opinion)...

regards treaki

cschramm:
Closing as 8 years have passed without any progress, this seems like a rather specific use case to me and I don't really see it happening to be honest.

PR #263, opened 2015-04-19
obexd

cschramm:
This is not (yet) a request to pull the changes, but only meant to discuss the ongoing development.

@infirit I found out what's going wrong when writing to the file, although the implementation is a little strange. After open() succeeded and created or truncated the file, obexd checks if it resided within its root path. The root path is an option to obexd and specified either absolute or relative to the user's home directory and it defaults to `g_build_filename(g_get_user_cache_dir(), "obexd", NULL)` which will probably be $HOME/.cache/obexd. If the shared path configured in blueman is different, receiving files will fail. I'll look into how to get the configuration from obexd - if possible at all.

For those curious: Please note that I'm currently focusing on BlueZ 5 only. BlueZ 4 with obexd 0.47 or 0.48 will not work yet, but of course you can provide tracebacks for that so that we know which parts we have to care about.


infirit:
If the open() call truncates existing files maybe we are not supposed to write directly to the destination. 

Looking at how gnome-user-share it looks like it does this, first [obex_agen_authorize_push](https://git.gnome.org/browse/gnome-user-share/tree/src/obexpush.c#n599) and then [move_temp_filename](https://git.gnome.org/browse/gnome-user-share/tree/src/obexpush.c#n471).

infirit:
Reading some more, [transfer_property_changed](https://git.gnome.org/browse/gnome-user-share/tree/src/obexpush.c#n574) watches for the ``complete`` status and then moves the file to its destination. So the logic looks like, create a tempfile in obexd's cache dir, store the original filename and move to the destination with the original filename on complete status.

cschramm:
Already working locally. Will push in a minute and go to sleep. :wink:

cschramm:
BlueZ 5 support looks pretty good now. :smiley:

infirit:
Did a quick test and it works for me now \o/.

infirit:
I pushed two commits, the important one is a first attempt to fix python3, Python2 is fine after these. 

When I ran this branch under Python3 it fails. I do not see blueman respond at all to the request. The only thing I see is that it connects to the phone, then disconnects. See below what obexd shows in the log.

Also ``shutils.move`` does not care if the file already exists and overwrites. Not sure where we need to check for this, before the transfer starts? But then how should this be handled, should we add some random character (unix timestamp?) to the filename or ask the user what to do (with an option to rename perhaps)?

```
Apr 20 16:41:56 infirit obexd[3579]: CONNECT(0x0), (null)(0xffffffff)
Apr 20 16:41:56 infirit obexd[3579]: CONNECT(0x0), (null)(0x0)
Apr 20 16:41:57 infirit obexd[3579]: PUT(0x2), (null)(0xffffffff)
Apr 20 16:41:57 infirit obexd[3579]: PUT(0x2), FORBIDDEN(0x43)
Apr 20 16:41:57 infirit obexd[3579]: DISCONNECT(0x1), (null)(0xffffffff)
Apr 20 16:41:57 infirit obexd[3579]: DISCONNECT(0x1), SUCCESS(0x20)
Apr 20 16:41:57 infirit obexd[3579]: disconnected: Transport got disconnected
Apr 20 16:41:57 infirit bluetoothd[2839]: Unable to get io data for Object Push: getpeername: Transport endpoint is not connected (107)
```

cschramm:
> But then how should this be handled, should we add some random character (unix timestamp?) to the filename

Probably a good idea when looking at #69.

> When I ran this branch under Python3 it fails. I do not see blueman respond at all to the request.

:open_mouth: Strange. Must be some suppressed exception, I guess. I'll see if I can reproduce that.

cschramm:
Looks like it does not even load the TransferService due to the use of ClassType. Please try a027a5c.

infirit:
Yep, works now under python3 as well :+1:

cschramm:
Just to point out the current state of this migration: I consider BlueZ 5 support done and hope I can do the BlueZ 4 implementation until the end of the week.

cschramm:
And finally here it is. No more blockers for 2.0 (we need to take care of #266, though)!

I implemented support for BlueZ 4 stand-alone obexd (0.47 and 0.48), fixed https://github.com/blueman-project/blueman/commit/8a18d882c0f8d125bb741d7a95599a568494b438#commitcomment-11109404, cherry-picked https://github.com/blueman-project/blueman/issues/275#issuecomment-99935715, and changed the wording and defaults as discussed in https://github.com/blueman-project/blueman/issues/275#issuecomment-99937842.

There's [one TODO](https://github.com/blueman-project/blueman/blob/b69bca36f8616965739e6242ae6a89f69711f38f/blueman/plugins/applet/TransferService.py#L63) left: I don't know yet how to get the root path for incoming files in obexd 0.x. `/tmp` works for my setup, but our `shared-path` produces an access violation, so I guess `/tmp` is just the default if no root path is specified.

cschramm:
Update on the TODO: We do not really need the root path at the position of the TODO as obexd 0.x accepts relative file names in the agent reply, so we can give it just the filename (although the docs state the opposite). The problem is then that we cannot find out the actual full path when doing the `shutil.move`.

If it turns out that we cannot resolve this, I guess it's best to just advise the user / maintainer not to set a custom root path for obexd for our integration to work. That's fair enough for supporting legacy and obviously unfinished software, I think.

cschramm:
Squashed and rebased on master. Any objections against getting it into master?

I'll stick with expecting /tmp as the root path of standalone obexd for now. Will point that out on https://github.com/blueman-project/blueman/wiki/Integrations#obex-file-transfer and in the release notes.

infirit:
Just one thing that is essential is to check if the file exists and not overwrite blindly.

Other than that, sound like a good plan to merge and have more ppl using and testing.

infirit:
https://github.com/infirit/blueman/commit/46316f4a34827d272cf852a29852bd57d245c874 just prefixes the destination filename with date and time. For now this is enough for ppl not to accidentally overwrite existing files.

cschramm:
Totally forgot about this. Updated the integrations wiki page and the release notes of 2.0.rc1 to include a note on the root path issue.

PR #265, opened 2015-04-20
PulseAudio python3 support

infirit:
I saw this was broken under python3 and made it work on par with current master. Except for the signal bars not showing under python3. But.... the ui has some problems in master __and__ my python3 branch with __both__ python versions.

What I see is that the menus get confused in that the ```active_profile``` in the menus do not correspond to the actual profile selected. The device menu works better than the right click one.

Work in progress still but my eyes can not handle more for today.

### DIFF
--- a/blueman/main/PulseAudioUtils.py
+++ b/blueman/main/PulseAudioUtils.py
@@ -363,10 +363,10 @@ class PulseAudioUtils(GObject.GObject):
 
     def __get_proplist(self, proplist):
         if proplist:
-            pla = pa_proplist_to_string_sep(proplist, "|")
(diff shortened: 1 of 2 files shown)


infirit:
https://github.com/infirit/blueman/commit/71651a3ee0d32746e6c18b55e4bee7c0d91b1e71 fixes the problem with the menus not updating properly.

cschramm:
Looks good, but we need to add module-bluez4-device.c to the drivers as well.

infirit:
Ok, updated the list with the bluez4 driver.

I noticed the signal bars are missing on __all__ connected devices under Python3 so I'll push this to master. I'll open an issue for the signal bars to make sure we keep track of this.

PR #269, opened 2015-04-22
Gtk, GLib, Gio and Gthread are runtime only deps.

infirit:
The PR to start a quick discussion :smile:.

There is nothing in libblueman or _blueman.pyx that requires the above mentioned libs at build time and should only be available at runtime. And the runtime deps are the introspection (gir/typelib) files not the -dev packages with the pkg-config and header files.

We could print a nice warning at the end of configure to alert the maintainer to make sure the introspection files are added as build deps. Or something else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * [nm] Enable IPv6 on created connections
 * Use GAppInfo to launch applications
 * [bluez] Auto power on adapters; can be disabled in PowerManager settings
+* Remove Gtk+, GLib and Gio as build time dependancies.
 
 ### Bugs fixed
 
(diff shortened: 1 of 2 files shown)


cschramm:
The change makes sense. I guess the package check is left over from the pre-gi days and / or the nautilus sendto plugin.

> We could print a nice warning at the end of configure to alert the maintainer to make sure the introspection files are added as build deps. Or something else?

I don't think that's necessary as we maintain a nice Dependencies.md file referenced from the README.md.

In any case we should mention in the Changelog that those unnecessary build dependencies were removed from the build script so maintainers can remove them.

By the way: Do we really use GThread?

infirit:
The sendto plugin and the startup notification in _blueman.pyx where using these before, both got removed.

>Do we really use GThread

Good you mention this as I forgot to check modem-prober as it uses threads. I updated the PR with this fixed and also fixed the pygobject check.

cschramm:
Looks good to me. :+1:

PR #271, opened 2015-04-28
Check for supported python3 versions at configure

infirit:
This should work fine but another set of eyes on it to make sure I am not forgetting something :smile:


cschramm:
:+1:

PR #279, opened 2015-05-10
Set a process title for the blueman apps

infirit:
I was getting sick of trying to figure out which python process was blueman-manager, applet etc etc.

It makes use of the c module setproctitle but one could also use ctypes and call into libc directly. Although it uses a 3rd party module it made more sense to me than reinventing the wheel with libc+ctypes.

Opinions?

### DIFF
--- a/apps/blueman-adapters
+++ b/apps/blueman-adapters
@@ -19,7 +19,7 @@ if os.path.exists(os.path.join(_dirname, "CHANGELOG.md")):
 
 import blueman.bluez as Bluez
 from blueman.Constants import *
-from blueman.Functions import setup_icon_path, enable_rgba_colormap, check_single_instance, check_bluetooth_status
+from blueman.Functions import *
 from blueman.main.SignalTracker import SignalTracker
 
 enable_rgba_colormap()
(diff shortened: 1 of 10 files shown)


cschramm:
Never had an issue with this as my standard ps options are `-e f` and `f` includes the arguments.

I don't really like to require an additional dependency for such a small change, but well, if you need it... Anybody else wanting this?

We need to add the library to Dependencies.md if we go that way.

infirit:
Most users will have a hard time figuring this out with their gui task managers. Also I currently can't ``killall blueman-applet`` because the process name is pythonX.Y /some/path. Can you?

The ctypes code is fairly simple but it only works on linux. I updated the PR accordingly so I'll leave it up to you which you prefer.

cschramm:
> Also I currently can't killall blueman-applet because the process name is pythonX.Y. Can you?

Yes, I can. Interesting. I actually ever wondered why that works. Maybe Debian's killall got some magic. A broken killall is definitely a better argument than ps.

> The ctypes code is fairly simple but it only works on linux. I updated the PR accordingly so I'll leave it up to you which you prefer.

:+1: 

If there are no objections, we can merge that.

infirit:
>If there are no objections, we can merge that

Pushed slightly modified (print a warning on failure).

PR #280, opened 2015-05-10
Fix some pylint errors

cschramm:
I think I fixed all actual errors, plus some false positives.

CellRendererPixbufTable and PixbufTable are not actually used anymore.

False positives left:
* [ ] Undefined variable 'ngettext'. @infirit? :smile: 
* [ ] Instance of 'int' has no 'foobar' member for virtually all usages of `self.props`. I don't know why pylint thinks `props` is `int`...
* [ ] In SerialManager's `exit_cleanup` function, pylint thinks `SerialManager.__instance__` is `BasePlugin` while it's actually `SerialManager`.
* [ ] Module 'cairo' has no 'OPERATOR_OVER' member. Well, it has...

I ignored issues in the module `blueman.Lib` as pylint obviously does not detect any of the module members. I did not try to use `_blueman` instead. Not sure why the `blueman.Lib` module exists at all...

Goal here is to eliminate all errors and add pylint -E to the travis script. :smiley: 

Edit: Just to point it out: This is currently based on the obexd branch which is subject to change as there are some errors in the now removed obex-data-server code.


infirit:
We add ngettext to builtins (like _ is by gettext) so we should add it to additional-builtins in pylintrc. But we should look why we do not use an underscore which is already in that list.

cschramm:
:point_right: #283

PR #283, opened 2015-05-12
Fix some pylint errors

cschramm:
I think I fixed all actual errors, plus some false positives.

CellRendererPixbufTable and PixbufTable are not actually used anymore.

False positives left:
* [x] Undefined variable 'ngettext'. Can be replaced with _? Should be ignored otherwise.
* [ ] Instance of 'int' has no 'foobar' member for virtually all usages of `self.props`. I don't know why pylint thinks `props` is `int`...
* [ ] In SerialManager's `exit_cleanup` function, pylint thinks `SerialManager.__instance__` is `BasePlugin` while it's actually `SerialManager`.
* [x] Module 'cairo' has no 'OPERATOR_OVER' member. Well, it has...

I ignored issues in the module `blueman.Lib` as pylint obviously does not detect any of the module members. I did not try to use `_blueman` instead. Not sure why the `blueman.Lib` module exists at all...

Goal here is to eliminate all errors and add pylint -E to the travis script. :smiley:


infirit:
Let's start with GI. The problem I suspect is that the ``props`` attributes are generated at runtime so pylint does not understand them. We could call the relevant .get/.set functions and the error goes away. I see two options, use the get/set functions or document and ignore them.

The other GI errors in StandardItems are probably similar runtime problems. I can workaround it by calling .new. Below is an example patch of both.

I pushed two commits, one to add ngettext to additional-builtins and the other to remove the ``blueman.Lib`` wrapper. The wrapper is just cosmetic and using ``_blueman`` directly works fine.

SerialManager confuses me, no clue why it complains..

The branch runs well for me so I think this is ok to merge.

```diff
diff --git a/blueman/gui/CommonUi.py b/blueman/gui/CommonUi.py
index 4d9fb3d..48b25cc 100644
--- a/blueman/gui/CommonUi.py
+++ b/blueman/gui/CommonUi.py
@@ -11,7 +11,7 @@ from gi.repository import Gtk
 
 
 def show_about_dialog(app_name, run=True):
-    about = Gtk.AboutDialog()
+    about = Gtk.AboutDialog.new()
     about.set_name(app_name)
     about.set_version(VERSION)
     about.set_translator_credits(_("translator-credits"))
diff --git a/blueman/gui/MessageArea.py b/blueman/gui/MessageArea.py
index aa25efc..872a279 100644
--- a/blueman/gui/MessageArea.py
+++ b/blueman/gui/MessageArea.py
@@ -150,7 +150,7 @@ class MessageArea(Gtk.EventBox):
             anim.disconnect(sig)
             anim.freeze()
 
-        if not self.props.visible:
+        if not self.get_visible():
             sig = self.anim.connect("animation-finished", on_finished)
 
             self.anim.thaw()
```

cschramm:
Hmm... I just realized that pylint complains over just some of our usages of props. Will take a closer look at that.

Not sure what you mean with StandardItems. Do you maybe see more errors than me? I'm using pylint 1.4.1.

PR #284, opened 2015-05-12
Update dependencies

cschramm:
* Split into runtime and build dependencies
* Add GdkPixbuf and Pango
* Point out which dependencies are via GI bindings

I'm not sure what's the state of the cairo dependency. It's not really via GI, is it?

Also, I think we can drop the icon scheme dependency as it should work with any GTK+ compliant icon scheme.

### DIFF
--- a/Dependencies.md
+++ b/Dependencies.md
@@ -3,6 +3,7 @@
 ## Runtime dependencies
 
 * [BlueZ](http://www.bluez.org/) (>= 4.61)
+* [cairo](http://cairographics.org/) (GI bindings)
 * [dbus](http://www.freedesktop.org/wiki/Software/dbus/)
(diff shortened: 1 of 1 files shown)


cschramm:
Added cairo. Ready to merge. Objections?

infirit:
>I'm not sure what's the state of the cairo dependency. It's not really via GI, is it?

Yes, pycairo is a normal c module, no GI involved. Other than that it look good to me.

cschramm:
I think it's not via GI _Repository_, but it's still a GI binding provided by PyGObject. Or where does the cairo Python module we're using come from? At least it's not the old pycairo.

I have a `dist-packages/gi/_gi_cairo.so` provided by the package python-gi-cairo and its source package is pygobject.

Issue #189, opened 2015-02-19
Too many dbus signal handlers


juanitotc:
Using bluez-5.24

There appears to be a timing issue when trying to connect to an iPhone.

When ititiating pairing from blueman, by the time (a few seconds) the random pin pops up on the iPhone with the message "OK to pair", blueman has stopped waiting and reports a pairing failure.

Generate (/usr/local/lib/python2.7/site-packages/blueman/gui/manager/ManagerDeviceMenu.py:215)
juanitotc iPhone 
_________
Run (/usr/local/lib/python2.7/site-packages/blueman/main/PluginManager.py:203)
Function on_request_menu_items on Services Failed 
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/blueman/main/PluginManager.py", line 207, in Run
    ret = getattr(inst, function)(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/blueman/plugins/manager/Services.py", line 39, in on_request_menu_items
    manager_menu.Signals.Handle("bluez", device, manager_menu.service_property_changed, "PropertyChanged")
  File "/usr/local/lib/python2.7/site-packages/blueman/main/SignalTracker.py", line 36, in Handle
    obj.handle_signal(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/blueman/bluez/PropertiesBlueZInterface.py", line 57, in handle_signal
    self._handle_signal(wrapper, 'PropertiesChanged', interface, self.get_object_path(), **kwargs)
  File "/usr/local/lib/python2.7/site-packages/blueman/bluez/PropertiesBlueZInterface.py", line 33, in _handle_signal
    self.get_bus().add_signal_receiver(handler, signal, interface, 'org.bluez', obj_path, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/dbus/bus.py", line 157, in add_signal_receiver
    watch = self.watch_name_owner(bus_name, callback)
  File "/usr/local/lib/python2.7/site-packages/dbus/bus.py", line 373, in watch_name_owner
    return NameOwnerWatch(self, bus_name, callback)
  File "/usr/local/lib/python2.7/site-packages/dbus/bus.py", line 80, in __init__
    arg0=bus_name)
  File "/usr/local/lib/python2.7/site-packages/dbus/bus.py", line 160, in add_signal_receiver
    self.add_match_string(str(match))
[...]

PR #290, opened 2015-05-22
Refactor BlueZ layer

### LINKED ISSUE #254: Refactor BlueZ layer
Just to note it down: I'm going to refactor the BlueZ layer to use gsignals as in the the obexd branch. The BlueZ classes shall no longer expose handle_signal and reply and error handler arguments.

The SignalTracker class shall be removed. Support for bluez objects is not relevant anymore with the previous change. dbus objects are only used in bluez (and obexd) classes and NM related plugins. Those classes should track and unhandle their signals themselves. The only type that would be left is gobject and I do not think it makes sense to track this as the handlers will be removed when the object is deleted anyway.

The changes should get rid of issues like #189.

cschramm:
This especially eliminates the  SignalTracker class as its blindfolded usage may cause trouble like #189. BlueZ related dbus signals are now handled by the BlueZ classes themselves by adding a receiver on creation and removing them on deletion. They events can be subscribed to via gobject signals.

This basically resolves #254 and I'm not sure if removing the reply_handler and error_handler arguments from BlueZ classes methods is worth it. For an example of what I was thinking of, see blueman.bluez.obex.Client.createSession. It would bring a little more abstraction from the BlueZ dbus implementation details as the gobject signal parameters and names are independent of the dbus signal parameters and it would make sure that we're not using blocking calls, but on the other hand it separates the handler from the specific method call so that in most cases we would have to register both signal handlers, do the call, on reply check that the reply is really for that one call we made and remove the signal handlers. I think that's too much overhead for too little benefits.


cschramm:
Let's wait for #291 first. Doesn't make sense to refactor dropped stuff.

PR #294, opened 2015-05-28
Blueman needs pulseaudio-module-bluetooth for audio (not pulseaudio).

leigh123linux:


cschramm:
blueman's pulseaudio integration requires the pulseaudio shared libraries, see [PulseAudioUtils.py](https://github.com/blueman-project/blueman/blob/0a0883f7ed68b5edab8ddc51a8c1c015dd1b0759/blueman/main/PulseAudioUtils.py). The user will need pulseaudio-module-bluetooth to actually use bluetooth audio devices (with recent pa versions) in pulseaudio, but that's separate from blueman.

leigh123linux:
The integration isn't complete without pulseaudio-module-bluetooth (what can blueman do without it?).


I added the following requirements to the fedora package.

```
Requires:       pulseaudio-libs-glib2
Requires:       pulseaudio-module-bluetooth
```

cschramm:
All you can do in blueman now is switch the audio profile of the device.

Current pulseaudio picks up connected audio services directly from BlueZ and exposes devices for them (if you have the bluetooth module of course). Previously blueman took over the discovery and created pulseaudio devices explicitly as this did not work sufficiently in pulseaudio itself (you would have needed pulseaudio-module-bluetooth for blueman to do that, except that it did not exist back then :wink: ).

Actually I'm not sure if blueman still needs any pulseaudio support at all, but I think switching the profile is a necessary step in most cases and is not done by pulseaudio automatically, so it's still a good idea for blueman to give that feature to the user (he could otherwise do it with `pactl set-card-profile` and probably some pa GUI tool, too).

If you want to support bluetooth audio out of the box, requiring pulseaudio-module-bluetooth is correct, but it has nothing to do with blueman. It would fit better into some bluetooth meta package together with blueman in some sense.

PR #295, opened 2015-05-28
fix broken schema

leigh123linux:


infirit:
These are not spelling mistakes. These are relocatable schemas and may have more than one set of settings under the schema path. For example one set of gsmsetting for each phone.

The way they are now is intentional and unless you have actual errors these should stay as they are.

cschramm:
In what way is it broken, @leigh123linux?

leigh123linux:
I thought the empty schemas  displayed in dconf-editor were broken because they have no keys.

infirit:
Yeah, relocatable schemas can be confusing :wink:

PR #309, opened 2015-06-19
Add checking for optional PulseAudio

yars068:
Check for PulseAudio is installed.

### DIFF
--- a/blueman/main/Makefile.am
+++ b/blueman/main/Makefile.am
@@ -1,7 +1,14 @@
 SUBDIRS = applet
 
 bluemandir = $(pythondir)/blueman/main
-blueman_PYTHON = PulseAudioUtils.py AppletService.py PPPConnection.py Config.py DhcpClient.py Device.py __init__.py Mechanism.py NetConf.py SpeedCalc.py DbusService.py PluginManager.py
+
+blueman_PYTHON = \
+	AppletService.py PPPConnection.py Config.py DhcpClient.py Device.py __init__.py Mechanism.py NetConf.py SpeedCalc.py DbusService.py PluginManager.py
+
+if HAVE_PULSEAUDIO
+blueman_PYTHON += PulseAudioUtils.py
+endif
+
 
 CLEANFILES =		\
 	$(BUILT_SOURCES)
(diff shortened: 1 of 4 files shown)


cschramm:
I like the fact that it's runtime optional so it's up to the user to have pa installed or not, so I would definitely ship the support in a binary package (given that pulseaudio is available in the distribution as well). But I guess it makes sense to be able to disable it for users who use the source version or some kind of source package where they can set configuration options.

If no objections are raised, I'll merge it in a couple of days. Thanks for the contribution! :smiley: 

Side note: As I mention in #294 I'm not sure if blueman's pulseaudio integration makes much sense anymore anyway.

infirit:
While I don't see the point as pulse module is a runtime del only and the exception is handled properly, but if this makes folks happy go for it :). I do have a couple of comments.

Just use blueman_PYTHON += pulsefile in Makefile.am and avoid the else part. And have_pulseaudio is not used outside configure.ac so no point calling AC_SUBST on it.

yars068:
@infirit, thanks for criticism, I just started learning autotools, python and C :)

cschramm:
If I see that right, you changed it so that it's disabled by default now. Please change this to enabled.

yars068:
Just changed, and also added checking for libappindicator.

cschramm:
It does not make sense to check for pulseaudio and libappindicator on the build system. The build may include the code or not, but it's the target system that may have the respective package. No need for it on the build system. So especially the warning of it's enabled but not present on the build system, does not make sense.

yars068:
Ok, I will remove the warnings about that.

yars068:
Ok, warnings is removed.

cschramm:
If I get the right, the option to not include those files is now only respected if the respective packages are not available on the build system. That doesn't make sense to me. Just remove the PKG_CHECK_MODULES completely and stay with the AC_ARG_ENABLE to provide an option to skip those files.

yars068:
Oh, sorry, that is my mistake, fixed.

cschramm:
Looks good to me now.

yars068:
Also some small fixes on AC_CONDITIONAL tests.

cschramm:
OK, please squash all those commits and I'll merge it.

yars068:
Done, forced update from my local repository.

Issue #315, opened 2015-06-22
Clicking the icon when blueman-manager is running should close blueman-manager


DavidGriffith:
Many applications that put an icon in the notification area will behave as toggles.  That is, click it once and a window of some sort pops up.  Click it again and the window vanishes.  This doesn't happen with Blueman.  I cloned the repo to see if I could add it myself, but I got lost with the problem of how to properly determine if blueman-manager is already running or not and then how to terminate it.

cschramm:
You can check it like

````python
lockfile = blueman.Functions.get_lockfile('blueman-manager')
if os.path.exists(lockfile):
    pid = blueman.Functions.get_pid(lockfile)
    if pid:
        if blueman.Functions.is_running('blueman-manager', pid):
````

Everything else means it's not running. See `blueman.Functions.check_single_instance`

PR #316, opened 2015-06-23
Clicking the applet icon now toggles on/off blueman-manager.

DavidGriffith:
Using the suggested code in #315, here's my implementation.  Seems to work as intended.

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -292,7 +292,10 @@ def get_lockfile(name):
 
 
 def get_pid(lockfile):
-    f = open(lockfile)
+    try:
+        f = open(lockfile)
+    except:
+        return False
     try:
         return int(f.readline())
     except:
--- a/blueman/plugins/applet/StandardItems.py
+++ b/blueman/plugins/applet/StandardItems.py
@@ -96,12 +96,10 @@ class StandardItems(AppletPlugin):
 
     def on_devices(self, menu_item):
         lockfile = get_lockfile('blueman-manager')
-        if os.path.exists(lockfile):
-            pid = get_pid(lockfile)
-            if pid:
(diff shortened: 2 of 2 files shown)


DavidGriffith:
It looks like the Travis CI test failed (for this and #309) because that testing environment's installations of Bluez and Gthreads are older than what Blueman now requires.

cschramm:
Yes, don't care about travis. :wink:

cschramm:
Looks good to me. I'll merge it in a couple of days.

infirit:
I would change this into a single if/else block, something like below pseudo code.

''''python
if exists and get_pid and is_running:
    kill it
else:
    launch
''''

It's clearer IMO what is going on :smile:.

cschramm:
The problem with that is that you need to use the return value of `exists` in `get_pid` and the return value of `get_pid` in `is_running` and `kill`, so you would need to call the same method twice or three times...

infirit:
Sorry was not clear, (on the mobile).

Yes put the pid and the lockfile in variables, but use a single if/else block instead of three :smile:.

yars068:
@infirit > Yes put the pid and the lockfile in variables, but use a single if/else block instead of three

:) As I see, this is not one three-if block in the code, see blueman.Functions.check_single_instance. IMO that may use as-is.

DavidGriffith:
I don't understand what I should do here?  Copy what yars068 is showing me?

yars068:
@DavidGriffith, notihng to do. This is just another realization of your functionality enhancement. If you like your realization, use it instead of my own.

infirit:
Merged as d8bf0b89292a7573cf4f00c93891f1655f14029b slightly modified.

PR #318, opened 2015-06-27
Functions.py, StandardItems.py: hide or show window on clicking the status icon.

yars068:
That is another realization of 4bfa4399db76b04ad78bca74bdf20d3e332c61f6, nothing three-if's :)

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -292,7 +292,11 @@ def get_lockfile(name):
 
 
 def get_pid(lockfile):
-    f = open(lockfile)
+    try:
+	f = open(lockfile)
+    except:
+	return False
+    
     try:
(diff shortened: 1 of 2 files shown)


infirit:
d8bf0b89292a7573cf4f00c93891f1655f14029b

cschramm:
This is buggy! If the lock file does not exist, the get_pid function raises an unhandled IOError.

infirit:
Not really a problem with this change more with 735a7544e64bb1a3bb0929e4cb2eb1f2f2f41ee0.

fixed with 3ce97e90c9f1dbe53cb3363245343fdccf216592.

cschramm:
Right, I didn't notice the except clause was more generic before.

PR #320, opened 2015-07-04
Fix non-standard icon names

City-busz:
This makes Blueman compatible with the Adwaita icon theme, which is the default theme in GTK+ 3.


cschramm:
Thanks!

PR #332, opened 2015-08-08
Always set codeset for gettext to utf8

mgorny:
We had a [bug report](https://bugs.gentoo.org/show_bug.cgi?id=414871) that blueman crashes when non-utf8 locale is used on the system. It seems that setting codeset for gettext fixes that.

### DIFF
--- a/apps/blueman-adapters
+++ b/apps/blueman-adapters
@@ -11,6 +11,7 @@ from gi.repository import Pango
 import os.path
 import sys
 import signal
+from locale import bind_textdomain_codeset
 
 #support running uninstalled
 _dirname = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
@@ -40,6 +41,7 @@ class BluemanAdapters:
 
         builder = Gtk.Builder()
         builder.set_translation_domain("blueman")
+        bind_textdomain_codeset("blueman", "UTF-8")
         builder.add_from_file(UI_PATH + "/adapters.ui")
         self.dialog = builder.get_object("dialog")
 
--- a/apps/blueman-assistant
+++ b/apps/blueman-assistant
@@ -19,6 +19,7 @@ import os
 import sys
 import signal
 from optparse import OptionParser
+from locale import bind_textdomain_codeset
 
 #support running uninstalled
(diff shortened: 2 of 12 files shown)


cschramm:
Thanks for that contribution! I'll merge it in a couple of days if no objections rise up. What I do not really understand is the selection of files you add that statement to. Shall this be every file where the `_` function is used?

infirit:
The "[upstream](https://bugs.launchpad.net/blueman/+bug/1015110)"  bug has some more info. It looks like this is only an issue with gtkbuilder files.

mgorny:
I added it everywhere where `set_translation_domain()` was called.

cschramm:
Merged. Thank you!

yars068:
Hmmm... Using a hard-coded UTF-8 codeset may cause problems for people who don't use UTF-8, or I'd mistake? I tried with `LANG=ru_RU.KOI8-R blueman-manager`, but no affects. I think, that is because  that hard-code does not allow to change the encoding.

mgorny:
Is there ever a case when `.po` files are recoded to another charset? Not to mention that GTK+ uses utf8 internally AFAIR.

infirit:
>Using a hard-coded UTF-8 codeset may cause problems for people who don't use UTF-8

No because of what the python manual says on [bind_textdomain_codeset](https://docs.python.org/2/library/gettext.html#gettext.bind_textdomain_codeset) (emphasis mine),
>Bind the domain to codeset, changing the encoding of strings __returned__ by the __gettext()__ family of functions. If codeset is omitted, then the current binding is returned.

Also the [LP bug report](https://bugs.launchpad.net/blueman/+bug/1015110), has a good explanation of what is going on.
>The problem is that gettext uses the encoding given in the environment and returns non-utf-8 translated strings but g_markup_escape_text which is called by the Gtk-Builder when loading the UI always expects utf-8 strings.

So it actually fixes problems for non-utf8 (and non iso-latin1) folks like you @yars068 :smile:.

Anyone still using older codesets should stop living in the dark ages and join the rest of us in the wonderful world of unicode :wink:.

PR #335, opened 2015-08-08
Fix TypeError in Gio.Settings.__init__()

yars068:
Hi folks,
I have that error after e657a16d10,  when tried to start blueman-applet:
```
$ blueman-applet 
blueman-applet version 2.0 starting
  File "/usr/bin/blueman-applet", line 112, in <module>
    BluemanApplet()
  File "/usr/bin/blueman-applet", line 56, in __init__
    self.Plugins = PersistentPluginManager(AppletPlugin, blueman.plugins.applet, self)
  File "/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py", line 247, in __init__
    self.__config = Config("org.blueman.general")
  File "/usr/lib64/python2.7/site-packages/blueman/main/Config.py", line 10, in __init__
    Gio.Settings.__init__(self, schema_id=schema_id, path=path)
TypeError: __init__() takes at least 2 arguments (2 given)
```
If I'll fixed it, I now have an error, but that is:
```
  File "/usr/bin/blueman-manager", line 226, in <module>
    Blueman()
  File "/usr/bin/blueman-manager", line 65, in __init__
    top, bottom, left, right = align3.get_padding()
AttributeError: 'NoneType' object has no attribute 'get_padding'
```
Because data/ui/manager-main.ui does not explained the "alignment3" object, should get_has_resize_grip() be removed?

### DIFF
--- a/apps/blueman-manager
+++ b/apps/blueman-manager
@@ -63,7 +63,7 @@ class Blueman:
         if self.window.get_has_resize_grip():
             statusbar = self.Builder.get_object("statusbar")
             margin_right = statusbar.get_margin_right()
-            statusbar.set_margin_right (margin_right +  10)
+            statusbar.set_margin_right (margin_right + 10)
 
         def do_present(time):
             if self.window.props.visible:
--- a/blueman/main/Config.py
+++ b/blueman/main/Config.py
@@ -6,8 +6,8 @@ from __future__ import unicode_literals
 from gi.repository import Gio
 
 class Config(Gio.Settings):
-    def __init__(self, schema_id, path=None):
-        Gio.Settings.__init__(self, schema_id=schema_id, path=path)
+    def __init__(self, schema_id, path=None, **kwargs):
+        Gio.Settings.__init__(self, schema=schema_id, path=path, **kwargs)
 
     def bind_to_widget(self, key, widget, prop, flags=Gio.SettingsBindFlags.DEFAULT):
         self.bind(key, widget, prop, flags)


infirit:
The whole point of 5ab15125dcfe5ebf5dbf578e50903816eb10c340 is to not use schema any-more as this has been [deprecated](https://developer.gnome.org/gio/stable/GSettings.html#GSettings--schema). So ``schema=schema_id`` basically reverts back to using the deprecated schema.

I do not have a Gtk version that has a resize grip so I did not run into your other problem. Please try https://gist.github.com/infirit/be1a3505c3c59598191d and let me know how things go.

cschramm:
`__init__() takes at least 2 arguments (2 given)`

:laughing: What a great error message...

yars068:
@infirit: If have appied be1a35, I get a error now:
```Traceback (most recent call last):
  File "/usr/bin/blueman-manager", line 226, in <module>
    Blueman()
  File "/usr/bin/blueman-manager", line 65, in __init__
    margin_right = status_bar.get_margin_right()
NameError: global name 'status_bar' is not defined```

infirit:
>NameError: global name 'status_bar' is not defined

I suck :cry: , that should be ``statusbar`` no underscore. Updated the gist.

yars068:
Thanks, but i noticed and already fixed that, and it works :)

infirit:
Ok, pushed that. Now for the weirdness.

What distro and version are you running this on?

yars068:
This is Slackware64-14.1, python-2.7.5, gtk+-3.8.2, Xfce 4.12, bluez-5.31.

yars068:
My problem with TypeError is going because the code is not comforms to the [PEP 3102](https://www.python.org/dev/peps/pep-3102):
> Keyword-only arguments are not required to have a default value.
    Since Python requires that all arguments be bound to a value,
    and since the only way to bind a value to a keyword-only argument
    is via keyword, such arguments are therefore 'required keyword'
    arguments.  Such arguments must be supplied by the caller, and
    they must be supplied via keyword.

cschramm:
I think the issue here is a (really) old version of PyGObject. What are you using @yars068?

cschramm:
OK, it's not that old. I think versions before 3.11.2 won't work as they require a schema argument. Maybe we could somehow use the `new_with_schema` method to avoid this...

infirit:
Or we can just revert that commit ignore the deprecation whining for Blueman 2.1. I read somewhere slackware is due for a release soonish and when that happens re-apply this commit.

cschramm:
Maybe it works if we put `schema=None` in front. Not sure if that triggers the warning...

cschramm:
Nope, triggers the warning as well.

Then I think we should revert and live with the deprecation message for now.

yars068:
> What are you using @yars068?

I use PyGObject 3.10.2.

cschramm:
As expected. So the situation is that older versions of PyGObject will always trigger the deprecation warning in newer GLib versions (the releases with the respective change happened nearly at the same time about 22 months ago) and cannot be used in the way introduced by @infirit. Either we switch back so that everybody get the deprecation message or we do a distinction like

````python
if Gio.Settings.__init__.__name__ == 'new_init':
  Gio.Settings.__init__(self, schema_id=schema_id, path=path)
else:
  Gio.Settings.__init__(self, schema=schema_id, path=path)
````

to fix the warning with current versions while staying compatible to older ones.

infirit:
@yars068, can you please confirm the following works for you? https://gist.github.com/infirit/9768d37ff2b8d4ca5d6a

yars068:
Yes, it works.

cschramm:
Let's stick with this then. :smiley:

infirit:
Pushed.

Issue #333, opened 2015-08-08
blueman-applet hanging with 'Bad rfkill event size'


mgorny:
After starting blueman-applet with current git (e657a16d10262d818278c6946472083e312750c6):

```
blueman-applet version 2.0 starting
Stale PID, overwriting
_________
Load (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:60)
['ExitItem', 'ShowConnected', 'AuthAgent', 'DhcpClient', 'NMPANSupport', 'DBusService', 'DiscvManager', 'NetUsage', 'Headset', 'KillSwitch', 'Menu', 'Networking', 'PowerManager', 'PPPSupport', 'NMDUNSupport', 'RecentConns', 'SerialManager', 'StandardItems', 'StatusIcon', 'TransferService', 'GameControllerWakelock', 'AppIndicator'] 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.Headset.Headset'> 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.Networking.Networking'> 
_________
load_nap_settings (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/Networking.py:34)
Loading NAP settings 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.StatusIcon.StatusIcon'> 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.DBusService.DBusService'> 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.NMDUNSupport.NMDUNSupport'> 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.Menu.Menu'> 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
loading <class 'blueman.plugins.applet.ShowConnected.ShowConnected'> 
_________
__load_plugin (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:133)
[...]

PR #336, opened 2015-08-09
RfKill: open /dev/rfkill r+b to avoid creating

mgorny:
Open /dev/rfkill in r+b mode rather than wb since the latter creates the
file if it does not exist. And we certainly can not create /dev/rfkill
as a regular file since this is never going to work and is going to
break a lot of other programs that assume /dev/rfkill is a valid device
node.

Fixes: https://github.com/blueman-project/blueman/issues/333

### DIFF
--- a/blueman/plugins/mechanism/RfKill.py
+++ b/blueman/plugins/mechanism/RfKill.py
@@ -12,6 +12,6 @@ from blueman.plugins.applet.KillSwitch import RFKILL_TYPE_BLUETOOTH, RFKILL_OP_C
 class RfKill(MechanismPlugin):
     @dbus.service.method('org.blueman.Mechanism', in_signature="b", out_signature="")
     def SetRfkillState(self, state):
-        f = open('/dev/rfkill', 'wb')
+        f = open('/dev/rfkill', 'r+b')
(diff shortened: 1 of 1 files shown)


cschramm:
Is this really enough? Doesn't PowerManager think bluetooth is disabled if you do not have /dev/rfkill?

infirit:
Can't we check for the device with os,path (or a glib gfile) and raise LoadException if the device is not found?

mgorny:
@cschramm, well, it is enough for my system, and it is certainly a huge improvement from the current DoS behavior.

@infirit, that doesn't really make sense. The net result is the same, except that you introduce a race condition.

cschramm:
What race condition?

mgorny:
Race condition if device gets created/removed between `os.path.exists()` check and the `open()` call. Unlikely but it's a good coding practice to avoid race conditions.

cschramm:
Those rules only set permissions for the file. I don't think it can appear and disappear on runtime unless it may be a kernel module. If it could appear, we'd have a race condition.

yars068:
Yes, on my system, rfkill is kernel module, but that is possible to unload it. That module is used by some other modules, including bluetooth, and only that does not allow to unload rfkill module.
```
 lsmod | grep rfkill
rfkill                 15763  6 cfg80211,acer_wmi,bluetooth
```

PR #340, opened 2015-08-13
TransferService: Fix UnicodeDecodeError in on_transfer_completed

yars068:
Hi folks,
After 78ffdde2 I upgraded my system to Slackware-current, (is actual on Wed Aug 12 05:50:41 UTC  2015), according to [ChangeLog](http://mirror.yandex.ru/slackware/slackware64-current/ChangeLog.txt). Now, I have a errors listed [here](http://pastebin.com/1Se9GJk7), when I tried to receive a file from my phone. One of these errors is now UnicodeDecodeError, and I fixed it.

### DIFF
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -207,6 +207,10 @@ class TransferService(AppletPlugin):
 
         src = attributes['path']
         dest_dir = self._config["shared-path"]
+        # We get bytes from pygobject under python 2.7
+        if hasattr(dest_dir, 'upper') and hasattr(dest_dir, 'decode'):
+    	    dest_dir = dest_dir.decode('UTF-8')
+    	
         filename = os.path.basename(src)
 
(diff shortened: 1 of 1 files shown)


infirit:
This implicit encode/decode still is causing problems, le sigh. The fix is not right as you are doing multiple implicit and explicit encode to bytes and decode to codepoints.

The main problem is that ``dest_dir`` under python2 is a byte string which we need to __decode__ to codepoints. But under python3 all of them are codepoints.

To fix it under python2 you use  ``dest_dir = self._config["shared-path"].decode("UTF-8")``. However this does not work under python3. I'll come up with a fix that works across versions.

infirit:
Please try https://gist.github.com/infirit/5d719c51d79856daa389.

yars068:
Works fine, thanks! :)

infirit:
Pushed as 1b430682ed5bb819f47916eedcadc375af6ff832

PR #343, opened 2015-08-15
TransferService: decode also source string to codepoints

yars068:
Hi,
UnicodeDecodeError is become back, as Terminator :) That is now, but also on 'src' variable. I tried to receive a files with same names two times and get this error again. Then, I just added a debug output like:
```python
dest_dir = os.path.basename(src)
+dprint("Object '\src\' is of type %s" % type(src)
```
and see (dest_dir is already decoded by  1b430682):
```
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'src' is of type <type 'unicode'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'dest_dir' is of type <type 'str'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'filename' is of type <type 'unicode'> 
adding action
_________
```
Then, I developed a patch.

### DIFF
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -213,12 +213,12 @@ class TransferService(AppletPlugin):
         if hasattr(dest_dir, "upper",) and hasattr(dest_dir, "decode"):
             dest_dir = dest_dir.decode("UTF-8")
 
-        if os.path.exists(os.path.join(dest_dir, filename)):
+        dest = os.path.join(dest_dir, filename)
+        if os.path.exists(dest):
             now = datetime.now()
             filename = "%s_%s" % (now.strftime("%Y%m%d%H%M%S"), filename)
             dprint("Destination file exists, renaming to: %s" % filename)
 
-        dest = os.path.join(dest_dir, filename)
         shutil.move(src, dest)
 
         try:


infirit:
Can you please post the traceback.

yars068:
Please, see [here](http://pastebin.com/UL22JdQ0). This is complete output. I received a number different files, big and small, from the phone. When transfer is done, the files does not moved to dest_dir, because the check
```python
        if os.path.exists(os.path.join(dest_dir, filename)):
            now = datetime.now()
            filename = "%s_%s" % (now.strftime("%Y%m%d%H%M%S"), filename)
            dprint("Destination file exists, renaming to: %s" % filename)
```
will fail, because, as I pointed above, unicode string tried to join with 'str' string, and if will failed.
But, I noticed, that is, if I launced blueman-applet with LANG=en. If blueman-applet is launched without LANG environment, it works fine. When I having patch applied, blueman-applet works fine in both cases .

infirit:
I do not understand, in your debug output it show ``Object 'src' is of type <type 'unicode'>`` so there is no need to decode ``src`` from bytes (str) to codepoints (unicode) :confused:..

cschramm:
Doesn't make sense to me either. You say you got

````python
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'dest_dir' is of type <type 'str'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'filename' is of type <type 'unicode'> 
````

That's be fore your patch, right?

After `dest_dir = dest_dir.decode("UTF-8")` `dest_dir` should be `unicode`, so I cannot see a reason for the exception...

**Comment by yars068** [comment]
Please try [this](https://gist.github.com/yars068/d4d20745920fd4b31fba), if I just launch blueman-applet, it says like
```
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Before object 'src' is <type 'unicode'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Before object 'dest_dir' is <type 'str'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'src' is <type 'unicode'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'dest_dir' is <type 'unicode'> 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py:201)
Object 'filename' is <type 'unicode'> 
```
That means, the patch is not need, yes, but if need to launch blueman with locale that differrent from system locale, this error will be appeared. And yes, that is not Unicode**Decode**Error, this is my mistake. This is UnicodeEncodeError...

**Comment by infirit** [comment]
Eventhough this patch is not correct we still have a real bug to fix, re-opening.

@yars068, do not get frustrated, bytes vs codepoints under python 2.7 is very confusing and hard to figure out. Which is why we need to make sure what the actual problem is.

Can you provide me the example strings for ``src`` and ``dest_dir`` so I can poke at them locally. I think I know what is going on and I suspect my earlier fix is incomplete.

**Comment by infirit** [comment]
btw, for anyone having to deal with bytes and codepoints in python the following is essential, http://nedbatchelder.com/text/unipain.html.

**Comment by yars068** [comment]
@infirit, thanks you. This is filenames such as 'MFL68003303_06_ENG_RS.pdf', 'The Moon.png', and so on. Also, I think, the problem is in encoding. When I pass  just `LANG=en blueman-applet`, the problem is present, but if I passed LANG=en_US.UTF-8, then no problem.

**Comment by infirit** [comment]
@yars068, also ``dest_dir`` please, this is the one causing problems :wink:.

``LANG="en"`` is an invalid locale and it then falls back to ``LANG="C"``` which is ascii __only__! If we fall back to ``C`` then any character outside of ascii will cause problems (not just blueman). A valid locale would be ```LANG="en_US"``` or ```LANG="ru_RU"```. I hope to understand more when we know what ``dest_dir`` is.

**Comment by yars068** [comment]
`dest_dir` is a path to directory `~/Общедоступные`, which determined with `self._config[shared_path]`.

**Comment by yars068** [comment]
Hmm, I think, no need to doing `os.path.join(dest_dir, filename), maybe just use `shutil.move(filename, dest_dir) instead?

**Comment by infirit** [comment]
Here is what happens, an __invalid__ locale is set with ``LANG="en"``. This triggers the fallback of locale ``C`` which is ascii __only. Then we try to check if the destination file already exists and we try to encode codepoints (``~/Общедоступные``) to bytes with ascii encoding which fails because Cyrillic is not part of ascii.

>It is possible to avoid this problem, just change ``dest_dir`` do directory with ASCII-name.

Sure but you should not need, it should just work with any locale except ascii only ones. After 1b430682ed5bb819f47916eedcadc375af6ff832 python2 should not __implicitly__ decode to ascii any-more so the original problem should not show up any more.

>use ``shutil.move(filename, dest_dir)`` instead?

No because we need to account for current path which is most definitely not correct. Also the problem is not with filename/src but with destdir (Общедоступные).

AFAIK ``~/Общедоступные`` is never going to work with __any__ program using ``C`` locale, nothing we can do about that. Either pick a locale that can handle Cyrilic script or use ascii only characters. My recommendation would be to use UTF-8, ``ru_RU.UTF-8``.

The main question I still have is, does it work properly with locale ``ru_RU`` in either UTF-8 or KOI8-R character encoding? You already confirmed it works on ``en_US.UTF-8`` so I suspect it will but please confirm.

**Comment by yars068** [comment]
The system locale already is ru_RU.UTF-8, so that should work with ru_RU.KOI8-R properly if system locale is set to that one. But, when system locale is ru_RU.UTF-8 and I invoke `LANG=ru_RU.KOI8-R`, I have that traceback when blueman treis to move received file to `dest_dir`:
```
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Manager.py:56)
/org/bluez/obex/server/session3/transfer2 True 
  File "/usr/lib64/python2.7/site-packages/blueman/plugins/applet/TransferService.py", line 225, in _on_transfer_completed
    shutil.move(src, dest)
  File "/usr/lib64/python2.7/shutil.py", line 302, in move
    copy2(src, real_dst)
  File "/usr/lib64/python2.7/shutil.py", line 130, in copy2
    copyfile(src, dst)
  File "/usr/lib64/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 2] ��� ������ ����� ��� ��������: u'/home/yars/\u041e\u0431\u0449\u0435\u0434\u043e\u0441\u0442\u0443\u043f\u043d\u044b\u0435/TP_of_data_transmission_2010.pdf'
```

**Comment by infirit** [comment]
>system locale is ru_RU.UTF-8 and I invoke LANG=ru_RU.KOI8-R

Why would you do this? Isn't UTF-8 good enough? Are you running __only__ blueman with this  or your whole __user__ session?

Please try the following and let us know if this works.

```diff
diff --git a/blueman/plugins/applet/TransferService.py b/blueman/plugins/applet/TransferService.py
index 458a100..6dcf3ba 100644
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -219,7 +219,9 @@ class TransferService(AppletPlugin):
             dprint("Destination file exists, renaming to: %s" % filename)
 
         dest = os.path.join(dest_dir, filename)
-        shutil.move(src, dest)
+
+        shutil.move(src.encode("UTF-8"), dest.encode("UTF-8"))
 
         try:
             status_icon = self._applet.Plugins.StatusIcon
```

yars068:
> Why would you do this?

That need, if need to post output of any program in english to english-speaking internet resource, just to keep english messages from that program. I should remember, that invoking `LANG=en whatever` is not correct, and locale settings will be fallback to `C`, instead of selecting one of available locales that matches `en`, as I was thiking before. Accrodingly, that is not a bug, but just restriction. Then, in future, I will use LANG=en_US.UTF-8 to post output from programs.

infirit:
>That need, if need to post output of any program in english

Ah, I see now :smile_cat:. Try the patch anyway, I am curious how it works for you.

yars068:
I tried the patch, and It lead to raising an UnicodeEncodeError exception, when used C locale, as expected, and works fine, when the locale is in valid UTF-8 encoding.

yars068:
Hmm, the problem is much more [complicated](https://bugs.python.org/issue19846) than I thought.

infirit:
>Hmm, the problem is much more complicated than I thought.

Yup, things get really complicated really quick.. there are no simple answers when dealing with languages.

@cschramm, I think there is nothing we can do in blueman. Encoding to bytes before sending it to ``shutils`` does workaround a corner case and __may__ prevent some issues in the future. Not sure if it is worth adding this but afaics it does not cause regressions, I am leaning to leaving things as they are now. Opinion or suggestion?

cschramm:
What is the edge case here? That the shared-path configuration has been stored in UTF-8 and is interpreted as KOI8 (in general: The encoding is different when reading the configuration.)? A solution to this would be to make sure that configuration is always read and wrote using UTF-8 and ignoring the user's encoding, right?

infirit:
When using KOI8-R with blueman on an UTF-8 system, [this](https://github.com/blueman-project/blueman/pull/343#issuecomment-131549768) traceback. I can reproduce this with https://gist.github.com/infirit/6726ecef4924f11a3f63 on both python 3.4 and 2.7.

cschramm:
I can reproduce it as well. Changing the coding of the file to KOI8-R makes it work.

I think the underlying issue is persistence. When shared-path was stored, @yars068 had his default locale ru_RU._UTF-8. When he now sets it to KOI8-R, the string stored in GSettings is expected to be KOI8-R as the encoding it was stored with is not known.

It works well if the string is stored and read with either KOI8-R or UTF-8. If it is stored with UTF-8 and read with KOI8-R I get IOError "No such file or directory" as it is obviously a valid, but different KOI8-R string. The other way produces a UnicodeDecodeError as the KOI8-R string is obviously not valid UTF-8.

The following script produces the exact same errors:

````python
#!/usr/bin/env python
# -*- coding: UTF-8 -*-

str = 'Общедоступные'

for wenc, renc in (('UTF-8', 'UTF-8'), ('KOI8-R', 'KOI8-R'),
                   ('UTF-8', 'KOI8-R'), ('KOI8-R', 'UTF-8')):
  try:
    print(
        # Write
        str.encode(wenc)
        # Read
           .decode(renc)
    )
  except Exception as e:
    print(e)
````

I think we can fix this by making sure that we always write and read UTF-8. My idea would be to override `__getattr__` and `__setattr__` in our `Config` class like:

````python
def __setattr__(self, name, value):
  if type(value) == 'str':
    value = value.encode('UTF-8').decode(locale.getdefaultlocale()[1])
  super(Config, self).__getattr__(name)

def __getattr__(self, name):
  value = super(Config, self).__getattr__(name)
  if type(value) == 'str':
    value = value.encode(locale.getdefaultlocale()[1]).decode('UTF-8')
  return value
````

That should lead to the following script:

````python
#!/usr/bin/env python
# -*- coding: UTF-8 -*-

str = 'Общедоступные'

for wenc, renc in (('UTF-8', 'UTF-8'), ('KOI8-R', 'KOI8-R'),
                   ('UTF-8', 'KOI8-R'), ('KOI8-R', 'UTF-8')):
  try:
    print(
        # Write
        str
          .encode('UTF-8')
          .decode(wenc)
          .encode(wenc)
        # Read
          .decode(renc)
          .encode(renc)
          .decode('UTF-8')
    )
  except Exception as e:
    print(e)
````

which works for me in both Python 2 and 3! :smiley:

infirit:
The problem is not with gsettings, we decode from bytes to codepoints properly.  It is only when we send these __codepoints__ to shutils, shutils runs into trouble not blueman, os.path has similar problems.

If we decoded incorrectly my [patch](https://github.com/blueman-project/blueman/pull/343#issuecomment-131577469) encoding ``filename`` and ``dest_dir`` to bytes (utf-8) should have failed with a similar IOError, it doesn't.

Regarding the changes to ``Config``,
``type(value) == 'str'`` does not work across python versions. In python3 ``str`` is codepoints and in python2 it is bytes. Also if we do ``str.encode()`` what python2 does is __decode__ to codepoints first to then encode back to bytes because encode is an operation done on codepoints not bytes. This implicit decoding is python2 trying to be helpful (it's not).

So the below under python2,
``
value = value.encode('UTF-8').decode(locale.getdefaultlocale()[1])
``

will in actuality do,
``
value = value.decode(sys.getdefaultencoding()).encode('UTF-8').decode(locale.getdefaultlocale()[1])
``

Which may work if ``sys.getdefaultencoding()`` gives you ``utf-8`` (true for me on python3) but will blow up eventually when it is ``ascii`` (true for me on python2).

cschramm:
> The problem is not with gsettings, we decode from bytes to codepoints properly. It is only when we send these __codepoints__ to shutils, shutils runs into trouble not blueman, os.path has similar problems.

Concerning `shared-path` we take the output of `get_filename` and put it into `Gio.Settings`' setter. That output is shown to me as `str` (which is codepoints to me). I suppose this is encoded by PyGObject using the current locale's encoding when converting it into a C string. This is the `.encode(wenc)` in my test scripts. When fetching the stored data from `Gio.Settings`' getter, we get a `str` again, which I suppose is the C string decoded with the current locale's encoding. If this encoding differs from the one it was stored in, we get wrong data here and need to re-encode it.

Actually I do not fully understand why my proposal seems to work, but it does.

> Which may work if sys.getdefaultencoding() gives you utf-8 (true for me on python3) but will blow up eventually when it is ascii (true for me on python2).

My python2 `sys.getdefaultencoding()` is `ascii` too, but I do not run into any issue with my test script. Do you?

infirit:
>That output is shown to me as str (which is codepoints to me).

Depends on which python version, under python2 this is bytes and under python3 it is codepoints. 

I do not want the below to be considered "belittling" or "rude", I just want to get this clear because codepoints/unicode and bytes/character-encoding under python can be very confusing (it was for me!). If ppl do get offended my apologies upfront as this is not my intention.

 * [codepoints](https://en.wikipedia.org/wiki/Unicode#Architecture_and_terminology) is the mapping between codes (U+XXXX) and characters. Under python2 this is represented as the ``unicode`` data type and in python3 as the ``str`` data type.
 * bytes is what we actually store in memory and on disk. This is represented under python2 as the ``str`` data type and under python3 as the ``bytes`` data type.
 * In python when we ``encode`` a string we encode codepoints to bytes  which requires a character encoding (utf-8, ascii, koi8-r, etc). The reverse is decoding, which mean we take the bytes with a certain character encoding and decode them to codepoints.
 * Python2 will implicitly encode/decode between ``unicode`` and ``str`` with ``sys.getdefaultencoding()``.
 * Python3 will not implicitly encode/decode any-more and we have to know our bytes from our codepoints and explicitly convert between them.
 * Under python2 pygobject gives bytes (``str``) which we need to decode to codepoints (``unicode``).
 * Under python3 pygobject gives us codepoints (the ``unicode`` data type from python2).
 * Glib/Gtk use "utf-8" encoded (byte) strings __internally__ but does not require them.
 * utf-8 is a superset of iso latin-1, so a lot of the __implicit__ encode and decode operation will appear to work fine.

When we run into users like @yars068 we get to deal with these issues in a major way and have the opportunity to make sure we got things right. I am happy to see we got things mostly right :smile:.

>Concerning shared-path we take the output of get_filename and put it into Gio.Settings' setter.

Under python2 ``get_filename()`` will return a utf-8 encoded (``str``) string which we store as is in the  ``shared-path`` key. Under python3 we only have to deal with codepoints and pygobject will handle the encoding to bytes for us.

>why my proposed re-encoding method seems to work, but it does.

Most likely because python2 will implicitly encode/decode for you. But the script never tries to open a file or call os.path.exists, I suspect when it does things go wrong.

Now, here is what I think happens when we send codepoints to shutils/os.path. At some point the codepoints will need to be encoded to bytes for the C library to use them. When we use ``LANG="ru_RU.KOI8-R"`` and when we let shutils do this it will use ``koi8-r``. What happens next I do not know but my guess is that the ``KOI8-R`` encoded string gets mangled by libc/python which "assumes" it is ``utf-8`` as this is the system's character encoding. My guess is based on the fact that when I __explicitly__ encode to utf-8 things work fine.

cschramm:
I just tried the following write script:

````python
from gi.repository import Gio
import os

Gio.Settings('org.blueman.transfer')['shared-path'] = os.getcwd()
````

and the following read script:

````python
from gi.repository import Gio
import os

assert os.path.exists(Gio.Settings('org.blueman.transfer')['shared-path'])
````

in a folder called Общедоступные and with different combinations of the locales de_DE.UTF-8 and ru_RU.KOI8-R.

Expected outcome: Same encoding for write and read should work. Different encodings should fail.

Confusingly the tests do not produce an error in python2 at all (something wrong with the `__future__` imports?). In python3 I get `AssertionError`s when encodings for write and read are different.

I can confirm that my recoding proposal does not fix this.

I do not understand your patch, @infirit, but do you think we could just `.encode('UTF-8')` any string we get from GSettings?

infirit:
>Confusingly the tests do not produce an error in python2 at all (something wrong with the __future__ imports?).

No nothing wrong, yes very confusing but understandable.

When we import from ``__future__`` we do not really import anything. We just tell the python interpreter to behave differently. In the case of ``unicode_literals`` it will instruct the interpreter to make ``unicode`` (codepoints) the default data type when dealing with strings. For example ``x = "test"`` under normal python2 will be of type ``str`` (bytes), but with ``unicode_literals`` this will be of type ``unicode`` (codepoints). However 3rd party modules (and several in the standard library) have no understanding of this and will keep returning bytes. This applies to gsettings/pygobject which continues to return bytes even-though we tell ``python2`` to use ``unicode_literals``. This was very annoying when adding [1] python3 support :smile:.

What I did when adding python3 support is make an unicode sandwich. Internally we deal only with codepoints. Then when we receive from the outside world decode to codepoints as soon as possible and when sending encode to bytes at the very last moment. We do this so we know what type of strings we are dealing with and to avoid issues like #340 where python2's nasty "feature" of implicit encoding/decoding shows it's ugly head.

My patch takes the "decision" away from shutils by forcing it to use the utf-8 encoded string. Like I said, at some point the codepoints will need to be encoded, I only made it explicit and forced the character encoding. Having thought about it a bit more my patch is wrong and should be tossed in the bin.

>but do you think we could just .encode('UTF-8') any string we get from GSettings?

No, this is what I meant with corner case. Manipulating the environment like we did here is not something we can  code for. If the whole system, including the user session, is running under an utf-8 character encoding and then for some reason we manipulate blueman to run with a different character encoding we should not be surprised things do not go well.

Re the script from [your](https://github.com/blueman-project/blueman/pull/343#issuecomment-131912591) comment. When the script is run under python2 ``os.path.exists`` is still passed bytes not codepoints. If you add a ``.decode("UTF-8")`` at the end with python2 you should see this problem appear.

[1] Porting would have been much easier, blueman now supports both version from the same codebase which complicates things considerably as this thread demonstrates :wink:

cschramm:
> Re the script from [your](https://github.com/blueman-project/blueman/pull/343#issuecomment-131912591) comment. When the script is run under python2 ``os.path.exists`` is still passed bytes not codepoints. If you add a ``.decode("UTF-8")`` at the end with python2 you should see this problem appear.

I think the script is pretty much doing the same thing blueman does with `shared-path`. I actually even tried a GtkFileChooserDialog instead of `getpwd` with the same result. But the issues describe in this thread by @yars068 were experience with python2 and my first script from [this comment](https://github.com/blueman-project/blueman/pull/343#issuecomment-131784351) also produces the same issues for python2 and python3. I did not try `shutil.move` instead of `os.path.exists`. Maybe that's the difference?

yars068:
There are no differences between `shutil.move` and `os.path.exist` in it's behavior. Both expects ASCII-strings with C locale.

PR #344, opened 2015-08-15
TransferService: small code optimization for on_transfer_completed

yars068:
Nothing to comment, just subject.


cschramm:
Makes sense. Let's wait for the outcome of #343 first.

PR #354, opened 2015-09-08
Add generic names for blueman.desktop and blueman-manager.desktop

Somasis:
Right now, application menus will use "Blueman Bluetooth Manager" for the GenericName, which is wrong

### DIFF
--- a/data/blueman-manager.desktop.in
+++ b/data/blueman-manager.desktop.in
@@ -1,6 +1,7 @@
 [Desktop Entry]
 _Name=Bluetooth Manager
 _Comment=Blueman Bluetooth Manager
+_GenericName=Bluetooth Manager
 Icon=blueman
 Exec=blueman-manager
 Terminal=false
--- a/data/blueman.desktop.in
+++ b/data/blueman.desktop.in
@@ -1,6 +1,7 @@
 [Desktop Entry]
 _Name=Blueman Applet
 _Comment=Blueman Bluetooth Manager
+_GenericName=Bluetooth Manager
 Icon=blueman
 Exec=blueman-applet
 Terminal=false


infirit:
How is this wrong?

Somasis:
http://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys

> GenericName
> Generic name of the application, for example "Web Browser". 

Menus such as XFCE's application menu, and Whisker Menu have the option of using the generic name over the application's "branded" name, and without this, Blueman doesn't use a proper GenericName. That makes it not consistent with the other applications; Firefox for example, has "Web Browser", Thunderbird is "Mail & News Client", Gimp is "Image Editor" and so on.

infirit:
I looked at mintmenu I have installed here and does not show anything with blueman in the menu name. It show the Name field as we set in the desktop file. It does show below it the comment field which indeed has blueman. I also looked at Xfce's menu in my ubuntu 15.04 vm and I also don't see blueman as entry. Same thing applies to the MATE menu's, no mention of blueman in the entries.

Also the Name field is identical to the new GenericName one you are adding, this makes no sense to me. And it looks like you copy/pasted the same GenericName to two desktop files which is wrong.

If you do see it somewhere please post a screenshot and the name of the menu.

Somasis:
Okay, it turns out I was misreading my menu, I didn't realize that was the comment. This change is not right :)

infirit:
Phew, I was scared I was missing something obvious :smiley:

Somasis:
Although... maybe the comment should be changed? It seems a little redundant, if it's mentioning Blueman in it. Maybe a description of what it does would be more fitting; "Configure bluetooth devices", or something similar.

PR #361, opened 2015-10-01
Add systemd system and user service units for the D-Bus services

smcv:
The D-Bus system service (the mechanism) becomes a systemd system
service, and the D-Bus session service (the applet) becomes a systemd
user service.

In both cases, when systemd is used for service activation, this allows
placing the blueman-related process in its own cgroup (and applying
systemd resource control and life-cycle tracking), whereas without
these units the blueman processes would be part of the
dbus.service cgroup.

dbus running under systemd has provided system service activation via
systemd for some time. User service activation via systemd is a
recent addition (D-Bus 1.10) and not always enabled, but installing
the necessary metadata is harmless in any case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,8 @@ Makefile
 /stamp-h1
 /blueman/Constants.py
 /blueman/plugins/mechanism/Rfcomm.py
+/data/configs/blueman-applet.service
+/data/configs/blueman-mechanism.service
 /data/configs/org.blueman.Applet.service
 /data/configs/org.blueman.Mechanism.service
 /module/.deps
--- a/configure.ac
+++ b/configure.ac
@@ -246,6 +246,8 @@ data/icons/main_icon/Makefile
 data/icons/pixmaps/Makefile
 data/ui/Makefile
 data/man/Makefile
+data/configs/blueman-applet.service
+data/configs/blueman-mechanism.service
 data/configs/org.blueman.Applet.service
 data/configs/org.blueman.Mechanism.service
 module/Makefile
--- a/data/configs/Makefile.am
+++ b/data/configs/Makefile.am
@@ -7,6 +7,12 @@ dbus_services_DATA = org.blueman.Mechanism.service
 dbus_sessdir = $(datadir)/dbus-1/services
 dbus_sess_DATA = org.blueman.Applet.service
 
+systemd_systemdir = $(prefix)/lib/systemd/system
+systemd_system_DATA = blueman-mechanism.service
+
+systemd_userdir = $(prefix)/lib/systemd/user
+systemd_user_DATA = blueman-applet.service
+
 if HAVE_POLKIT
 @INTLTOOL_POLICY_RULE@ 
 policykitdir = $(datadir)/polkit-1/actions
@@ -14,12 +20,17 @@ policykit_in_files = org.blueman.policy.in
 policykit_DATA = $(policykit_in_files:.policy.in=.policy)
 endif
 
-EXTRA_DIST = org.blueman.Mechanism.conf		\
+EXTRA_DIST = \
+	blueman-applet.service.in		\
+	blueman-mechanism.service.in		\
+	org.blueman.Mechanism.conf		\
 	org.blueman.Applet.service.in		\
 	org.blueman.Mechanism.service.in	\
 	org.blueman.policy.in
 
 CLEANFILES =		\
+	blueman-applet.service		\
+	blueman-mechanism.service	\
 	org.blueman.Mechanism.service	\
 	org.blueman.policy		\
 	org.blueman.Applet.service		\
--- a/data/configs/blueman-applet.service.in
+++ b/data/configs/blueman-applet.service.in
@@ -1,0 +1,7 @@
+[Unit]
+Description=Bluetooth management applet
+
+[Service]
+Type=dbus
+BusName=org.blueman.Applet
+ExecStart=@BINDIR@/blueman-applet
--- a/data/configs/blueman-mechanism.service.in
+++ b/data/configs/blueman-mechanism.service.in
@@ -1,0 +1,7 @@
+[Unit]
+Description=Bluetooth management mechanism
+
+[Service]
+Type=dbus
+BusName=org.blueman.Mechanism
+ExecStart=@LIBEXECDIR@/blueman-mechanism
--- a/data/configs/org.blueman.Applet.service.in
+++ b/data/configs/org.blueman.Applet.service.in
@@ -1,3 +1,4 @@
 [D-BUS Service]
 Name=org.blueman.Applet
 Exec=@BINDIR@/blueman-applet
+SystemdService=blueman-applet.service
--- a/data/configs/org.blueman.Mechanism.service.in
+++ b/data/configs/org.blueman.Mechanism.service.in
@@ -2,3 +2,4 @@
 Name=org.blueman.Mechanism
 Exec=@LIBEXECDIR@/blueman-mechanism
 User=root
+SystemdService=blueman-mechanism.service


smcv:
This backports cleanly onto the 2.0 branch after cherry-picking 49e11eb6.

infirit:
What is the impact on non systemd ditributions?

smcv:
On non-systemd distributions, dbus-daemon will continue to use traditional activation like it does now. The `SystemdService` parameter is ignored unless the dbus-daemon is run as `dbus-daemon --systemd-activation` option, and nothing will be reading the systemd service files.

I can make the installation of files into `/usr/lib/systemd/user` optional if you want, at the cost of a few extra lines in configure.ac.

infirit:
>I can make the installation of files into /usr/lib/systemd/user optional

I don't think that is necessary if the line is ignored. On Gentoo unit files are installed unconditionally and binary disto's have their own ways to drop these files.

I don't use systemd so I can't really test this. But...

 * I noticed you use ``$(prefix)/lib`` and not ``$(libdir)`` in ``data/configs/Makefile.am``. Is there a specific reason for this?
 * Not sure if ``data/configs`` is the right location for the unit files , perhaps it should just go under ``data`` directly, @cschramm?

smcv:
> I noticed you use $(prefix)/lib and not $(libdir) in data/configs/Makefile.am. Is there a specific reason for this?

Yes. systemd always looks in `lib` to keep the location constant across architectures, even on platforms where `${libdir}` is something like `lib64` (Red Hat derivatives) or `lib/x86_64-linux-gnu` (Debian derivatives).

> Not sure if data/configs is the right location for the unit files , perhaps it should just go under data directly

The D-Bus service files and the systemd service files are basically the same thing (integration with the wider system), so they should probably go in the same place. I have no opinion on `data/configs` vs. `data`.

cschramm:
Yes, I agree with that.

The addition looks good to me, but I do not really see what we - or the user - get from moving the management of the (system) service(s) from D-Bus itself to systemd. So basically I do not see a better reason why systemd supports this at all, other than the fact that it wants to manage just everything. What's the actual improvement here?

smcv:
> I do not really see what we - or the user - get from moving the management of the (system) service(s) from D-Bus itself to systemd

For context, I'm a D-Bus maintainer and occasional systemd contributor, not the other way round; this is not me saying "use the technology I maintain", but rather "use this other thing instead, it's better than the one I maintain" :-)

Here are a few reasons, pick your favourite:

dbus-daemon is not really designed to be a service manager; the traditional activation code was added because there was a need for it, but it isn't particularly featureful or high-quality. While testing this and similar patches for various other services, I found and fixed a series of error-handling bugs which seem to have been there since system activation was added in 2007. The code in dbus-daemon that talks to systemd for activation is simpler and more testable (it's just message-passing); for the actual fork/exec, systemd is rather better at starting services, because starting services is most of the reason it exists.

systemd knows how to manage services beyond starting them (restart, stop, check status, report normal or abnormal exit and so on), whereas dbus-daemon can only start services. The administrator of a systemd-based Linux system can mostly assume that the systemd tools manage all system-level services; D-Bus system services are currently an exception to that.

systemd services can have attributes beyond just the name and executable, such as per-service resource limits, automatic restart after crashes, and [security hardening](http://0pointer.de/blog/projects/security.html) via the same kernel mechanisms that containers use. For instance, with a systemd system unit, preventing the privileged blueman-mechanism process from accessing anything below /home (assuming it doesn't need that during normal operation) would be a 1-line change.

systemd puts all the processes of each service in a cgroup (control group), which is a reliable way to associate processes with services so that a service can be managed as a group, without the processes being able to "escape" from the group via double-forking or similar tricks. Services started by dbus-daemon have no particular grouping: D-Bus can tell you the process ID at the other end of a connection, but if a traditional D-Bus service forks, the child has no particular connection to the service name. On a systemd system, the symptom of this is that traditional D-Bus services show up as part of the D-Bus cgroup ("part of D-Bus"). For instance, in this `systemd-cgls` output, `cups-pk-helper-mechanism` is a traditional activated service like the current blueman-mechanism, whereas polkitd was activated via systemd:

```
% systemd-cgls
Control group /:
…
├─system.slice
│ ├─dbus.service
│ │ ├─ 1010 /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopid...
│ │ └─11619 /usr/lib/x86_64-linux-gnu/cups-pk-helper-mechanism
│ ├─polkitd.service
│ │ └─1120 /usr/lib/policykit-1/polkitd --no-debug
```

systemd units can be masked (disabled) by sysadmin configuration. D-Bus services don't have that feature.

And this last one is only really applicable to the most locked-down systems right now, but it's potentially more widely applicable for the future: dbus-daemon comes with a setuid-root binary (dbus-daemon-launch-helper) to help with the system activation process. If all the D-Bus system services on a particular OS are activated via systemd, then that helper isn't needed and can be removed. The D-Bus maintainers believe d-d-l-h is secure, but it has had security bugs in the past, and the most reliable way to be sure a setuid binary doesn't have exploitable bugs is to delete it :-)

cschramm:
> For context, I'm a D-Bus maintainer and occasional systemd contributor, not the other way round; this is not me saying "use the technology I maintain", but rather "use this other thing instead, it's better than the one I maintain" :-)

Enough of a reason for me. :wink: 

Will merge if nobody raises his hand the next days.

M0E-lnx:
This is a step in the wrong direction in my opinion.  You become one more project bowing to systemd.  If dbus option will still work, why go the other way?  There are those of us who will not implement the current systemd philosophy, and here goes another project that will eventually lock us out.

Just to give you a rundown of what will happen, you will sooner or later find yourselves having to maintain systemd and dbus (for non-systemd) compatibility code.  At some point, somebody will get tired of doing both.

smcv:
> here goes another project that will eventually lock us out

Nothing in this change makes blueman work differently on non-systemd systems than it does now.

> you will sooner or later find yourselves having to maintain systemd and dbus (for non-systemd) compatibility code

In the case of dbus-daemon, "sooner or later" was 5 years ago. The non-systemd activation code path is still there, and isn't going away.

cschramm:
I basically agree with @M0E-lnx. It makes sense to support systemd as it looks like D-Bus does the process handling only out of necessity, but this cannot replace the D-Bus way as systemd is not everywhere and probably will not be anytime soon. If we really reach the point where somebody gets tired of maintaining both integrations (which I doubt as it's really small and simple), we'll just throw out the systemd stuff and stick with the more universal one.

smcv:
> this cannot replace the D-Bus way

I don't intend it to. As I said, traditional activation is not going away. My intention is to make blueman integrate slightly better on systemd systems, without making it any worse on non-systemd systems.

yars068:
> without making it any worse on non-systemd systems.

Just a notice: because I don't want to use systemd, I don't want to see systemd-related files on my system, so I added the `--disable-systemd` option to `configure`. That just provide a way to not install these files at build time, without using a various hacks (like removal these files before package creation or maintaining the distribution-specific patches) by maintainers. Now, `configure` will detect have you systemd installed, or not, and do respective things. Please see 3be44340.

cschramm:
That's where it gets complex... :wink: 

@yars068: Please open a pull request for this.

smcv:
If you want compile-time configurability for this, the convention seems to be a `--with-systemduserunitdir` option, with `--without-systemduserunitdir` or `--with-systemduserunitdir=no` disabling installation (if it had been up to me I'd have included more spaces, but a convention is a convention). I didn't include that in my pull request because the complexity didn't seem worth it to save 271 bytes of installed size, but here is a similar change where the maintainer did ask for it: https://git.gnome.org/browse/evolution-data-server/commit/?id=51a37f76bec193847f0ec2c5df4aa879947a47c2

yars068:
smcv> If you want compile-time configurability for this, the convention seems to be a --with-systemduserunitdir option, with --without-systemduserunitdir or --with-systemduserunitdir=no disabling installation

But blueman is used system wide units, right?

@cschramm, done - #373 :)

shevegen:
I came to this thread indirectly, via the new gvfs having a dependency on systemd through the meson build system (URL http://ftp.gnome.org/pub/gnome/sources/gvfs/1.35/gvfs-1.35.4.tar.xz). The old gvfs was working fine without systemd. The meson build system says that you can disable systemd, but it provides no information as to how - old GNU autoconfigure options such as simple ones, like "--disable-systemd", often do not work with meson-based build systems.

Everything changes in the linux ecosystem and definitely not to empower the old "powerusers".

So why do I comment here, in a project that is unrelated? Well, firstly, because the issue tracker was linked in from a discussion by someone else on linuxquestions-forum; secondly to say that I would have less of an issue with the whole ecosystem if OLD FUNCTIONALITY would be preserved and if people could easily select what they want to use.

At any rate, I hope that more projects can settle and "standardize" on options such as "--without-systemduserunitdir" or even just "--with-systemduserunitdir=no". Problem is that sometimes things propagate slowly; I know that the meson guys are at the least listening to input in regards to their issues so I have no qualms with that, but don't even get me started on the systemd folks ...

smcv:
> smcv> If you want compile-time configurability for this, the convention seems to be a --with-systemduserunitdir option, with --without-systemduserunitdir or --with-systemduserunitdir=no disabling installation
>
> But blueman is used both (system and user wide) units, right? But, to follow the convention, I can rework this.

If you (also) have system-wide units, the conventional options are the same, but replacing `user` with `system`. Take a look at the `configure.ac` in dbus, which has both. They're usually configured separately instead of sharing an option that points to the parent directory, because distributions that support "split /usr" install system units in `/lib/systemd/system` and user units in `/usr/lib/systemd/user`.

> I would have less of an issue with the whole ecosystem if OLD FUNCTIONALITY would be preserved

Good, because that's what this pull request did (even before the option was added). On non-systemd-booted machines, blueman starts in the same way it always did.

> I came to this thread indirectly, via the new gvfs

This github repository is nothing to do with gvfs.

PR #365, opened 2015-10-07
RFC: GtkAnimation: Fix CellFade fading

infirit:
Recently I was looking at the old gtk2 blueman version in Ubuntu 14.04. And unfortunately the move to Gtk3/GI broke most of the fading. I fixed the TreeRow(Color)Fade, which leaves CellFade.

I have a partial fix but it is not perfect as the rows in the treeview have a slight gradient for me. Which means the are small color inconsistencies. I will keep poking at this to make it work properly but do we want to have it merged as it is now, imperfect as it is? Below is the commit message,

```
* Clip only after all rectangles are added to the cairo context
* Use background color as source color.

TODO: Background color is not completely correct but Gtk.render_background
does not work either..
```

### DIFF
--- a/blueman/gui/GtkAnimation.py
+++ b/blueman/gui/GtkAnimation.py
@@ -221,8 +221,6 @@ class CellFade(AnimBase):
         self.sig = tw.connect_after("draw", self.on_expose)
         self.row = Gtk.TreeRowReference.new(tw.props.model, path)
         self.selection = tw.get_selection()
-        self.style = Gtk.rc_get_style(tw)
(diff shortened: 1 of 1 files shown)


cschramm:
I think it's an improvement over the current state, isn't it?

infirit:
Before it was very broken and  flashed to/from black, very ugly and noticeable. I then disabled the fading (current state).

It actually fades now in almost the right background so it is an improvement over that yes.

cschramm:
So let's merge it. :smiley:

infirit:
Merged :smile:

PR #369, opened 2015-10-17
Migrate MessageArea from EventBox to InfoBar

infirit:
Looking at this now and we need to discuss this, so opening a PR :smile:. This is based on a slightly modified version of what is on the main git repo.

> We are losing the color animation with this.

How important is this?

> Actually colors do not seem to work at although set_message_type should change them. :/

This highly depends on the theme, for some themes the error is red others use something differently. For the MATE traditional themes it is red but on Adwaita it is blue. We can force by using a StyleProvider if it has to be a specific color.

ps: With the StyleProvider we can force other styling, like background colors, transition effects etc.

### DIFF
--- a/blueman/gui/GtkAnimation.py
+++ b/blueman/gui/GtkAnimation.py
@@ -123,7 +123,7 @@ class TreeRowFade(AnimBase):
         super(TreeRowFade, self).__init__(1.0)
         self.tw = tw
 
-        self.sig = self.tw.connect_after("draw", self.on_expose)
+        self.sig = self.tw.connect_after("draw", self.on_draw)
 
         self.row = Gtk.TreeRowReference.new(tw.props.model, path)
         self.stylecontext = tw.get_style_context()
@@ -137,7 +137,7 @@ class TreeRowFade(AnimBase):
     def get_iter(self):
         return self.tw.props.model.get_iter(self.row.get_path())
 
-    def on_expose(self, widget, cr):
+    def on_draw(self, widget, cr):
         if self.frozen:
             return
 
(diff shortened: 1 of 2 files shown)


infirit:
>We can force by using a StyleProvider if it has to be a specific color.

Here are some commits [1] that show what we can do with StyleProvider. The css styles in blueman.css are silly and just to show what is possible/experiment with. I opted to have a single css file but we could have the css in python modules.

[1] https://github.com/blueman-project/blueman/compare/master...infirit:styleprovider?expand=1

infirit:
Some css that shows how to style the GtkInfoBar.
```css
GtkInfoBar.error GtkButton,
GtkInfoBar.error GtkBox {
background: red;
color: black;
}
```

cschramm:
Well, I'm not sure if it makes sense for the interface to force custom styling, but it probably does in this case.

infirit:
Rebased and fixed the animation problem. However sometimes we get no InfoBar and when it comes back we see the below.
```
(blueman-manager:4616): Gtk-WARNING **: Negative content height -1 (allocation 1, extents 1x1) while allocating gadget (node infobar, owner blueman+gui+MessageArea+MessageArea)
```

infirit:
Fixed the negative size problem by queuing a resize. Now the big question is, is there still a need to force styling now that the animation is fixed?

infirit:
>Now the big question is, is there still a need to force styling now that the animation is fixed?

Added this anyway, can drop it later 😄

infirit:
I left out the commit to make the infobar red as I can't get the button behave on all themes. I'll open a new PR with this commit and try to figure out why it cause me trouble.

PR #374, opened 2015-10-21
Move to GtkGrid

infirit:
With [GtkGrid](https://developer.gnome.org/gtk3/stable/GtkGrid.html#GtkGrid.description) we can better layout were widgets and boxes need to go. It replaces the deprecated GtkTable widgets and reduces the number of GtkV/HBoxes.

As this is a fairly invasive change I would like to get a few people test this out. before pushing it.


cschramm:
:+1:

infirit:
Thx, pushed. Let me know if you run into problems.

PR #378, opened 2015-10-25
configure: added --with-systemd-userunitdir and... 

yars068:
...--with-systemd-systemunitdir options.

P.S. How to edit existing pull-requests, instead of creating new one?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Drop browse feature
 * Add title to status icon
 * Add systemd service units (@smcv / Simon McVittie)
+* Add `--disable-systemd` option to configure (@yars068 / Yaroslav Shmelev)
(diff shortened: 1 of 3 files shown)


infirit:
Try https://github.com/infirit/blueman/commit/df0c2097be8bb9f083f25d99f8de970de6987bdd.

yars068:
Yor suggestion is simpler than my, but what if need to specify system and user units directories separately?

infirit:
Do you have a system where this is the case? afaik it is not allowed and all unit files go into ``$prefix/lib/systemd/``

yars068:
Then, IMO, it does not make sense, to provide configurabre option to set up path to unit directories, but only to disable it.

infirit:
I pushed the commit, thanks for reporting and the suggestion.

PR #379, opened 2015-10-25
Remove enable_rgba_colormap which does nothing

infirit:
AFAICS there is nothing in blueman that needs to update the GdkVisual (Gtk3's colormap replacement).

commit message:
```
If we need to handle compositing changes we can connect to the
composited-changed signal on GtkWidget and update with the new
GdkVisual.
```


cschramm:
:+1:

Issue #377, opened 2015-10-23
blueman-adapters: New adapters fail to show


infirit:
Does this work for others? Removing seems to be ok.

infirit:
Ok, I tracked it down to [``on_adapter_added``](https://github.com/blueman-project/blueman/blob/master/apps/blueman-adapters#L85) which never calls ``add_to_notebook``.  It seems the local connected ``on_property_changed`` never gets the signal, thus failing to add. I can make it work with the below patch but I am not too familiar with the Bluez.Manager signals so I am not sure this is the right way to go :)

```diff
diff --git a/apps/blueman-adapters b/apps/blueman-adapters
index 8be339b..5e5ed3b 100755
--- a/apps/blueman-adapters
+++ b/apps/blueman-adapters
@@ -60,7 +60,8 @@ class BluemanAdapters:
             self.manager.connect_signal('adapter-added', self.on_adapter_added)
             adapters = self.manager.list_adapters()
             for adapter in adapters:
-                self.add_to_notebook(adapter)
+                path = adapter.get_object_path()
+                self.on_adapter_added(self.manager, path)
         except Exception as e:
             print(e)
             self.manager = None
@@ -83,13 +84,8 @@ class BluemanAdapters:
         Gtk.main_quit()
 
     def on_adapter_added(self, _manager, adapter_path):
-        def on_property_changed(_adapter, name, value, _path):
-            if name == 'Powered' and value:
-                adapter.disconnect_signal(sig)
-                self.add_to_notebook(adapter)
-
         adapter = Bluez.Adapter(adapter_path)
-        sig = adapter.connect_signal('property-changed', on_property_changed)
+        self.add_to_notebook(adapter)
 
     def on_dbus_name_owner_change(self, owner):
         print('org.bluez owner changed to '+owner)
```

yars068:
Unfortunatelly, I cannot check it, because I have a only one adapter :(

cschramm:
To me it looks like the current code adds all adapters on start and additional adapters as soon as they get powered. We have to check why the signal does not appear. It should come from `blueman.bluez.manager._on_interfaces_added`.

What seems strange to me is that initial adapters are added in any case while new adapters are required to be powered first. Not sure why that's the case...

infirit:
>What seems strange to me is that initial adapters are added in any case while new adapters are required to be powered first

[...]

PR #380, opened 2015-10-28
Add dict like functionality to PropertiesBase

infirit:
While working on ``blueman-adapters`` I found it strange that we were storing the adapter properties in a separate dict instead of retrieving them when needed. The first commits allows ``PropertiesBase`` to be used as a dictionary. The second commit implements the new functionality in ``blueman-adapters`` getting rid of all the duplicate storage of the properties.

We can also rename the ``get_properties`` to ``items`` which matches the python dictionary terminology better. Which requires changes across all of blueman so I left this out for now.

The new dict like functionality of-course applies to all classes that derive from ``PropertiesBase``.

### DIFF
--- a/apps/blueman-adapters
+++ b/apps/blueman-adapters
@@ -81,15 +81,14 @@ class BluemanAdapters(Gtk.Dialog):
         self.show_all()
 
     def on_dialog_response(self, dialog, response_id):
-        for hci, settings in self.tabs.items():
-            if settings['changed']:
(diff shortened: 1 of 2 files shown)


infirit:
I added a commit that fixes [this](https://github.com/blueman-project/blueman/issues/377#issuecomment-151245475)

cschramm:
Looks good to me. The underlying issue is probably that connecting a signal on an object creates an internal reference so that the object will stay alive.

PR #389, opened 2015-11-10
Introduce BLUEMAN_SOURCE variable

### LINKED ISSUE #387: Constants.py makes assumptions on directory structure
Lines 31-35 of blueman/Constants.py have

if os.path.exists("../apps") and os.path.exists("../data"):
	BIN_DIR = "./"
	ICON_PATH = "../data/icons"
	PIXMAP_PATH = "../data/icons/pixmaps"
	UI_PATH = "../data/ui"

Obviously, this would fail if blueman-manager (or anyone else importing Functions or Constants) is run in a directory which happens to have those two directory names in it's parent directory (as my ~ does). Is there any way to make this test a bit more specific?

cschramm:
Replaces the different file systems based detection mechanisms for the source tree

Fixes #387

### DIFF
--- a/apps/blueman-adapters
+++ b/apps/blueman-adapters
@@ -9,13 +9,14 @@ import dbus
 from dbus.mainloop.glib import DBusGMainLoop
 from gi.repository import Pango
 import os.path
+import os
 import sys
 import signal
 from locale import bind_textdomain_codeset
 
 #support running uninstalled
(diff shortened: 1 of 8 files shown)


infirit:
Look ok, just two missing imports of ``os`` in blueman-adapters and blueman-manager. It is probably imported indirectly but let's make it explicit :smile:

cschramm:
Added

PR #401, opened 2015-11-29
NetConf: Handle a few corner cases

infirit:
* dhcp handler died and we attempt to kill a non existing process
 * a stale lockfile was found but no pid was populated

### DIFF
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -11,7 +11,7 @@ import errno
 import re
 from socket import inet_aton, inet_ntoa
 from blueman.Constants import *
-from blueman.Functions import have, mask_ip4_address, dprint
+from blueman.Functions import have, mask_ip4_address, dprint, is_running
 from _blueman import create_bridge, destroy_bridge, BridgeException
 from subprocess import call, Popen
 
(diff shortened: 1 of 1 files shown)


cschramm:
Hrm... I agree it's enough to solve point 1, but for point 2 I don't think the issue is a stale lockfile, but we actually just lost track of the PID... The patch solves a crash, but I think we will end up spawning multiple processes (or at least trying, they will probably die as the socket is occupied) which is exactly what we want to avoid with those lockfiles. So we could actually just drop them. :confused: 

Regarding the code, I'd put the unlock line in a `finally` section.

infirit:
Ok, how about..

If self.pid is None we use the pid file (/var/run) to find the pid. Then we can use ``is_running`` from Functions to, 1: check if it is still running and 2: if the name matches what we expect. If everything is in order go and kill the process and use finally to cleanup the lock file.

Will this be enough or should we go a step further and see we have a process name that matches? The risk is that we then start killing handles we did not start..

cschramm:
Sounds good to me. :smiley:

infirit:
Pushed a new version which handles the two cases. That should be it but feel free to find more issues with it :smile:

cschramm:
:+1:

PR #402, opened 2015-12-01
Udhcpd support

infirit:
This work well for me but would be nice if others can confirm this works for them as well.


cschramm:
Works for me.

infirit:
thx, merged

Issue #164, opened 2015-01-05
Close notification after PIN code pairing is completed


cschramm:
Follow-up to #120, @patrakov. The notification never closes as BlueZ does not call the Cancel method although [it should](https://git.kernel.org/cgit/bluetooth/bluez.git/tree/doc/agent-api.txt?id=793fb9ed655c80b43791b46874e7d4525c17eb92#n96).

[This](https://github.com/blueman-project/blueman/blob/bc58e5dc636a5fd3981727025e58b536f49e0ff0/blueman/gui/DeviceList.py#L176) should be the best place to close any left over pairing notification, but we need to keep track of it up to this point or retrieve it from the agent.

infirit:
I am in unfamiliar territory, how would I retrieve the ``BluezAgent``?  Once I get it I can call the Cencel method I guess..

cschramm:
I think the `BluezAgent` plugin should handle this itself. It define a handler in `DisplayPinCode` and `DisplayPasskey` like `device.connect_signal('property-changes', self._on_device_property_changed)` and then doing something like this in the handler:

````python
def _on_device_property_changed(self, device, key, value, path):
  if key == 'Paired' and value:
    self.Cancel()
````

However I really wonder why Cancel does not get called by BlueZ, also it's specified for the API...

infirit:
Ah, yes I was following your initial suggestions but that makes more sense.

The below works for me with my logitech mx5000 keyboard with the notification bubble. I'll try the dialog as well.

Assuming we work around this in blueman, do you want to keep this issue open?

````diff
diff --git a/blueman/main/applet/BluezAgent.py b/blueman/main/applet/BluezAgent.py
index 4fdf5df..ddc3e40 100644
--- a/blueman/main/applet/BluezAgent.py
+++ b/blueman/main/applet/BluezAgent.py
@@ -145,6 +145,11 @@ class BluezAgent(_GObjectAgent, Agent, GObject.GObject):
         self.dialog.connect("response", passkey_dialog_cb)
         self.dialog.present()
 
+    # Workaround BlueZ not calling the Cancel method, see #164
+    def _on_device_property_changed(self, device, key, value, path):
+        if key == "Paired" and value:
+            self.Cancel()
+
     @AgentMethod
     def Release(self):
         dprint("Agent.Release")
@@ -183,6 +188,8 @@ class BluezAgent(_GObjectAgent, Agent, GObject.GObject):
     @AgentMethod
[...]

PR #410, opened 2015-12-12
BluezAgent: Close Notification when pair in successful

infirit:
According to [1] BlueZ should call the Cancel method on the agent but for
some reason it doesn't. See [2]

[1] https://git.kernel.org/cgit/bluetooth/bluez.git/tree/doc/agent-api.txt
[2] https://github.com/blueman-project/blueman/issues/164

### DIFF
--- a/blueman/main/applet/BluezAgent.py
+++ b/blueman/main/applet/BluezAgent.py
@@ -53,6 +53,7 @@ class BluezAgent(_GObjectAgent, Agent, GObject.GObject):
         self.status_icon = status_icon
         self.dialog = None
         self.n = None
+        self.signal_id = None
         self.time_func = time_func
 
         Bluez.AgentManager().register_agent(self, "KeyboardDisplay", default=True)
@@ -145,6 +146,12 @@ class BluezAgent(_GObjectAgent, Agent, GObject.GObject):
         self.dialog.connect("response", passkey_dialog_cb)
         self.dialog.present()
 
+    # Workaround BlueZ not calling the Cancel method, see #164
+    def _on_device_property_changed(self, device, key, value, path):
+        if (key == "Paired" and value) or (key == "Connected" and not value):
+            device.disconnect_signal(self.signal_id)
+            self.Cancel()
+
     @AgentMethod
     def Release(self):
         dprint("Agent.Release")
@@ -183,6 +190,9 @@ class BluezAgent(_GObjectAgent, Agent, GObject.GObject):
     @AgentMethod
     def DisplayPasskey(self, device, passkey, entered):
         dprint('DisplayPasskey (%s, %d)' % (device, passkey))
+        dev = Bluez.Device(device)
+        self.signal_id = dev.connect_signal("property-changed", self._on_device_property_changed)
+
         notify_message = _("Pairing passkey for") + " %s: %s" % (self.get_device_alias(device), passkey)
         self.n = Notification("Bluetooth", notify_message, 0,
                               pixbuf=get_icon("blueman", 48), status_icon=self.status_icon)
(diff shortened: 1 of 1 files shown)


cschramm:
I think we should also disconnect the signal in the handler, because a) it's a memory leak as every device showing such a dialog creates a blueman.bluez.Device object that never gets destroyed and b) there might be an edge case where we cancel the wrong notification / dialog:

* Pairing is initiated for device A -> a notification is shown and a handler is connected.
* Either pairing for A is not completed or it gets unpaired after successfully pairing it -> now it's still / again unpaired, no notification is shown, and the handler resides.
* Pairing is initiated for device B -> a notification is shown (and a handler is connected).

If you now manage to pair A (outside of blueman, e.g. using bluetoothctl), the first handler that's still in place is triggered and cancels the notification for B.

infirit:
> it's a memory leak as every device showing such a dialog creates a blueman.bluez.Device object that never gets destroyed

Does a connected signal count as a reference? If not then the local reference is gone and garbage collection should kick in. To be safe we can make sure to diconnect, more below.

>there might be an edge case where we cancel the wrong notification / dialog

Would that not be the same if BlueZ called the Cancel method on it's own? iow this looks like a design flaw/bug in the BluezAgent not keeping track of the notification. For example [``self.n``](https://github.com/blueman-project/blueman/blob/master/blueman/main/applet/BluezAgent.py#L187) is not checked at all and blindly replaces it with a new notification, or am I missing something? I am not saying we shouldn't handle this, just saying this commit does not change the logic. I can add a dict like I did in [``blueman-adapters``](https://github.com/blueman-project/blueman/commit/89bb6fb0c3f54a556ce4a9e00f0fcb1957675ad5) which holds all the notifications and signal id's and lookup notifications by device path.

cschramm:
> Does a connected signal count as a reference?

Yes, the glib code holds its own reference, so either all handlers must be disconnected or you can delete the object explicitly (`blueman.bluez.Base.__del__` should disconnect the handlers then).

> Would that not be the same if BlueZ called the Cancel method on it's own?

I don't think so. In my example we would trigger `Cancel` for a pairing process our agent did not even handle (as e.g. another handler was set up in `bluetoothctl`). I would not expect BlueZ to trigger `DisplayPasskey` on one agent (`bluetoothctl`'s) and `Cancel` for the same process on another one (blueman's), however we would do that if we have a PropertyChanged handler for the device.

infirit:
Ok, so in short; 
 1. keep track of devices we initiate pairing.
 2. call ``Cancel`` on only those to close the notification.
 3. clean-up after ourselves.

I'll code this up tonight and update this PR.

cschramm:
I think achieving point 3 also fixes the others. It should be enough to keep the latest return value of `dev.connect_signal` to do a `disconnect_signal` in `Cancel`.

With this patch, `Cancel` should now be called no matter what happens to the pairing process, so we're guaranteed to clean-up and to not falsely running into `Cancel` later on.

infirit:
Updated the PR which now clean up. I did end up storing the signal and the object so we can properly disconnect later. I now also included the case were the user may have entered the wrong pin/passkey and we end up calling the Display* functions more than once for the same device.

However, I did think of another potential corner case. What if the user ends up pairing more devices this way? We probably end up with multiple BluezDevice objects of which we can only properly handle the last one.

cschramm:
I thought about that case as well. I think - but I'm in the dark here - that in case pairing for a device was started the agent should always get `Cancel` before getting `Display*` or whatever for another device. However if BlueZ has the concept of multiple parallel pairing processes, this obviously does not hold. If it's really a problem, a solution would be to also disconnect the signal from the last device when connecting it on another device.

Regarding the current PR: You would not need to store the object as it's the first parameter (`device`) for the signal handler (thus the reference). The patch looks good to me. :+1:

infirit:
Fixed the storing of the device and pushed. Thx

PR #412, opened 2015-12-14
Fix building on musl libc

somasis:
\__GLIBC\__ is defined if not defined in order to make the kernel headers work with userspace headers better. musl does not define \__GLIBC\__.

Other distributions usually work around the kernel headers issue by patching kernel headers.

In addition, XCASE must be defined since it is normally provided by glibc's termios.h.

### DIFF
--- a/blueman/main/PPPConnection.py
+++ b/blueman/main/PPPConnection.py
@@ -131,7 +131,7 @@ class PPPConnection(GObject.GObject):
         attrs[0] &= ~(termios.IGNCR | termios.ICRNL | termios.IUCLC | termios.INPCK | termios.IXON | termios.IXANY |
                       termios.IGNPAR)
         attrs[1] &= ~(termios.OPOST | termios.OLCUC | termios.OCRNL | termios.ONLCR | termios.ONLRET)
-        attrs[3] &= ~(termios.ICANON | termios.XCASE | termios.ECHO | termios.ECHOE | termios.ECHONL)
+        attrs[3] &= ~(termios.ICANON | getattr(termios, 'XCASE', 4) | termios.ECHO | termios.ECHOE | termios.ECHONL)
         attrs[3] &= ~(termios.ECHO | termios.ECHOE)
         attrs[6][termios.VMIN] = 1
         attrs[6][termios.VTIME] = 0
--- a/module/libblueman.c
+++ b/module/libblueman.c
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <sys/ioctl.h>
 #include <sys/socket.h>
+#include <sys/time.h>
 #include <unistd.h>
 #include <linux/sockios.h>
 #include <linux/if.h>
--- a/module/modem-prober.c
+++ b/module/modem-prober.c
@@ -34,6 +34,10 @@
 
 #include "modem-prober.h"
 
+#ifndef XCASE
+#define XCASE 0000004
+#endif
+
 #if PY_MAJOR_VERSION >= 3
 #define PyString_FromString PyUnicode_FromString
 #endif


cschramm:
Hrm... Defining `__GLIBC__` looks plain wrong. We neither are glibc, nor do do we rely on that macro, nor can we even expect to have something glibc compatible so we should not pretend to, nor do I think we should fix the incompatibility between the libc and the kernel headers we're built with. That's up to the distribution / maintainer. It's also [documented on the musl side](http://wiki.musl-libc.org/wiki/Building_Busybox) that it expects the kernel headers to be patched.

Concerning XCASE I agree that this needs a patch on the blueman side if we want to provide compatibility with musl libc (and of course we want :smiley:) as it's not POSIX (but all the [libc's musl compares itself to](http://www.etalabs.net/compare_libcs.html) define XCASE). The `modem-prober` patch looks good to me, but I expect issues in the python module `blueman.main.PPPConnection` as well. If the distribution / user (is this some musl based distribution you're working on?) also compiles Python with musl libc, `termios.XCASE` will be undefined and the PPPSupport applet plugin will not work (you will get a crash when a dialup network service shall be connected).

somasis:
That's not really documentation, more just one hack that people have used
to fix it. That's fine though, I'll remove it; that can be passed as a flag
in CFLAGS anyway.

And yes, I'm trying to use blueman on a musl distribution (Exherbo Linux
just recently started supporting musl)

If that's the case with Python though, I'm not sure what should be done to
fix Python's termios stuff.

cschramm:
I can draft a patch for that. Just to confirm my assumption, please fire up a python shell and type:

````python
import termios
termios.XCASE
````

If python was compiled with musl, you'll get an `AttributeError`. Otherwise the result should be `4`.

**Comment by somasis** [comment]
Yeah, you're right.

```
somasis@cicero:~/ % python3
Python 3.5.1 (default, Dec 11 2015, 08:57:37) 
[GCC 5.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import termios
>>> termios.XCASE
  File "<stdin>", line 1, in <module>
AttributeError: module 'termios' has no attribute 'XCASE'
>>> 
```

**Comment by cschramm** [comment]
So here's a simple patch you can include:

````patch
diff --git a/blueman/main/PPPConnection.py b/blueman/main/PPPConnection.py
index 0aab1d9..1a3a992 100644
--- a/blueman/main/PPPConnection.py
+++ b/blueman/main/PPPConnection.py
@@ -131,7 +131,7 @@ class PPPConnection(GObject.GObject):
         attrs[0] &= ~(termios.IGNCR | termios.ICRNL | termios.IUCLC | termios.INPCK | termios.IXON | termios.IXANY |
                       termios.IGNPAR)
         attrs[1] &= ~(termios.OPOST | termios.OLCUC | termios.OCRNL | termios.ONLCR | termios.ONLRET)
-        attrs[3] &= ~(termios.ICANON | termios.XCASE | termios.ECHO | termios.ECHOE | termios.ECHONL)
+        attrs[3] &= ~(termios.ICANON | getattr(termios, 'XCASE', 4) | termios.ECHO | termios.ECHOE | termios.ECHONL)
         attrs[3] &= ~(termios.ECHO | termios.ECHOE)
         attrs[6][termios.VMIN] = 1
         attrs[6][termios.VTIME] = 0
````

**Comment by somasis** [comment]
Thanks. :) Commit amended.

**Comment by cschramm** [comment]
The commit message doesn't fit anymore now. What's the time.h for?

**Comment by somasis** [comment]
time.h is for `tv`'s usage in libblueman.c.

```
/bin/sh ../libtool  --tag=CC   --mode=compile x86_64-pc-linux-musl-cc -DHAVE_CONFIG_H -I. -I..    -pthread -I/usr/x86_64-pc-linux-musl/include/glib-2.0 -I/usr/x86_64-pc-linux-musl/lib/glib-2.0/include   -I/usr/x86_64-pc-linux-musl/include/pygobject-3.0 -I/usr/x86_64-pc-linux-musl/include/glib-2.0 -I/usr/x86_64-pc-linux-musl/lib/glib-2.0/include -I/usr/x86_64-pc-linux-musl/lib/libffi-3.2.1/include   -I/usr/x86_64-pc-linux-musl/include/python3.5m   -DSN_API_NOT_YET_FROZEN -D__GLIBC__ -MT _blueman_la-libblueman.lo -MD -MP -MF .deps/_blueman_la-libblueman.Tpo -c -o _blueman_la-libblueman.lo `test -f 'libblueman.c' || echo './'`libblueman.c
libtool: compile:  x86_64-pc-linux-musl-cc -DHAVE_CONFIG_H -I. -I.. -pthread -I/usr/x86_64-pc-linux-musl/include/glib-2.0 -I/usr/x86_64-pc-linux-musl/lib/glib-2.0/include -I/usr/x86_64-pc-linux-musl/include/pygobject-3.0 -I/usr/x86_64-pc-linux-musl/include/glib-2.0 -I/usr/x86_64-pc-linux-musl/lib/glib-2.0/include -I/usr/x86_64-pc-linux-musl/lib/libffi-3.2.1/include -I/usr/x86_64-pc-linux-musl/include/python3.5m -DSN_API_NOT_YET_FROZEN -D__GLIBC__ -MT _blueman_la-libblueman.lo -MD -MP -MF .deps/_blueman_la-libblueman.Tpo -c libblueman.c  -fPIC -DPIC -o .libs/_blueman_la-libblueman.o
libblueman.c:91:58: warning: 'struct timeval' declared inside parameter list
 static inline unsigned long __tv_to_jiffies(const struct timeval *tv)
                                                          ^
libblueman.c:91:58: warning: its scope is only this definition or declaration, which is probably not what you want
libblueman.c: In function '__tv_to_jiffies':
libblueman.c:95:30: error: dereferencing pointer to incomplete type 'const struct timeval'
         jif = 1000000ULL * tv->tv_sec + tv->tv_usec;
                              ^
libblueman.c: In function '_create_bridge':
libblueman.c:116:17: error: storage size of 'tv' isn't known
  struct timeval tv;
                 ^
Makefile:523: recipe for target '_blueman_la-libblueman.lo' failed
```

somasis:
Amended again to fix the commit message.

PR #419, opened 2015-12-19
Functions: Add function to determine type string from BT Class

infirit:
Inspired by gnome-bluetooth

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -381,3 +381,31 @@ def set_proc_title(name=None):
         dprint("Failed to set process title")
 
     return ret
+
+def bt_class_to_string(bt_class):
+    n1 =  (bt_class & 0x1f00) >> 8
+    if n1 == 0x03:
+        return "network"
+    elif n1 == 0x04:
+        n2 = (bt_class & 0xfc) >> 2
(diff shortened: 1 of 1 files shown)


infirit:
Was a nice exercise for me to convert a C swtich to python :smile:

cschramm:
I think we should reduce this a little bit. In case you did not look at it already, here is a specification for the classes: https://www.bluetooth.com/specifications/assigned-numbers/baseband

Obviously the gnome implementation is not complete. They seem to only care about distinctions they somehow use in their program, but they use it for more than just the PIN database, so they need more distinctions than we do.

The types that may occur in the XML file are in the doctype, see https://github.com/blueman-project/blueman/commit/12f2e28bba01652d15208befe166f8d02d30235f.

We do not need the computer, phone, VCR, video camera, camcorder, joypad, remote control, scanner, wearable, display, camera, and toy cases.

On the other side, audio and headphones are missing in the Audio / Video major class (0x04).

infirit:
>In case you did not look at it already, here is a specification for the classes

I didn't, very helpful, thanks.

I fixed up the other points and also matched the strings returned to those found in the xml file.

cschramm:
:+1:

PR #420, opened 2015-12-19
RFC: Service ui cleanup

infirit:
This is untested right now but will make sure it does not break anything if ppl agree to remove. The ui widgets do nothing and the code is either missing (NAT) or never called (GN), time to clean them out. Agreed?

### DIFF
--- a/blueman/plugins/mechanism/Network.py
+++ b/blueman/plugins/mechanism/Network.py
@@ -40,17 +40,6 @@ class Network(MechanismPlugin):
         except Exception as e:
             err(e)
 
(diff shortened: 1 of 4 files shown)


infirit:
Ok, I looked and the group network bits got removed some time ago in dafa2ac9c71eb694839b61ac09b963c5b525aeb1.

The Other is a left over from an even earlier in 14d1b603a3f28f511a8af4ef26677569cc84b92a so this can go as well.

I'll test these now and push them to master.

cschramm:
:+1:

yars068:
:+1:

PR #424, opened 2016-01-09
Add the type and implment the "max:X" pin entry.

infirit:
I have not tested if the entry is retrieved properly from the xml but this should work.


infirit:
https://github.com/infirit/blueman/commits/defaultpin

PR #427, opened 2016-01-11
RFC: Replace the main Device with the bluez Device object

infirit:
I see ``Device`` class in ``blueman.main.Device`` doing the the following things:

 1.  caches the properties
 2. Handles optional properties missing [see lp bug 327718](https://bugs.launchpad.net/blueman/+bug/327718).
 3. Translates uuid to something meaningful
 4. Add an invalidated signal that wraps the device removed signal from the manager.

Now for the questions :smile:.

Regarding 1, why is this done? If you do not know why not instead handle this in the PropertiesBase or (bluez) Device class?
Regarding 2, We could move this to bluez Device class and handle properties that are optional. Or fix this in RecentConns where the icon is used and what the lp bug is about.
Regarding 3, Move this into the bluez Device class?
Regarding 4, While this is useful why not connect the manager singnal in the right places to handle device removals?

### DIFF
--- a/blueman/main/Device.py
+++ b/blueman/main/Device.py
@@ -8,7 +8,6 @@ import inspect
 from blueman.Functions import dprint
 from blueman.Sdp import uuid128_to_uuid16
 from blueman.bluez import Manager
-from blueman.bluez.Device import Device as BluezDevice
 import blueman.services
 import blueman.services.meta
 import weakref
@@ -26,10 +25,7 @@ class Device(GObject.GObject):
         self.Properties = {}
         self.Temp = False
 
-        if hasattr(instance, "format") and hasattr(instance, "upper"):
-            self.Device = BluezDevice(instance)
-        else:
-            self.Device = instance
+        self.Device = instance
 
         #set fallback icon, fixes lp:#327718
         self.Device.Icon = "blueman"
--- a/blueman/plugins/applet/DBusService.py
+++ b/blueman/plugins/applet/DBusService.py
@@ -8,6 +8,7 @@ from blueman.main.PluginManager import StopException
 from blueman.plugins.AppletPlugin import AppletPlugin
 
 from blueman.main.Device import Device
+from blueman.bluez.Device import Device as BluezDevice
 
 from gi.repository import GLib
 import dbus
@@ -36,7 +37,7 @@ class DBusService(AppletPlugin):
 
     @dbus.service.method('org.blueman.Applet', in_signature="o", out_signature="", async_callbacks=("ok", "err"))
     def DisconnectDevice(self, obj_path, ok, err):
-        dev = Device(obj_path)
+        dev = Device(BluezDevice(obj_path))
 
         self.Applet.Plugins.Run("on_device_disconnect", dev)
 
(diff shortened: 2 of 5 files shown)


infirit:
If we move to GDBus then caching is done by GDBusProxy already so we can remove the caching. And I also saw that several GDBus* object signals provide invalidated properties, not sure if this is the same as what is done here but worth checking out.

cschramm:
1. Don't no. May have made sense with BlueZ 4, but I don't really think so.
2. See below for RecentConns. We also need to look into Class. Looks like it's mainly used in ManagerDeviceList.
3. You mean get_service(s)? I think it's best to keep it separate as blueman.bluez is mainly just a layer abstracting the dbus API and the services do not belong there, so they are in blueman.main. However there's no need to do it in a wrapper Device class. It can also be just simple functions getting the device as an argument.
4. It turns out that it's not really useful, see below. I think it's just a left-over of the BlueZ 4 way.

I drafted a patch to make Service.device be a bluez.Device and in conjunction with that RecentConns work with bluez.Device as well:
https://github.com/blueman-project/blueman/commit/8e59f586a4c0d729e458f4f6c142f3792e66b993

While doing that I thought I realized that at some point your patches make some method expect a bluez.Device now, but there's one case where it's still handed a main.Device. I cannot find that anymore, but I thought my patch fixes that anyway...

Regarding the migration of RecentConns: The code only expects the `device` property of an item to have the `get_object_path` method which is true for both bluez.Device and main.Device, so as long as we do not drop main.Device, using existing data should work. When we drop the class, reading existing data is not possible anymore and we should probably switch to just storing the path instead of an object.

infirit:
> 1 Don't know. May have made sense with BlueZ 4, but I don't really think so.

Ok, when GDBus is done we get this for free anyway so let focus on getting that done :smile:.

> See below for RecentConns.

Cool, added the disconnect in unload and squased it with your.

> 2 We also need to look into Class. Looks like it's mainly used in ManagerDeviceList.

Yep, pushed a patch for that.

> 3 You mean get_service(s)? I think it's best to keep it separate as blueman.bluez is mainly just a layer abstracting the dbus API and the services do not belong there, so they are in blueman.main. 

Yes.

> However there's no need to do it in a wrapper Device class. It can also be just simple functions getting the device as an argument.

Ok, I'll see if I can move this stuff into blueman/services. If you think of a better way or place let me know.

infirit:
I updated the branch you created with the latest changes. The functions for the services are in [blueman/services/Functions](https://github.com/blueman-project/blueman/commit/c98927bd68178934de99db870309303551080882). I have one issue left and that is that blueman manager does not show any icons but I'll fix that later.

Other than the icons I think all instances of main.Device are now gone.

infirit:
Ok, fixed everything up and I think I got it all now. Everything works as master. I like dome feedback and especially on https://github.com/blueman-project/blueman/commit/80f04f550dfae6af93006aaabe1144ecaed3474e which adds a fallback when certain device lack properties.

When I search I see a device that does not have many properties, it is not mine but I still see it. We then end up with a dbus error and we end up with a single line in the list with nothing.

cschramm:
Opened #437 for that branch.

cschramm:
Regarding the dbus error: Do you also get it in current master? If so, which property is requested?

infirit:
>Regarding the dbus error: Do you also get it in current master

No. main.Device caches the properties and adds a fallback for main.Device.Class/Icon in it's init. Here is the [traceback](https://gist.github.com/infirit/72630a90ca7680687e3f).

cschramm:
That's fixed by https://github.com/blueman-project/blueman/commit/a091753213e8d7e44619414af0b1e78c6a136c33, right? Job done? :smiley:

infirit:
Yep, all done :smiley:

PR #428, opened 2016-01-11
RFC: Proof of concept use of GObject properties

infirit:
This allows us to use Device.props and interact with it. While it does not add a whole lot it does give a cleaner way of accessing properties instead of using get/set.

Feel free to shoot this down :wink: .


cschramm:
Like it. :smiley:

baybal:
@infirit can you rerun the CI? The conflict seem to be trivial because of no bluez >5.0

infirit:
I really haven't found a time when I wanted to use properties, closing. If someone want to use it later they can still look at this PR.

PR #437, opened 2016-01-20
Remove main.Device

cschramm:
Here we go then. I additionally removed the main.Device class and made RecentConns persist the object path string instead of a python object. There's a try block which should just ignore existing incompatible data (with persisted main.Device objects).


infirit:
Work ok for me with it removed and changed recentconns :smile:

PR #439, opened 2016-01-20
PropertiesBase: Merge the two property changed functions

infirit:
>_on_property_changed does nothing but emit a signal which we can just
as well do from _on_properties_changed.

So the only reason I can think of to have two functions is the name printed on the console. Other than that I find it pretty useless, any objection to merge the two?


cschramm:
`_on_property_changed` used to be the handler for BlueZ 4.

Go for it. :smile:

PR #440, opened 2016-01-22
Make DiscvManager handle it's own signals

infirit:
While this works I am not sure we are doing the right thing here now. I am contemplating to make the GDBus port handle this by subscribing to the bus anyway but only for the case we want to get all signals for all adapters and devices. I'll experiment a bit with proxies and see if this can be done without interacting directly with the bus.

So this PR is just for quick review and discussion on the validity of doing it like this.

### DIFF
--- a/apps/blueman-applet
+++ b/apps/blueman-applet
@@ -61,9 +61,6 @@ class BluemanApplet(object):
         self.bus = dbus.SystemBus()
         self.bus.watch_name_owner("org.bluez", self.on_dbus_name_owner_change)
 
-        self._any_adapter = Bluez.Adapter()
-        self._any_adapter.connect_signal('property-changed', self._on_adapter_property_changed)
-
         Gtk.main()
 
     def manager_init(self):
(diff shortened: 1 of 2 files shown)


cschramm:
Looks good, but we should drop `AppletPlugin.on_adapter_property_changed` as it's neither implemented nor triggered anywhere now.

infirit:
I think we are doing the wrong thing by moving signal handling into the plugins. The design of the plugin system shows that plugins should not need to handle their own signals but instead the plugin manager runs functions that are the equivalent of connecting to signals (on_adapter_property_changed for example).

Rather than pushing this into the plugins I like to extend this to also include the signals for devices and add the below functions and modify the plugins accordingly. This has the benefit that we have a lot less signals to keep track of (plus disconnect) and it makes the plugins simpler.

 * ``on_device_added``
 * ``on_device_removed``
 * ``on_device_property_changed``

This way we can get rid of things like below sprinkled all over the place.
```python
self._any_device = Bluez.Device()
self._any_device.connect_signal('property-changed', self._on_device_property_changed)
```

Opinions, objections or something else :smile: 

ps: It may also be interesting to think about sending proper Device and Adapter objects on top of/instead of the path. Not sure about that so maybe I'll come back on that later.

PR #445, opened 2016-01-30
Networking: Do not register already registered adapters

infirit:
When a new adapter was added we uncritically register Networking for
all adapters even if we already registered it before.


cschramm:
:+1:

PR #449, opened 2016-02-02
GDBus migration part one

infirit:
This is a partial rewrite from the gdbus branch. Here are the differences:

 * It uses the signals provided by the GDBusProxy objects and does not interact with dbus directly for this any-more.
 * It does not ``prepare_arguments`` and leaves it up to the caller. The main reason is clarity and simplicity, the caller knows what parameters to provide and what it expects back. ``Base._call`` just gets out of the way and let the caller handle things.
 * Both Managers now subclass from GObject instead of PropertiesBase. The reason is simple, Gio's DBusObjectManager handles everything and we do not need to subscribe the bus directly for signals any-more. They only shares a single identical function right now but at some point when this increases we should probably think of having a ManagerBase class.
 * It does not use the get/set_cached_property functions. These only ever interact with the __cached__ properties so when we set a property it only updates the cache not the actual dbus object.
 * It has new Any* clases if there is a need to catch all signals for a adapter, device and network objects. This does not include the changes suggested, not forgotten and looking at the plugin design a bit.

It works ok for me but there are probably small bugs here and there lurking. I would propose we use this as a basis and replace the current gdbus branch with this.

edit: It looks like my phone has file transfer problems so this needs more testing,

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -6,8 +6,6 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import dbus
-from dbus.mainloop.glib import DBusGMainLoop
 import os.path
 import os
 import sys
@@ -28,8 +26,6 @@ gi.require_version("Gtk", "3.0")
 gi.require_version("Pango", "1.0")
 from gi.repository import Gtk
 from gi.repository import Pango
-
-DBusGMainLoop(set_as_default=True)
 
 # Workaround introspection bug, gnome bug 622084
 signal.signal(signal.SIGINT, signal.SIG_DFL)
@@ -57,8 +53,7 @@ class BluemanAdapters(Gtk.Dialog):
         self._adapters = {}
 
         setup_icon_path()
-        self.bus = dbus.SystemBus()
-        self.bus.watch_name_owner('org.bluez', self.on_dbus_name_owner_change)
+        Bluez.Manager.watch_name_owner(self._on_dbus_name_appeared, self._on_dbus_name_vanished)
 
         check_single_instance("blueman-adapters", lambda time: self.present_with_time(time))
 
@@ -110,11 +105,13 @@ class BluemanAdapters(Gtk.Dialog):
         hci_dev = os.path.basename(adapter_path)
         self.remove_from_notebook(self._adapters[hci_dev])
 
-    def on_dbus_name_owner_change(self, owner):
-        print('org.bluez owner changed to '+owner)
-        if owner == '':
-            self.manager = None
-        #fixme: show error dialog and exit
+    def _on_dbus_name_appeared(self, _connection, _name, owner):
+        dprint(owner)
+
+    def _on_dbus_name_vanished(self, _connection, _name, _owner):
+        dprint()
+        self.manager = None
+        # FIXME: show error dialog and exit
 
     def build_adapter_tab(self, adapter):
         def on_hidden_toggle(radio):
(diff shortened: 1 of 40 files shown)


cschramm:
Great work! :+1:

Did not review this in detail, but I think we should get it into master. So the open points where dbus-python is used from what I see are:

Client parts for something else than the BlueZ service:

* apps/blueman-mechanism
* blueman/plugins/applet/NetUsage.py
* blueman/plugins/applet/NMDUNSupport.py
* blueman/plugins/applet/NMPANSupport.py
* blueman/main/AppletService.py
* blueman/main/Mechanism.py

Own services (dbus.service):

* blueman/bluez/Agent.py
* blueman/bluez/obex/Agent.py
* blueman/main/DbusService.py

Usages of DBusGMainLoop (can just be dropped when everything else is gone)

Usages of DBusException. (I guess we're not 100 % consistent on what type of exception is raised and what is caught at the moment.)

infirit:
>Great work!

Thx :smile: 

>but I think we should get it into master

Ok, then let me go through the code and make sure I have not missed things like old imports, old exceptions usage and other things.

infirit:
ok, so it is not working that well when the Agent is needed. It registers properly (not errors in the log) but then does absolutely nothing :cry:. Now if only I knew why...

infirit:
The old dbus-python Agent seems to not cooperate well with the new gdbus AgentManager so I decided to port the agent. I have a __very__ WIP commit that at least shows how this should be done in this PR. It took me some time to figure out how to do this as documentation on it is, well, how should I put it.., terrible, lacking and missing. Had to look at gnome-bluetooth to get going :disappointed: .

But now that I understand, it is actually pretty nice :smile:.

infirit:
Ok, it mostly works now so have a go if one is curious/interested :smile:

infirit:
Pushed also an untested obex agent wip commit (sorry all signals gone here as well :wink: )

cschramm:
The agent works fine for me. I just spotted some issues in the code. `pylint3 -E blueman/bluez blueman/main/applet/BluezAgent.py` (on https://github.com/infirit/blueman/commit/c829c71705819c033acc175ca9d50d3182052723) knows them two.

infirit:
We should ignore those false positive exception errors for ```GLib.Error```. The others in the BluezAgent should be fixed now.

cschramm:
I still get 

> ************* Module blueman.bluez.PropertiesBase
E: 18,35: Undefined variable 'unicode' (undefined-variable)

in https://github.com/infirit/blueman/commit/0bf13822f8d35b6276f6e6a8bb6fdb19343a11e5.

infirit:
That is python2 only code to remain compatible. It does not affect python3.

cschramm:
Ah, sure, missed that.

infirit:
Just pushed a new set of commits that fixes lots (especially in blueman.bluez.error) of small problems. The obex agent stuff still broken but the rest is looking good now.

infirit:
Ok, so finally figured why the obex agent was not working, basically we were watching the wrong bus (see https://github.com/blueman-project/blueman/pull/449/commits/1a5477ef3ed2e7ccfc19e381db65aadaabaec5f7). In the next week I'll clean things up and try to refactor some of the code to reduce code duplication.

infirit:
Refactor done so now more testing and it is ready for master.

infirit:
Ok, so fixed some bugs and I think it is in a reasonable shape now. Some points from earlier in the thread.

>Usages of DBusGMainLoop (can just be dropped when everything else is gone)

Yes, I removed them from the parts ported to GDBus.

>Usages of DBusException. (I guess we're not 100 % consistent on what type of exception is raised and what is caught at the moment.)

I opted to parse the errors instead of catching them directly in the app and plugins. The BlueZ classes should only return parsed errors, anytime we get an unparsed one we probably have a bug.

Next up for me are the NM client bits for which I'll open a new PR.

cschramm:
Perfect. This should be merged then. :smiley:

PR #452, opened 2016-02-05
BluezAgent: Drop the unused signal "released"

infirit:
It is unused and without it things get a whole lot less complicated.


cschramm:
:smile: I liked it...

infirit:
Ok, slightly confused and unsure :smiley:. You are ok with the change?

cschramm:
Sure, makes not sense to keep this.

PR #455, opened 2016-02-10
blueman-manager: Fix authentication failed handling

### LINKED ISSUE #453: We do not handle authentication failed
I initially thought this was due to my gdbus work (confused the hell out of me) but we have the same problem in master. I have not checked the 2-0-stable branch so we may need to fix it there as well.

Documenting so we do not forget :smile:.

```python
_________
RequestConfirmation (/usr/lib64/python3.4/site-packages/blueman/main/applet/BluezAgent.py:201)
Agent.RequestConfirmation 
_________
DBusException(dbus.String('Authentication Failed'),)
Cancel (/usr/lib64/python3.4/site-packages/blueman/main/applet/BluezAgent.py:153)
Agent.Cancel 
'org.bluez.Error.AuthenticationFailed: Authentication Failed'
['org.bluez.Error.AuthenticationFailed:', 'Authentication Failed']
  File "/usr/lib64/python3.4/site-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
  File "/usr/lib64/python3.4/site-packages/blueman/bluez/Device.py", line 26, in err
    self._call('Pair', reply_handler=ok, error_handler=err)
blueman.bluez.errors.DBusAuthenticationFailedError: Authentication Failed
```

infirit:
This picks a commit that changes the handling of bluez methods from the gdbus migration. It is useful as it allows for more things to be pushed down into ``blueman.bluez.Base`` and simplify things.

Just like in my gdbus branch we now check if we either have an error or reply handler and make the call asynchronous.

Then we show a notification when we fail to pair in blueman-manager. Please let me know if we need to specify what error we got in the notification.

This fixes #453

### DIFF
--- a/apps/blueman-manager
+++ b/apps/blueman-manager
@@ -31,6 +31,7 @@ from blueman.gui.manager.ManagerProgressbar import ManagerProgressbar
 from blueman.main.Config import Config
 
 from blueman.gui.MessageArea import MessageArea
+from blueman.gui.Notification import Notification
 
 from blueman.main.PluginManager import PluginManager
 import blueman.plugins.manager
@@ -197,7 +198,12 @@ class Blueman(Gtk.Window):
         launch(command, None, False, "blueman", _("Bluetooth Assistant"))
 
     def bond(self, device):
-        self.List.Adapter.find_device(device['Address']).pair()
+        def error_handler(e):
+            dprint(e)
+            message = 'Pairing failed for:\n%s (%s)' % (device['Alias'], device['Address'])
+            Notification('Bluetooth', message, pixbuf=get_icon("blueman", 48))
+
+        device.pair(error_handler=error_handler)
 
     def adapter_properties(self):
         adapter = self.List.Adapter
--- a/blueman/bluez/Adapter.py
+++ b/blueman/bluez/Adapter.py
@@ -6,47 +6,39 @@ from __future__ import unicode_literals
 from gi.repository import GObject
 from blueman.Functions import dprint
 from blueman.bluez.PropertiesBase import PropertiesBase
-from blueman.bluez.errors import raise_dbus_error
 from blueman.bluez.Device import Device
 import dbus
 
 
 class Adapter(PropertiesBase):
-    @raise_dbus_error
     def __init__(self, obj_path=None):
         super(Adapter, self).__init__('org.bluez.Adapter1', obj_path)
         proxy = dbus.SystemBus().get_object('org.bluez', '/', follow_name_owner_changes=True)
         self.manager_interface = dbus.Interface(proxy, 'org.freedesktop.DBus.ObjectManager')
 
-    @raise_dbus_error
     def find_device(self, address):
         devices = self.list_devices()
         for device in devices:
             if device.get_properties()['Address'] == address:
                 return device
 
-    @raise_dbus_error
     def list_devices(self):
-        objects = self.manager_interface.GetManagedObjects()
+        objects = self._call('GetManagedObjects', interface=self.manager_interface)
         devices = []
         for path, interfaces in objects.items():
             if 'org.bluez.Device1' in interfaces:
                 devices.append(path)
         return [Device(device) for device in devices]
 
-    @raise_dbus_error
     def start_discovery(self):
-        self._interface.StartDiscovery()
+        self._call('StartDiscovery')
 
-    @raise_dbus_error
     def stop_discovery(self):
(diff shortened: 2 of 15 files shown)


cschramm:
Not sure why my commit is in here.

Changes look good, but some thoughts:

I think using `_call` with a `reply_handler` but no `error_handler` is invalid and should be forbidden. It does not make sense to raise the exception in the asynchronously called `err` function.

It looks like we only do asynchronous calls so we should probably make `error_handler` mandatory. This saves us from synchronous calls that do not catch errors.

infirit:
>Not sure why my commit is in here

I took it from the gdbus branch as It makes things easier to implement, everything is now done in ``Base._call``.

>I think using _call with a reply_handler but no error_handler is invalid and should be forbidden.

Ok, I can do an ``if reply_handler: assert(error_handler != None)`` and it will raise an assertion error.

>It does not make sense to raise the exception in the asynchronously called err function.

So, it most likely does this as a way of hand holding the caller. It checks if it received a callable (function usually) and if not raises the error. I can drop the callable check and we would get a ``TypeError`` if it is a non callable object.

>It looks like we only do asynchronous calls so we should probably make error_handler mandatory

We do synchronous calls in Adapter, for example Stop/StartDiscovery. We could add handlers all over the place but I am not sure if that makes sense everywhere..? If we fail to call ``StartDiscovery`` for example, what should the error handler do?

cschramm:
>> It does not make sense to raise the exception in the asynchronously called err function.
>
> So, it most likely does this as a way of hand holding the caller. It checks if it received a callable (function usually) and if not raises the error. I can drop the callable check and we would get a TypeError if it is a non callable object.

The thing is the exception will get nowhere as it ends up in the main loop.

> We do synchronous calls in Adapter, for example Stop/StartDiscovery. We could add handlers all over the place but I am not sure if that makes sense everywhere..? If we fail to call StartDiscovery for example, what should the error handler do?

Right, there are actually quite some cases. Also in NetworkServer and AgentManager. Well, an error handler should at least forward the error to the user through the GUI. Otherwise he experiences a silent failure or a crash and does not even get to know that there's a problem, unless he checks the blueman-applet output. On the other hand it looks like NetworkServer should not raise any errors at all, AgentManager only has errors that result from wrong usage (our fault), and Adapter errors are at least not likely, so it's probably ok that we do not have handlers there. But maybe we should require an `error_handler` in all cases where the keyword argument exists or are there cases where it is unused?

infirit:
Just to be clear, my intention is not to be difficult, just trying to understand and get thing clear in my mind :smile:.

>The thing is the exception will get nowhere as it ends up in the main loop

I am not sure why this is bad? There is, I think, no way we can handle me messing up with an error handler :smile:. But I do see there is a case for adding more strict error handling, more below.

>Well, an error handler should at least forward the error to the user through the GUI

Agreed, assuming we are talking about the gui bits, not mechanism. I do think this is a bit out of scope for this specific PR/problem and needs some more thinking on how to implement this properly.

>But maybe we should require an ```error_handler``` in all cases where the keyword argument exists or are there cases where it is unused?

Yes that sounds reasonable. In the pair case there are well defined errors in the [api](https://git.kernel.org/cgit/bluetooth/bluez.git/tree/doc/device-api.txt#n87) we should be able to handle. But, I do see on the Adapter and Agent apis there are errors we do not handle at all, should we? If so we probably should review all bluez errors and come up with a plan :smile:.

I do think that in the unexpected error case we should (re)raise (the) an exception in the handlers. I do not like them to be silently ignored just because we did not think of the use case.

Perhaps we can have some sort of base class from which we create all handlers?

----
For this specific PR, I am thinking to not merge b70a98e184800e59b5ced08e0094c062b82d2a15 but we do leave the assertion 4bfe7fb16623788c5665c543988b7818cfd0d0d8. You ok to merge it like that, if so I'll open new issues for the handler rework and/or bluez error review :smiley:.

cschramm:
> For this specific PR, I am thinking to not merge b70a98e but we do leave the assertion 4bfe7fb. You ok to merge it like that, if so I'll open new issues for the handler rework and/or bluez error review :smiley:.

Yes, fine with that, although it should only differ in the case where error_handler is set but not callable which should not happen at all.

infirit:
Ok, issues created and merged this PR. Just one question, should I push this also to 2-0-stable? It's not a major problem and does not affect the user afaik so I am in doubt :smile:

cschramm:
Better not. :laughing:

PR #456, opened 2016-02-10
Extend the function we have the PluginManager run

infirit:
Applet plugin no longer need to connect their own signals for Adapter and Devices. This is how the plugin system was for the adapter already, we now extend this to the devices too.

See  #440 

There is also a couple of small cleanups in here (first two).


cschramm:
:+1:

PR #462, opened 2016-02-14
DeviceList: cleanups round one

infirit:
First set of cleanups.

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -41,7 +41,6 @@ import atexit
 import sys
 import errno
 from ctypes import cdll, byref, create_string_buffer
-from subprocess import Popen
 from gi.repository import GLib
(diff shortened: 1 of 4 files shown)


cschramm:
The assumption in 75f2e10 is not true in case of the device-created signal from the Manager object, is it?

infirit:
There are a couple of cases where we call ``device_add_event`` which is the only one that calls ``add_device``.

 * It is a brand new discovered bluetooth device and it will not exist in the list.
 * We switch to another adapter and we clear the old list.
 * We reorder the list and we again first clear the list and then rebuild.

cschramm:
Brand new, sure. :laughing: 

No more objections then. :+1:

infirit:
Thanks, merged :smile:.

Issue #460, opened 2016-02-13
Trying to get properties from removed device


infirit:
What happens,

 * Discover at least __two__ new devices and select one of them. It may also happen with a single removed device but I have not yet been able to reproduce that.
 * Bluez tries expired them because they were not paired or connect, we properly notice and remove them from the list. 
 * We respond to the selection being changed by the removal of the first one
 * Selected changes to the second removed device
 * We try to lookup uuids to update the send file button and fail

It looks like this is what the ``Valid`` property was for on ``blueman.main.Device`` and now that is gone we run into trouble.

There are a couple of options to fix this.

 1. Re-introduce the ``Valid`` property and set it to False when a device is removed. Then add the necessary checks in the right places.
 2. Add an error_handler (and new eerror DBusObjectUnknown) to PropertiesBase and handle the case and return None (or something else).
 3. Something I did not think off :smile:.

```python
_________
on_device_selected (/usr/lib64/python3.4/site-packages/blueman/gui/manager/ManagerToolbar.py:84)
*** <ManagerDeviceList object at 0x7f843b9b3ea0 (blueman+gui+manager+ManagerDeviceList+ManagerDeviceList at 0x244c430)> <Device object at 0x7f842dd22828 (blueman+bluez+Device+Device at 0x21786a0)> <GtkTreeIter at 0x2181140> on_selection_changed *** 
Traceback (most recent call last):
  File "/usr/lib64/python3.4/site-packages/blueman/gui/manager/ManagerToolbar.py", line 117, in on_device_selected
    self.update_send(device)
  File "/usr/lib64/python3.4/site-packages/blueman/gui/manager/ManagerToolbar.py", line 122, in update_send
    for uuid in device['UUIDs']:
  File "/usr/lib64/python3.4/site-packages/blueman/bluez/PropertiesBase.py", line 61, in __getitem__
    return self.get(key)
  File "/usr/lib64/python3.4/site-packages/blueman/bluez/PropertiesBase.py", line 43, in get
    raise e
  File "/usr/lib64/python3.4/site-packages/blueman/bluez/PropertiesBase.py", line 38, in get
    prop = self.__properties_interface.Get(self._interface_name, name)
  File "/usr/lib64/python3.4/site-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/lib64/python3.4/site-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
[...]

PR #464, opened 2016-02-16
RFC: Store if device has object push capability in GtkListStore

infirit:
This relates to issue comnment https://github.com/blueman-project/blueman/issues/460#issuecomment-184472046.

This does not solve all the errors we get trying to get properties from a removed device though and we need to do the same thing for the audio service. We could also store all the uuids and work of them instead of reading them over and over.

This is a proof of concept to store values in the GtkTreeStore for use in the UI.

Opinions, objection or suggestions (especially on my naming skills :smile:).

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -41,6 +41,7 @@ class ManagerDeviceList(DeviceList):
             # trusted/bonded icons
             # ["tb_icons", 'PyObject', CellRendererPixbufTable(), {"pixbuffs":5}, None],
 
+            ["alias", str], # used for quick access instead of device.GetProperties
             ["connected", bool], # used for quick access instead of device.GetProperties
             ["bonded", bool], # used for quick access instead of device.GetProperties
             ["trusted", bool], # used for quick access instead of device.GetProperties
@@ -235,7 +236,7 @@ class ManagerDeviceList(DeviceList):
         caption = self.make_caption(name, klass, address)
 
         # caption = "<span size='x-large'>%(0)s</span>\n<span size='small'>%(1)s</span>\n<i>%(2)s</i>" % {"0":name, "1":klass.capitalize(), "2":address}
-        self.set(tree_iter, caption=caption, orig_icon=icon)
+        self.set(tree_iter, caption=caption, orig_icon=icon, alias=name)
 
         try:
             self.row_update_event(tree_iter, "Trusted", device['Trusted'])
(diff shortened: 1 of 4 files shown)


cschramm:
Hmm... so your strategy would be to harden any use of device properties? Nor sure how deep that barrel is...

I'm also not sure if caching makes sense here. UUIDs may especially change when the device is getting (un)paired or (dis)connected. Afaik an OBEX service is usually also available on an unpaired device, but not necessarily.

infirit:
To avoid confusion I think we need to be clear on who exposes what and when. I see that UUIDs expose services on the device and then we have bluez exposing services through dbus. And that we can have a UUID expose a service on the device which has not yet been exposed by bluez over dbus. Which is why we catch and handle the error [1] in ``NetworkService`` when the service ``org.bluez.Network1`` is only provided after pairing.

But your question made me think about and check how this is done for other things stored in the list and everything is already in place, see below for more.

>so your strategy would be to harden any use of device properties

Not really that but store what we need in the UI just like we do for connected, trusted etc. It makes for easy access and solves the problem with removed devices as we do not rely solely on dbus. The code already handles when those properties update/change so extending it to objectpush (or UUIDs really) should be easy.

In fact it already handles changed UUIDs and updates the send file button accordingly, but I have never seen it run. The only difference is that instead of looping over the UUID's on the fly we store something in the list. I have pushed an updated commit to show how this would look like.

Storing all the UUID's may be a better option as we may need them for more than objpush in the future and we only add one new entry to the list.

[1]
```
_________
connected (/usr/lib64/python3.4/site-packages/blueman/services/meta/NetworkService.py:18)
Could not get properties of network service: org.freedesktop.DBus.Error.InvalidArgs: No such interface 'org.bluez.Network1' 
_________
Generate (/usr/lib64/python3.4/site-packages/blueman/gui/manager/ManagerDeviceMenu.py:172)
Galaxy Nexus 
```

cschramm:
That new commit makes absolute sense to me. :smiley:

infirit:
Ok, I think I got all situation covered now. I will shuffle/split/merge commits so it makes more sense than it does now. Feel free to ask more questions :smile:.

I like some feedback on https://github.com/infirit/blueman/commit/977d8398c3641d00445f8396d0bc2f8b4e9c734e, while it works I am not sure this is the best way to do this. If you have suggestions please let me know.

cschramm:
Guess the services part is fine. We cannot do much more than returning an empty list if we get there.

infirit:
Took a bit longer to get myself to clean this up.. It works well for me now and have not seen backtraces or regressions. I'll merge it in few days time unless there are objections :smile:

Issue #104, opened 2014-09-27
blueman-adapters: Setting "friendly" name broken in both master and gsettings branch


infirit:
``` python
Traceback (most recent call last):
  File "/usr/bin/blueman-adapters", line 78, in on_dialog_response
    adapter.SetProperty('Name', settings['name'])
AttributeError: 'Adapter' object has no attribute 'SetProperty'
```

@cschramm If you have a working 1.23 version can you please check if it is also broken there?

PR #466, opened 2016-02-18
Big cleanup and rework of Adapter menu in ManagerMenu class

infirit:
These commits change the way we create and update the Adapter menu. The major changes are,

 * No more regenerating the __complete__ Adapter menu on every change (add/remove/select etc).
 * Use a bluez.Manager object to add and remove adapter menu entries
 * Move the creation of the adapter menu items to on_adapter_added/removed
 * Connect and disconnect the ``property-changed`` signal on the bluez.Adapter objects when added/removed. We do not rely on the DeviceList for this any-more.

There should be no functional or visual changes so let me know if you run into any of them.

Why all this work you may ask, well it came about my idea to have different sorting options (by time added, alias, ascending/decending) for the DeviceList (and the current one is horrendous). And when one starts poking at things one finds lots of things that probably made sense in the past but are now getting in the way :smile:.

What I like to do next is do more work on the ManagerDeviceList and change the way we use the strore and display the view. The idea is to have a single list with __all__ devices from __all__ adapters and then have a filter function to only project the rows that are relevant to the currently selected adapter the view.

ps: i'll also finish things up on #464 in the next couple of days.

### DIFF
--- a/blueman/gui/manager/ManagerMenu.py
+++ b/blueman/gui/manager/ManagerMenu.py
@@ -140,39 +140,10 @@ class ManagerMenu:
         blueman.List.connect("device-selected", self.on_device_selected)
         blueman.List.connect("adapter-changed", self.on_adapter_changed)
 
-        self.on_adapter_changed(blueman.List, blueman.List.GetAdapterPath())
+        for adapter in self._manager.list_adapters():
+            self.on_adapter_added(None, adapter.get_object_path())
 
         self.device_menu = None
-
-    def _add_adapter_menu_item(self, adapter_path):
-        adapter = bluez.Adapter(adapter_path)
-        menu = self.item_adapter.get_submenu()
-        object_path = adapter.get_object_path()
-
-        item = Gtk.RadioMenuItem.new_with_label(self._adapters_group, adapter['Alias'])
-        item.show()
-        self._adapters_group = item.get_group()
-
-        item_sig = item.connect("activate", self.on_adapter_selected, object_path)
-        adapter_sig = adapter.connect_signal("property-changed", self.on_adapter_property_changed)
-
-        menu.insert(item, self._insert_adapter_item_pos)
-        self._insert_adapter_item_pos += 1
-
-        self.adapter_items[object_path] = (item, item_sig, adapter, adapter_sig)
-
-        if adapter_path == self.blueman.List.Adapter.get_object_path():
-            item.props.active = True
-
-    def _remove_adapter_menu_item(self, adapter_path):
-        item, item_sig, adapter, adapter_sig = self.adapter_items.pop(adapter_path)
-        menu = self.item_adapter.get_submenu()
-
-        item.disconnect(item_sig)
-        adapter.disconnect_signal(adapter_sig)
-
-        menu.remove(item)
-        self._insert_adapter_item_pos -= 1
 
     def on_device_selected(self, List, device, tree_iter):
         if tree_iter and device:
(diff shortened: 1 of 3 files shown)


blueman/gui/manager/ManagerMenu.py
```diff
@@ -18,88 +18,115 @@ class ManagerMenu:
...
         self.item_help.set_submenu(help_menu)
         help_menu.show()
 
-        item = create_menuitem(_("_Report a Problem"), get_icon("dialog-warning", 16))
-        item.connect("activate", lambda x: launch("xdg-open %s/issues" % WEBSITE, None, True))
-        help_menu.append(item)
-        item.show()
+        report_item = create_menuitem(_("_Report a Problem"), get_icon("dialog-warning", 16))
+        report_item.show()
+        help_menu.append(report_item)
```
cschramm: Theoretically you display a visible item in an already visible menu here that does not react to its activation yet. Just mentioning, not relevant as nobody will click that fast. ;)

blueman/gui/manager/ManagerMenu.py
```diff
@@ -132,85 +158,56 @@ def on_device_selected(self, List, device, tree_iter):
...
 
-            item.show()
-            menu.prepend(item)
+        self.adapter_items[object_path] = (item, item_sig, adapter, adapter_sig)
 
-        sep = Gtk.SeparatorMenuItem()
-        sep.show()
-        menu.prepend(sep)
+        if adapter_path == self.blueman.List.Adapter.get_object_path():
+            item.props.active = True
```
cschramm: This property must be updated when `blueman.List` emits `adapter-changed`.

blueman/gui/manager/ManagerMenu.py
```diff
@@ -132,85 +158,56 @@ def on_device_selected(self, List, device, tree_iter):
...
+    def on_adapter_added(self, _manager, adapter_path):
+        adapter = bluez.Adapter(adapter_path)
+        menu = self.item_adapter.get_submenu()
+        object_path = adapter.get_object_path()
 
-        settings = create_menuitem("_Preferences", get_icon("preferences-system", 16))
-        settings.connect("activate", lambda x: self.blueman.adapter_properties())
-        settings.show()
-        menu.append(settings)
+        item = Gtk.RadioMenuItem.new_with_label(self._adapters_group, adapter['Alias'])
```
cschramm: https://github.com/blueman-project/blueman/issues/104

cschramm:
It looks like adapter items do not get activated and deactivated as adapter-changed is not handled.

blueman/gui/manager/ManagerMenu.py
infirit: I remember that although looking at it now I am not sure set/get_name is really needed. If I ask for the Alias while it is not set I do not get an error I get the name property. See below

```python
>>> adapter['Alias'] = ''
>>> adapter['Alias']
dbus.String('BlueZ 5.37', variant_level=1)
>>> adapter['Alias'] = 'BluezMine'
>>> adapter['Alias']
dbus.String('BluezMine', variant_level=1)
>>> adapter['Alias'] = ''
>>> adapter['Alias']
dbus.String('BlueZ 5.37', variant_level=1)
```

cschramm: May have been added in some recent BlueZ version. Maybe Alias just did not exist in BlueZ 4.

blueman/gui/manager/ManagerMenu.py
infirit: I looked at this before changing and the one actually calling ``SetAdapter`` (which emits the signal) on the List is this same menuitem through ``on_adapter_selected`` :smile:. Also we now handle adapter-added/removed locally and set the select the active adapter accordingly here.

blueman/gui/manager/ManagerMenu.py
infirit: Looking back at the issue it comes from the time that the name property was writeable under bluez4 but no under bluez5 (which has the Alias property now). I'll change it to set/get_name to be safe :smile:. Any objection that I add a FIXME to these function so we can revisit this in the future?

blueman/gui/manager/ManagerMenu.py
cschramm: Hm, you're right, this looks broken overall... :confused:

blueman/gui/manager/ManagerMenu.py
cschramm: Of course not.

infirit:
Ok, pushed a couple commits per discussion, if nothing else is left I'll push this later today :smile: .

PR #470, opened 2016-02-24
Add new UUIDs to Sdp

cschramm:


infirit:
Added a few more updates and some small fixes. Also point people to the bluez and official sdp sources. My phone also listed generic access/attribute UUID's so added those as well for completeness.

If no objection's I'll merge this later today. Thanks :smile:

cschramm:
:+1:

PR #472, opened 2016-02-25
Bluez: Implement singleton for bluemans bluez objects

infirit:
So this does not work as we do not have the necessary info to properly keep track of the objects.

The main problem is that we do not have the ``interface_name`` in ``*args`` or ``**kwargs`` provided to ``__new__`` because we do not create the object with it and provide it as default (keywword)argument. We could force all instances to use keywords and provide the both interface_name and object_path like, ``bluez.Device(interface_name='org.bluez.Device1', obj_path='/org/bluez/...')``.

Another way is to have some sort of BluezObject store/manager that is the only one creating and distributing the objects.

### DIFF
--- a/blueman/bluez/Adapter.py
+++ b/blueman/bluez/Adapter.py
@@ -12,8 +12,10 @@ import dbus
 
 
 class Adapter(PropertiesBase):
-    def __init__(self, obj_path=None):
-        super(Adapter, self).__init__('org.bluez.Adapter1', obj_path)
+    _interface_name = 'org.bluez.Adapter1'
+
+    def _init(self, obj_path=None):
+        super(Adapter, self)._init(interface_name=self._interface_name, obj_path=obj_path)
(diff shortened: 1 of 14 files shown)


cschramm:
If we have one interface name per class (looks like it), we could specify it as a class attribute. We would then have e.g. `Adapter._interface_name = 'org.bluez.Adapter1'` and `__new__` could use `cls._interface_name`.

infirit:
Let me try that out, good one thanks :smile:.

infirit:
Perfect, this works well now :smiley:. Pushed the working commit to this PR if you want to try it out.

cschramm:
Awesome! Small change, big effect. :smiley:

infirit:
Pushed 19d3908 that parses the arguments and fixes a possible case where we are provided an interface name. With that everything works well for me.

I'll run it for a couple of days before merging, unless you have issues with this :smile:.

PR #473, opened 2016-02-27
New MediaPlayer plugin and service

infirit:
This is a pet project of me to add a new plugin that contols media player devices over bluetooth. I can control my android phone with it properly at this point but the UI is still lacking in lots of places. Lots of check and validation has to be added but folks may find it interesting to see and/or try out :smile:.

**To be clear, this is by now means done and not meant to be merged any time soon.**

edit: rebased against master 19d39087557339fb8021e674c8fd511122e49efc 2016-03-03

### DIFF
--- a/blueman/bluez/AVRemote.py
+++ b/blueman/bluez/AVRemote.py
@@ -1,0 +1,40 @@
+# coding=utf-8
+from __future__ import print_function
+from __future__ import division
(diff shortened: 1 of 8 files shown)


cschramm:
Nice! I thought about this once, but did not find the motivation to actually start it.

Issue #482, opened 2016-03-14
Shebang: blueman uses hard-coded "#!/usr/bin/env python"


MingcongBai:
When building Blueman on AOSC OS, a distribution using Python 2.7 as its default Python interpreter, Blueman will refuse to start after specifying `PYTHON=/usr/bin/python3` as an option for `configure`, because the installed programs would still contain `#!/usr/bin/env python` as its Shebang. As reported in the issue below.

https://github.com/AOSC-Dev/aosc-os-abbs/issues/170

Our build manifests can be found here (with workaround for shebang issue).

https://github.com/AOSC-Dev/aosc-os-abbs/tree/staging/extra-utils/blueman

cschramm:
See #435. Distributions usually replace the shebang. In Gentoo they have the python_fix_shebang tool, in Debian / Ubuntu debhelper does that job.

While we could set the shebang to $PYTHON during configure / make, I think that would be wrong as what you set with that variable is the python interpreter to be used during build. So it is the one on the build system, while the one on the target system which is set by the shebang could be different.

However, I guess it's a good idea to set either `#!/usr/bin/env python2` or `#!/usr/bin/env python3` depending on the major version of $PYTHON. I'll draft a patch for that.

MingcongBai:
> See #435. Distributions usually replace the shebang. In Gentoo they have the python_fix_shebang tool, in Debian / Ubuntu debhelper does that job.

So I did. While it really shouldn't have happened.

> So it is the one on the build system, while the one on the target system which is set by the shebang could be different.

Well that was usually done in GNOME packages and they would work flawlessly as to set Shebang to the `PYTHON` variable specified during build.

> However, I guess it's a good idea to set either #!/usr/bin/env python2 or #!/usr/bin/env python3 depending on the major version of $PYTHON. I'll draft a patch for that.

Yes it is, thanks for the quick response mate XD.

cschramm:
> Well that was usually done in GNOME packages and they would work flawlessly as to set Shebang to the PYTHON variable specified during build.

That does not make sense to me. Think about a complex build system that sets $PYTHON to say `/var/buildsystem/python/3.4.4/python` while the target system just has `/usr/bin/python(3(.4))`, so if $PYTHON would be used in the shebang, it would not be found, but `/usr/bin/env python(3)` will.

Patch proposal for major versions is in #483. It should fix your specific case up to the point where you add pyenv to the distribution and have to start replacing Python shebangs anyway.

[...]

Issue #435, opened 2016-01-18
pyenv maybe doesn't like usage of "env python" in blueman     


rworkman:
We got a report of problems with pyenv due to blueman's usage of "#!/usr/bin/env python" instead of "#!/usr/bin/python" - I'm not suggesting that this *is* a bug with blueman; I'm more interested in thoughts and how to address it with the reporter: http://www.linuxquestions.org/questions/slackware-14/current-blueman-using-env-sourced-python-4175564239/

infirit:
I am not sure we want to support switching version on the fly for blueman as we build a c module. Switching within the same major would probably be ok but I am pretty sure things go badly when switching major version.

infirit:
There is no abi guarantee for python2 and a limited one for python3 so I would be very careful with switching version.

What most distro's do it rewrite the shbang so it works for them. Gentoo ppl of-course take this to the extreme and basically solved this problem already through rewriting shbang [1], python-exec wrapper and building for any version installed or needed/wanted.

[1] https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/python-utils-r1.eclass#n1059

zakame:
I'm the one who reported my problem on the Slackware forum; let me post in more detail here.

I noticed this first when Slackware-current updated blueman to the latest current stable (2.0.3) where in my Fluxbox startup I had `/usr/bin/blueman-applet &`, which failed.  Upon running on an xterm, I saw

```
θ61° [zakame@drift:~] % /usr/bin/blueman-applet
Traceback (most recent call last):
  File "/usr/bin/blueman-applet", line 12, in <module>
    import dbus
ImportError: No module named 'dbus'
```

So naturally I checked the script, and saw it had a `#!/usr/bin/env python` shebang.  Running with system python set explicitly seems to work:

```
θ80° [zakame@drift:~] % /usr/bin/python /usr/bin/blueman-applet
/usr/bin/blueman-applet:20: PyGIWarning: Notify was imported without specifying a version first. Use gi.require_version
('Notify', '0.7') before import to ensure that the right version gets loaded.                                         
  from gi.repository import Notify
blueman-applet version 2.0.3 starting
_________
Load (/usr/lib64/python2.7/site-packages/blueman/main/PluginManager.py:60)
['NetUsage', 'DBusService', 'Networking', 'RecentConns', 'DiscvManager', 'Headset', 'ExitItem', 'GameControllerWakelock
[...]

PR #483, opened 2016-03-14
Put Python major version in shebang

cschramm:
This allows building for systems with both Python 2 and 3 without replacing the shebang. Replacing will only be necessary if env shall not be used e.g. as PyEnv might be present (see #435) or a specific minor version needs to be defined (i.e. there are multiple minor versions on the target system and blueman is not built for the newest one; we cannot use minor versions as they are not mandatory, see PEP 394).

This is a solution for #482 (build for AOSC OS without shebang replacement), however it only makes sense for very few distributions that neither have pyenv and nor multiple Python minor versions. More major distributions have some shebang replacement mechanism anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,15 @@ Makefile
 /config.status
 /libtool
 /stamp-h1
+/apps/blueman-adapters
+/apps/blueman-applet
+/apps/blueman-assistant
+/apps/blueman-manager
+/apps/blueman-mechanism
+/apps/blueman-report
+/apps/blueman-rfcomm-watcher
+/apps/blueman-sendto
+/apps/blueman-services
 /blueman/Constants.py
 /blueman/plugins/mechanism/Rfcomm.py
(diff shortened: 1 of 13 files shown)


.gitignore
```diff
@@ -35,6 +35,7 @@ Makefile
 /config.status
 /libtool
 /stamp-h1
+/apps/blueman-*
```
infirit: This includes all files beginning with blueman-, also the ``.in`` files which I think is not intended here. ``/apps/blueman-*.py`` I think should only include the generated files but exclude the ``'in`` files.

infirit:
We should clean the generated files, if this does not happen automatically we will have to add them to Makefile.am CLEANFILES variable. Also we have to make sure the ``.in`` files are included in the tarball so it most likely is necessary to add them to EXTRA_DIST.

cschramm:
Updated.

infirit:
looks good to me.

PR #486, opened 2016-03-16
Use the ListStore to sort instead of rebuilding

infirit:
Ok, this is WIP but it works. Opening a PR so people can comment, criticize and suggest things.

This makes use of the sorting capabilities of the ListStore so we do not rebuild the list every-time. I will also change how the list behaves when multiple adapters are available. As with the sorting, we should not be rebuilding the list but instead put all devices in the list and use a [filter function](http://python-gtk-3-tutorial.readthedocs.org/en/latest/treeview.html#filtering) to modify the view.

The only "drawback" currently is that it does not animate the row as it does not rebuild, it still animates on new/removed device though. However we gain the ability to apply any sorting algoritm we like on any field in the list.

The old sort menu items are still there but here is a screenshot of how it looks. Any feedback here on the ui would be nice :smile: 
manager_sort_menu

### DIFF
--- a/apps/blueman-manager.in
+++ b/apps/blueman-manager.in
@@ -81,10 +81,6 @@ class Blueman(Gtk.Window):
             self.Config["window-properties"] = [w, h, x, y]
             Gtk.main_quit()
 
-        def on_latest_last_changed(ignored1, ignored2):
-            try: self.List.DisplayKnownDevices(autoselect=True)
-            except: pass
-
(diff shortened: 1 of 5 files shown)


cschramm:
I like it. :smiley:

infirit:
Ok, I am happy with the state of this now. A test would be nice :smile:.

cschramm:
Works fine for me. :smiley:

PR #487, opened 2016-03-21
Add dialog with remote services

cschramm:
Not sure where this is going and how it is most useful. I just hacked together a little plugin for the device menu to show UUIDs and service names a given device provides as I always have the feeling that "leave blueman and check the UUIDs with bluetoothctl" is an advice we should not need to give.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
-## master
+## master (unreleased)
+
+### New features
+
+* Information dialog on device's services
 
 ### Changes
 
--- a/blueman/plugins/manager/Info.py
+++ b/blueman/plugins/manager/Info.py
@@ -1,0 +1,76 @@
+import dbus
+from gi.repository import Gtk
+
+from blueman.Functions import create_menuitem, get_icon
+from blueman.Sdp import uuid128_to_uuid16, uuid16_to_name
+
+from blueman.plugins.ManagerPlugin import ManagerPlugin
+
+
+def show_info(device, parent):
+    dialog = Gtk.Dialog()
+    dialog.set_transient_for(parent)
+    dialog.set_icon_name('blueman')
+    dialog.set_title('blueman')
+    store = Gtk.ListStore(str, str)
+    view = Gtk.TreeView()
+    view.set_headers_visible(False)
+    view.set_model(store)
+    for i in range(2):
+        column = Gtk.TreeViewColumn()
+        cell = Gtk.CellRendererText()
+        column.pack_start(cell, True)
+        column.add_attribute(cell, 'text', i)
+        view.append_column(column)
+    dialog.get_content_area().pack_start(view, True, False, 0)
+    view.show_all()
+
+    def format_boolean(x):
+        return _('yes') if x else _('no')
+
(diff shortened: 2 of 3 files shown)


infirit:
In general good to have something that shows these. Maybe add some more properties and call the menu properties?

cschramm:
> Maybe add some more properties

I thought about that but it looks like all other (interesting) properties are actually already visualized in our device list.

> and call the menu properties?

What do you mean?

infirit:
My thought was to call the menu entry (sorry forgot entry last time) "properties" and just expose all of them regardless if they are interesting (now?). Similar to how file managers have an entry for properties :smile:.

cschramm:
Got you now.

I just collected all things we could possibly show:

```
device.Address
device.Name  # maybe
device.Alias
"0x{:06x}".format(device.Class)  # maybe
"0x{:04x}".format(device.Appearance)  # maybe
device.Icon  # maybe
_("yes") if device.Paired else _("no")
_("yes") if device.Trusted else _("no")
_("yes") if device.Blocked else _("no")
_("yes") if device.LegacyPairing else _("no")
if device.RSSI in [0x99, 0x7f]:  # maybe
    'invalid (0x{:02x})'.format(device.RSSI)
else
    '{} dBm (0x{:02x})'.format(device.RSSI, device.RSSI)
_("yes") if device.Connected else _("no")
for uuid in device.UUIDs:
    uuid + ' ' + uuid16_to_name(uuid128_to_uuid16(uuid))
device.Modalias  # maybe
device.Adapter
```

The ones marked with `maybe` could be missing if I get the `GDBusPropertyTable device_properties` array from the BlueZ source right (an additional `exists` method is defined for those properties, but I did not look up how it's used). Note that this would confirm that our `get_name` method is in fact obsolete with BlueZ 5 as `Alias` always exists while it's actually `Name` that could be absent.

infirit:
Yes, all of those :)

>get_name method is in fact obsolete

Ok, good to see it confirmed.

cschramm:
Updated the dialog, but there must be a bug in my GTK logic. When I close the the dialog immediately blueman-manager exits with 139.

infirit:
Works for me. You could try to connect to the response signal and have it handle destroy. This is functionally the same so I doubt it makes any difference.

It is best to pass in the parent window to avoid messages like ```Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.```.

cschramm:
OK, looks like the crash happened in conjunction with `GTK_DEBUG=interactive`, so it's not really relevant.

> It is best to pass in the parent window

I agree, but I don't see a good way to get the parent window at the point where I'm creating the dialog... :confused:

infirit:
Try, ```manager_menu.get_toplevel()``` and then pass that it returns to the ``show_info``.

cschramm:
So glad somebody here has an idea of GTK. Hint: it's not me. :wink: 

Added the parent.

infirit:
>has an idea of GTK

I know enough to be dangerous :smile_cat:.

>Added the parent.

Does what it does on the box. Perhaps at some point, later, add a way to copy the clipboard :smile:

PR #497, opened 2016-04-17
2 0 stable

jpd0735:
my program don't even look for devices guess i will try and reinstall it


cschramm:
I cannot make any sense of this PR.

Issue #443, opened 2016-01-29
blueman error


nimmuk09:
Hi,

I have an erro when starting blueman

> Resource Not Ready
> Traceback (most recent call last):
>   File "/usr/lib/python3.5/site-packages/blueman/bluez/errors.py", line 146, in warp
>     return func(*args, **kwargs)
>   File "/usr/lib/python3.5/site-packages/blueman/bluez/Adapter.py", line 134, in start_discovery
>     self.get_interface().StartDiscovery()
>   File "/usr/lib/python3.5/site-packages/dbus/proxies.py", line 145, in __call__
>     **keywords)
>   File "/usr/lib/python3.5/site-packages/dbus/connection.py", line 651, in call_blocking
>     message, timeout)
> dbus.exceptions.DBusException: org.bluez.Error.NotReady: Resource Not Ready
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/bin/blueman-manager", line 210, in inquiry
>     self.List.DiscoverDevices()
>   File "/usr/lib/python3.5/site-packages/blueman/gui/DeviceList.py", line 399, in DiscoverDevices
>     self.Adapter.start_discovery()
>   File "/usr/lib/python3.5/site-packages/blueman/bluez/errors.py", line 148, in warp
>     raise parse_dbus_error(exception)
> blueman.bluez.errors.DBusNotReadyError: Resource No

bluetooth is on

> [stoop@TiFS ~]$ bluetooth
> bluetooth = on

[stoop@TiFS ~]$ sudo systemctl status bluetooth

> [sudo] password for stoop: 
> ● bluetooth.service - Bluetooth service
>    Loaded: loaded (/usr/lib/systemd/system/bluetooth.service; enabled; vendor preset: disabled)
>    Active: active (running) since Fri 2016-01-29 08:00:04 PHT; 35min ago
>      Docs: man:bluetoothd(8)
>  Main PID: 754 (bluetoothd)
>    Status: "Running"
>     Tasks: 1 (limit: 512)
>    CGroup: /system.slice/bluetooth.service
>            └─754 /usr/lib/bluetooth/bluetoothd
> 
> Jan 29 08:04:04 TiFS bluetoothd[754]: Failed to set mode: Failed (0x03)
> Jan 29 08:09:08 TiFS bluetoothd[754]: Failed to set mode: Failed (0x03)
[...]

PR #498, opened 2016-04-18
Check if bluetooth activation is successful

cschramm:
As mentioned in #443 we previously never checked if the activation (`SetBluetoothStatus(True)`) is actually successful or if it is still disabled afterward. This is expected if a hardware killswitch is switched off.


infirit:
I don't have a hw killswitch so can't really test this but it looks sane so if it works for you :+1:

Issue #423, opened 2016-01-06
Blueman inclusion in openSUSE


posophe:
Hi !

I'm working on openSUSE packaging actually, to provide it as default bluetooth manager in Mate. The Security just gave me a negative answer (I quote) :
"Still the same issues with blueman. Every DBUS function running as root,
allowed to be called by anyone without any authorization.

CVE-2015-8612 as an recent example.

Cant be whitelisted."

Can anyone tell me if it is possibly fixable on upstream ?

Regards.
Benjamin

cschramm:
No it's not. blueman does some things that require root privileges (network, PPP, rfcomm, rfkill). Those tasks are done by blueman-mechanism which provides an API to anybody on the system as obviously anybody could potentially use blueman. If that's not the case on openSUSE and there is only one potential user, you could install respective dbus rules, however that does not make much sense both from a functional and a security viewpoint.

infirit:
Do you know why policykit is not used for this? Policykit is designed to do privileged operations like enabling port forwarding.

cschramm:
It actually is. The DHCP part of blueman-mechanism checks the policykit authentication. The policy file also includes two other actions, but it looks like they are not used anymore. We could add the check to all API methods. However the only improvements I see would be that inactive sessions would not be allowed to use the methods by default and distributions could easily overwrite the rules. I don't think that's what openSUSE security "asks" for as it is not such a big improvement.

infirit:
Sorry, I meant why do we not use polkit as the normal user to do the privileged actions. Or put it another way, why do we not run mechanism directly as the user (not as root) and if needed it will prompt for credentials.

cschramm:
Ah, I see. It would be pretty annoying to have to do the authorization on every start / action / connection, wouldn't it? 😐

cschramm:
Ah, I see. It would be pretty annoying to have to do the authorization on every start / action / connection, wouldn't it? 😐

infirit:
Well, yes and no :smile:. There are options for distro's to create rules that allow this to be done without password for user in group like admin or wheel. I use the below for MATE. I think this a large part of the Suse folks object to

````
/* Allow users in wheel group to change various MATE settings requiring root priviliges */
polkit.addRule(function(action, subject) {
    if (action.id == "org.mate.settingsdaemon.datetimemechanism.settimezone" ||
        action.id == "org.mate.settingsdaemon.datetimemechanism.settime" ||
[...]

PR #506, opened 2016-04-24
Mechanism: dbus polkit authorisation checking

infirit:
@posophe this is for #423

For all, I don't have a machine with a hardware killswitch (only have usb dongles) so I really would appreciate if people can make sure this works properly.

To not have your users to authenticate all the time distros typically ship with a polkit rules file. See below for an example.

```
/* Allow users in wheel group to run blueman network config without authentication */
polkit.addRule(function(action, subject) {
    if ((action.id == "org.blueman.network.setup" ||
         action.id == "org.blueman.dhcp.client" ||
         action.id == "org.blueman.rfkill.setstate" ||
         action.id == "org.blueman.pppd.pppconnect") &&
        subject.isInGroup("wheel")) {
        return polkit.Result.YES;
    }
});
```

### DIFF
--- a/blueman/plugins/mechanism/Network.py
+++ b/blueman/plugins/mechanism/Network.py
@@ -45,6 +45,7 @@ class Network(MechanismPlugin):
     @dbus.service.method('org.blueman.Mechanism', in_signature="ayays", out_signature="", sender_keyword="caller",
                          byte_arrays=True)
     def EnableNetwork(self, ip_address, netmask, dhcp_handler, caller):
+        self.confirm_authorization(caller, "org.blueman.network.setup")
         nc = NetConf.get_default()
         nc.set_ipv4(ip_address, netmask)
         nc.set_dhcp_handler(DHCPDHANDLERS[dhcp_handler])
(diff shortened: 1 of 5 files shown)


infirit:
Question, do we really need polkit to require admin/root authentication (auth_admin) or should we instead request the session user (auth_self)? To be clear, the process is still executed as root AFAIK but the authorisation requested changes from root to the user.

posophe:
No idea. Anyhow, thanks for your improvements. I hope to see blueman available in openSUSE ASAP ;-)

cschramm:
Works fine for me.

infirit:
Great. Now last question, should we commit an example policy rule in the git repo or on the wiki?

cschramm:
Makes sense, yes. Maybe add a polkit section to https://github.com/blueman-project/blueman/wiki/Integrations?

infirit:
Good, added a new polkit wiki page with the example rule and linked to it from the Integrations.

@posophe this should end up in blueman 2.1 which has no release date yet. The changes are not too invasive so if you want I can provide a patch against 2.0 so you can at least try it out.

posophe:
Yeah that would be very helpful :)

posophe:
Many thx, BTW

infirit:
@posophe, branch based on 2.0 with these commits https://github.com/infirit/blueman/tree/polkitauth-2-0, see diff https://gist.github.com/infirit/edb76e72be50164ebd3367982efa9a5f

Create with git command:
```
git diff 660199c5cd7e8369754f250e6c9d765900cf8588
```

posophe:
You're the best :). I do some tests right now.

PR #507, opened 2016-04-28
Disable Gtk overlay scrolling

infirit:
With many (all?) themes this overlaps the content of the ScrolledWindow, horrible.


cschramm:
:+1:

PR #509, opened 2016-04-29
need sudo to install

dlech:


cschramm:
Hm, arguable. It neither makes sense if you're root or have the necessary access rights to the configured prefix nor if you don't have the sudo application. Assuming both is a bad habit from the Ubuntu ecosystem.

infirit:
There are plenty of distributions that do not even install ``sudo`` so I would prefer to mention that ``make install`` _may_ require root privileges.

dlech:
I would also argue that running `./configure` and `make` as root is a bad idea. 

How about splitting it up like this:

      To install a packaged release of blueman, run `./configure && make`, then as root, `make install`.

PR #515, opened 2016-05-02
NetConf: Treat UnicodeDecodeError as if there was no file

infirit:
This is unfortunate but most people will not even notice we created a new
state file.

We really should just use a plain text format like json or simple ini file. There are only a couple of variables we need to store, ip address, pid and DhcpHandler the rest, afaik, are constructed on the fly.


cschramm:
:heavy_check_mark:

Issue #519, opened 2016-05-10
Can Pair, but not connect/select Audio device in PAvuControl


lyubod:
blueman: 2.0.4-1 (from distro package)
BlueZ: 5.36-1 (ditto)
Distribution: OS: Debian Stretch (Testing) with all current updates
Desktop environment: LXDE

Please refer to the following github gists for additional logging information:

https://gist.github.com/lyubod/bec8f1da0da2bc02e730b383ba0aa4fe

https://gist.github.com/lyubod/2be136dc66fbd6824f47099e20d2250c

https://gist.github.com/lyubod/cd5209da9b4c869d153eea3daa244708

https://gist.github.com/lyubod/b651f08dab4ab7e314f06007058dda6d

Have followed guide at:
https://wiki.debian.org/BluetoothUser/a2dp

all required packages installed

Have no directory path /var/lib/gdm3/.config/pulse/client.conf in fact only /var/lib nothing about GDM3 or anything after that

Already added load-module module-switch-on-connect to /etc/pulse/default.pa and rebooted

No improvement, connecting still fails in both GUI and CLI, and no additional audio device shows in PAvuControl

cschramm:
Multiple issues in the logs:

blueman fails to show a notification -> What provides notification-daemon on your system? (`aptitude search '~Pnotification-daemon ~i'`)

module-bluetooth-discover cannot be loaded -> That's expected if it is already loaded. Check with `pactl list modules`. However...

bluetoothctl fails to connect to the device -> You need to load module-bluez5-discover for that to work.

lyubod:
Thanks for your help.

1. aptitude search '~Pnotification-daemon ~i' shows nothing but another prompt, after some other text indicating it's searching. I thought even the really basic GUIs like LXDE and XFCE had some notification, obviously not. So, I'd need to install that, but I've no idea which one, I'll look into it, but appreciate suggestions.

2. pactl list modules shows module-bluez5-discover, so it seems that already loads.

cschramm:
`aptitude search '~Pnotification-daemon'` lists all available packages that provide `notification-daemon`. If you really have none installed, your setup is broken as the `blueman` package depends on it. Xfce has `xfce4-notifyd`. I think there's actually no LXDE specific one.

Regarding your BlueZ problem, you can stop and start bluetoothd with the -d switch to get debugging output in the syslog (/var/log/daemon.log).

lyubod:
[...]

PR #520, opened 2016-05-13
Notification: Fix fallback when no daemon is available and no actions

infirit:
For #519

### DIFF
--- a/blueman/gui/Notification.py
+++ b/blueman/gui/Notification.py
@@ -167,10 +167,18 @@ class Notification(object):
     def actions_supported():
         return "actions" in Notify.get_server_caps()
 
+    @staticmethod
+    def body_supported():
+        return "body" in Notify.get_server_caps()
+
     def __new__(cls, summary, message, timeout=-1, actions=None, actions_cb=None, pixbuf=None, status_icon=None):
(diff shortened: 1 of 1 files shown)


blueman/gui/Notification.py
```diff
@@ -167,10 +167,18 @@ class Notification(object):
     def actions_supported():
         return "actions" in Notify.get_server_caps()
 
+    @staticmethod
+    def daemon_available():
+        return Notify.get_server_caps() != []
```
cschramm: Maybe we should check `'body' in Notify.get_server_caps()`. That's the minimal capability we need for useful notifications and I'm not sure if it would always return `[]` if the daemon is not available (could possibly `None` or something...).

infirit: The api is supposed to return a list which may be empty, see [docs](https://developer.gnome.org/libnotify/0.7/libnotify-notify.html#notify-get-server-caps). But I agree checking for ``body`` is more clean. Updated the commit, thx.

cschramm:
:+1:

infirit:
Also put this in 2-0-stable?

cschramm:
Why not?

infirit:
No reason, just not sure. It's there now :smile:

PR #526, opened 2016-05-24
[Serial] Open rfcomm device as unprivileged user

### LINKED ISSUE #524: Unprivileged rfcomm
Using blueman's serial service I just realized a fundamental usability issue it shares with `rfcomm connect`:

To create the rfcomm device root privileges are required, so `rfcomm connect` needs to be run as root and blueman uses its mechanism. The issue with running `rfcomm connect` is that it then opens the device - also as root of course as that's the user who's running `rfcomm`. The problem is that the user will then only be able to access the device as root too, even if he would have sufficient file system privileges. Otherwise he gets EBUSY.

In the same manner blueman runs its rfcomm-watcher as root as well and has the same issue then.

I just tried and replaced the `open_rfcomm` and `close_rfcomm` calls in `blueman.services.meta.SerialService` with the respective code from `blueman.plugins.mechanism.Rfcomm` and it works perfectly fine for me. blueman-mechanism creates the rfcomm device with `crw-rw---- 1 root dialout` and as I'm in the dialout group the unprivileged group blueman-rfcomm-watcher is able to open it.

The problem with this are the words "for me". If I would not be able to access the file after creation, i.e. especially if I would not be in the dialout group, blueman-rfcomm-watcher would fail. So this is not a generic solution.

I'm not sure how to best handle this. Maybe (applet's perspective):

```
ask mechanism to create the device
if I have read access:
  start rfcomm-watcher
else:
  ask mechanism to start rfcomm-watcher
```

cschramm:
...if he has read and write access.

This requires the watcher to wait if the device is busy which means it's opened by root (or any other user) and typically happens if ModemManager is present and probes the new device. The same method is used for PPPSupport now which previously always waited for five seconds unconditionally. While that is unnessary if ModemManager is not present, it seems to be too short if it is.

Fixes #524.

I'm not 100 % sure about the 0.1 sec delay before checking the access permissions, but if I do it immediately after creating the device, R_OK and W_OK result in False (while E_OK works, so it's not that the file would be absent).

### DIFF
--- a/blueman/services/meta/SerialService.py
+++ b/blueman/services/meta/SerialService.py
@@ -4,10 +4,16 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import logging
+import os
+import subprocess
+from time import sleep
+
 from blueman.bluez.Adapter import Adapter
 from _blueman import rfcomm_list, release_rfcomm_device, create_rfcomm_device
 from blueman.Service import Service
 from blueman.main.Mechanism import Mechanism
+from blueman.Constants import RFCOMM_WATCHER_PATH
 
 
 class SerialService(Service):
@@ -27,9 +33,17 @@ class SerialService(Service):
         try:
             # TODO: Channel?
             port_id = create_rfcomm_device(Adapter(self.device["Adapter"])['Address'], self.device["Address"], 1)
-            Mechanism().open_rfcomm(str('(d)'), port_id)
+            filename = '/dev/rfcomm%d' % port_id
+            # We have to delay the access check to make sure the rfcomm system set up the permissions
+            sleep(0.1)
+            if os.access(filename, os.R_OK | os.W_OK):
+                logging.info('Starting rfcomm watcher as current user')
+                subprocess.Popen([RFCOMM_WATCHER_PATH, filename])
+            else:
+                logging.info('Starting rfcomm watcher as root')
+                Mechanism().open_rfcomm(str('(d)'), port_id)
             if reply_handler:
-                reply_handler('/dev/rfcomm%d' % port_id)
+                reply_handler(filename)
         except Exception as e:
             if error_handler:
                 error_handler(e)
--- a/blueman/services/meta/SerialService.py
+++ b/blueman/services/meta/SerialService.py
@@ -6,8 +6,10 @@ from __future__ import unicode_literals
 
 import logging
 import os
+import signal
 import subprocess
-from time import sleep
+
+from gi.repository import Gio
 
 from blueman.bluez.Adapter import Adapter
 from _blueman import rfcomm_list, release_rfcomm_device, create_rfcomm_device
@@ -19,6 +21,7 @@ from blueman.Constants import RFCOMM_WATCHER_PATH
 class SerialService(Service):
     def __init__(self, device, uuid):
         super(SerialService, self).__init__(device, uuid)
+        self.file_changed_handler = None
 
     def serial_port_id(self, channel):
         for dev in rfcomm_list():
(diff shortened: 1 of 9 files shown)


infirit:
>I'm not 100 % sure about the 0.1 sec delay before checking the access permissions, but if I do it immediately after creating the device, R_OK and W_OK result in False (while E_OK works, so it's not that the file would be absent).

To avoid waiting perhaps use a GFile with a GFileMonitor.
 1. Create the GFile before creating the device
 2. Create the GFileMonitor and connect to the changed signal. 
 3. Create the device
 4. Wait for the permission to be applied.
 5. Do stuff
 6. Clean-up?

cschramm:
Step four would block forever if the user does not get the permissions which is a valid case where we would want to use mechanism.

infirit:
I am thinking more in the below direction. Then store whether the user can read/write/whatever and update this as attributes change.

```python
from gi.repository import Gio, GLib

can_read_attr = Gio.FILE_ATTRIBUTE_ACCESS_CAN_READ

def on_device_changed(mon, f, other_f, event):
    if event == Gio.FileMonitorEvent.ATTRIBUTE_CHANGED:
        print("Attribute changed")
        info = f.query_info(can_read_attr, Gio.FileQueryInfoFlags.NONE)
        can_read = info.get_attribute_boolean(can_read_attr)
        print(can_read)

dev = Gio.File.new_for_path('/dev/kmem')
monitor = dev.monitor(Gio.FileMonitorFlags.NONE)
monitor.connect('changed', on_device_changed)

loop = GLib.MainLoop()
loop.run()
```

cschramm:
It's a udev rule that changes the group on my system:
`KERNEL=="tty[A-Z]*[0-9]|pppox[0-9]*|ircomm[0-9]*|noz[0-9]*|rfcomm[0-9]*", GROUP="dialout"`

We cannot wait for that to happen as such a rule might not be present.

Two schemes I could think of:

1. wait for the change, but apply a timeout as it might not happen
2. if the user gains access, start a user watcher, otherwise start a root watcher

This is somewhat close to my solution with a timeout of 0.1 s. :wink: 

1. start a root watcher
2. wait for the change
3. if the user gains access, kill the root watcher and start a user watcher

That would be better in cases where udev is not involved or the user does not get access (he's not in the dialout group for my system) as we do not wait for the timeout, but it's a little messy otherwise.

cschramm:
I added an implementation of the second scheme. Not really tested yet.

blueman/services/meta/SerialService.py
```diff
@@ -23,14 +32,31 @@ def serial_port_id(self, channel):
     def connected(self):
         return False
 
+    def on_file_changed(self, monitor, file, other_file, event_type, root_watcher_pid):
+        if event_type == Gio.FileMonitorEvent.DELETED:
+            dprint('%s got deleted' % file.get_path())
+        elif event_type == Gio.FileMonitorEvent.ATTRIBUTE_CHANGED and os.access(file.get_path(), os.R_OK | os.W_OK):
```
infirit: Had a quick look and you could consider querying the attributes on the gfile instead of os module. I don't like the Cish api though so won't blame you to keep it as it is :smile: 

```
info = gfile.query_info('*', Gio.FileQueryInfoFlags.NONE) # * gets all attributes
can_write = info.get_attribute_boolean(Gio.FILE_ATTRIBUTE_ACCESS_CAN_WRITE)
```

infirit:
I don't really have any real serial devices to test so as long as it works ok for you 👍

cschramm:
Just half a year later :see_no_evil: I finally rebased on master, played with this a little bit, and fixed the second commit. Two issues there: Obviously the user cannot kill the root watcher himself but has to use mechanism for that. Second, when trying this, the change did not take place for me, probably because permissions are already set when the monitor gets set up.

As a solution, I now check permissions once right after that's done. There might be a race condition where both the explicit check *and* a change event take place but I guess that's not too bad. The only issue I could think of is that we could start two user watchers for the same port.

Overall the current state works fine for me. I can connect to the serial service and get a root watcher that immediately gets replaced by a user watcher and can then connect to my test device (a Naze32) as if it would be connected via USB.

There's only one inconvenience: I have to disable ModemManager. If it probes the device no connection is possible afterward. I guess the commands it sends bring the Naze32 into some strange state so that the software cannot address it. That does not happen with USB but I guess that's just ModemManager checking on rfcomm devices but not ttyUSB, so there's little to nothing blueman could do about that.

infirit:
If it works for you 👍. But we then have to deal with grep in PPPSupport, https://github.com/blueman-project/blueman/pull/601#issuecomment-255648251. Shall I do the same as I did in [2.0 stable](https://github.com/blueman-project/blueman/commit/aa2ee8abf39bec159140f626c7656b43ea7e1566) for master?

cschramm:
Makes sense, yes. :+1:

PR #528, opened 2016-05-29
Drop old NetworkManager and ModemManager 0.x legacy code

infirit:
So one question, we now rely completely on bluez/networkmanager to handle the connection. This is all ok except the user may remove the connection themselves and we cannot activate the connection. The  question then is, should we still create the connection? I am leaning towards we still create the connection even if it is missing.

This works for me with a PAN phone but I don't have a phone that supports DUN. Would be nice if someone can test this with DUN :smile:

### DIFF
--- a/blueman/plugins/applet/NMDUNSupport.py
+++ b/blueman/plugins/applet/NMDUNSupport.py
@@ -22,9 +22,6 @@ class ConnectionHandler:
         self.rfcomm_dev = None
         self.timeout = None
 
-        # ModemManager 0.x
-        self.parent.bus.add_signal_receiver(self.on_mm_device_added, "DeviceAdded", "org.freedesktop.ModemManager")
-        # ModemManager 1.x
         self.parent.bus.add_signal_receiver(self.on_interfaces_added, "InterfacesAdded",
                                             "org.freedesktop.DBus.ObjectManager")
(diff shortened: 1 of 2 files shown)


cschramm:
> This is all ok except the user may remove the connection themselves

What exactly do you mean? A script that catches new connections and removes them? Or do you refer to some NM setting?

cschramm:
Ah, I think I got it: The connection gets created ones the device becomes available, but the user could choose to activate it at any later point in time, so he could have removed the connection in between. Well, as he removed it on purpose, I guess the ideal solution would be not to offer connecting to the service anymore then, however, that would require quite some changes. Keeping the fallback is definitely the easier solution, yep...

infirit:
Yes indeed. Updated the PR and added a couple more commits. Do you have a DUN phone to test this with?

cschramm:
I don't think so, no. I could probably emulate something with an HC-05, but that's not worth the effort.

infirit:
Agreed simulating something is not worth the effort for two lines. I doubt it will cause problems but would have been nice. Merged, now wait for the fallout :smile:.

cschramm:
Needs a changelog entry then at some point.

infirit:
Done.

PR #543, opened 2016-07-03
Port AppletService to GDBus

infirit:
Also added a couple small changes that I noticed while working on blueman-manager for this port.


cschramm:
Not sure why you use that `_init` instead of `__init__` but looks good overall.

infirit:
The reason is that even-though in ``__new__`` we return an existing instance we still end up calling ``__init__``  every time on that instance. So instead we use ``_init`` and call it explicitly in ``__new__`` so we only ever initialize the instance once.

I did the same for the bluez client classes.

PR #544, opened 2016-07-04
Re-implementation of NM Pan support with GDBus proxy objects

infirit:
Direct translation to GDBus ended up being quite ugly so decided to rewrite using actual GDBusProxy objects connecting to the signals. The goals is to not use the NewConnectionBuilder class and it will be removed.  This is not meant to be merged, just to put what I have done so far out there so people know about it.

### DIFF
--- a/blueman/plugins/applet/NMPANSupport.py
+++ b/blueman/plugins/applet/NMPANSupport.py
@@ -4,219 +4,342 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+from gi.repository import Gio, GLib, GObject
 from blueman.Functions import dprint
-from gi.repository import GLib
-import dbus
 from blueman.plugins.AppletPlugin import AppletPlugin
 from uuid import uuid4
 
+devtypes = {1: "Ethernet",
+            2: "Wi-Fi",
+            5: "Bluetooth",
+            6: "OLPC",
+            7: "WiMAX",
+            8: "Modem",
+            9: "InfiniBand",
+            10: "Bond",
+            11: "VLAN",
+            12: "ADSL",
+            13: "Bridge",
+            14: "Generic",
+            15: "Team",
+            16: "TUN",
+            17: "IPTunnel",
+            18: "MACVLAN",
+            19: "VXLAN",
+            20: "Veth"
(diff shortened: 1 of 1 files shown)


infirit:
Updated, not working yet, but we are getting somewhere. Took me a bit of time to figure NetworkManager out but we got it now.

infirit:
Ok, this is pretty much finished except that there is an annoying bug which I am confused about :confused:.

"Something" appears to initiate the network manager connection outside of the [service_connect_handler](https://github.com/blueman-project/blueman/pull/544/commits/db1f6573853fc30c51213de4e71a1358b05e8464#diff-5b460f3488fed97a2f213612fe0b466dR312) and as a result when I call ``ActivateConnection`` on NetworkManager the proxy returns that the connection is already in progress.

```
fail (/usr/lib64/python2.7/site-packages/blueman/gui/manager/ManagerDeviceMenu.py:126)
fail g-io-error-quark: GDBus.Error:org.freedesktop.DBus.Python.blueman.bluez.errors.DBusInProgressError: DBusInProgressError:  Operation already in progress
```

Commenting out the ``ActivateConnection`` line I do see the connection states progress and either fail or succeed. I am properly confused by this and if anybody has ideas what is going on do reply here :smiley:

cschramm:
Hm... It doesn't look like the old code catches that somehow. I think it would always run `ActivateConnection` for BlueZ 5 unless it runs into the `Network Manager did not support the connection` case. Do you still have your code of the "Direct translation to GDBus" to check if that suffers from the same problem?

infirit:
> I think it would always run ActivateConnection for BlueZ 5 

Yes, trying to disconnect it will result in an error saying the connection was not activated.

>Do you still have your code of the "Direct translation to GDBus" to check if that suffers from the same problem?

No but I'll do that now to see if that does work ok. I was really happy with the NMDevice class though but if I can't get it to work.. :smile:

infirit:
Completely untested but here is the direct translation to GDBus in my [nm_gdbus_direct](https://github.com/infirit/blueman/tree/nm_gdbus_direct) branch.

infirit:
Learned a lot from trying to re-implment this but see  #567

PR #551, opened 2016-07-13
RFC: blueman-applet: Do not create and tear down manager objects

infirit:
* GDBus does not error if we create a proxy when bluez is not running.
 * If bluez is not running we are unlikely to see signals
 * We now only run "on_manager_state_changed" when the name (dis)appears.

------
This does expose a logic bug in BluezAgent resulting in failing to unregister. How are we supposed to unregister when the bluez was stopped? There are also some other smallish changes needed so this is not meant to be merged just yet. I need to go now and revisit this later.

```
__del__ (/usr/lib64/python3.5/site-packages/blueman/main/applet/BluezAgent.py:37)

Exception ignored in: <bound method BluezAgent.__del__ of <blueman.main.applet.BluezAgent.BluezAgent object at 0x7f23c9963588>>
  File "/usr/lib64/python3.5/site-packages/blueman/main/applet/BluezAgent.py", line 39, in __del__
    Bluez.AgentManager().unregister_agent(self.__agent_path)
  File "/usr/lib64/python3.5/site-packages/blueman/bluez/AgentManager.py", line 25, in unregister_agent
    self._call('UnregisterAgent', param)
  File "/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py", line 120, in _call
    raise parse_dbus_error(e)
blueman.bluez.errors.DBusServiceUnknownError:  The name org.bluez was not provided by any .service files
```

### DIFF
--- a/apps/blueman-applet.in
+++ b/apps/blueman-applet.in
@@ -50,6 +50,7 @@ class BluemanApplet(object):
 
         self._signals = []
 
+        self.plugin_run_state_changed = False
         self.Manager = None
         self.DbusSvc = DbusService("org.blueman.Applet", "/")
 
@@ -71,7 +72,10 @@ class BluemanApplet(object):
     def manager_init(self):
         try:
             self.Manager = Bluez.Manager()
-            self.Plugins.Run("on_manager_state_changed", True)
+            self.manager_state = True
+            self.plugin_run_state_changed = True
+        
+            self.Plugins.Run("on_manager_state_changed", self.manager_state)
 
             self._signals.append(self.Manager.connect_signal('adapter-added', self.on_adapter_added))
             self._signals.append(self.Manager.connect_signal('adapter-removed', self.on_adapter_removed))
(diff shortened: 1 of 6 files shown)


cschramm:
> How are we supposed to unregister when the bluez was stopped?

There's no necessity to do it then (as the agents are not persisted if I'm correct). Catching the exception and just `pass`ing should be fine.

infirit:
The only time the Agent knows it has to unregister the **object** on the bus is when bluez calls ``Release`` on it. Bluez calls ``Release`` when the daemon is stopped (at this point the agent is already unregistered) and the name disappears (and we run ``on_manager_state_changed``). Other than that the Agent is not able to figure this out so we should just let the plugin handle things explicitly, this is in the second commit.

Now, there is implicit behaviour to indicate the manager's state when ``applet.Manager`` is ``None`` or not. This imho is not so nice as ``AppletPlugin`` determines to ``run on_manager_state_changed`` in ``_on_load`` based off this and caused me some confusion what was going on. I now made it explicit what the state is and whether ``AppletPlugin`` has to run ``on_manager_state_changed`` on newly loaded plugins. I added this to the first commit but I will most likely split this into it's own commit.

Pff, ok then, time for bed :smile:.

cschramm:
Full ack.

infirit:
Thanks, split and pushed.

Also fixed unloading of Networking plugin while I am at it. At some point it could use some love and add a register/unregister function :smile:.

PR #552, opened 2016-07-13
RFC: Add some comments explaining what the plugin methods should do and do

infirit:
The plugin system is nice but the design is not really documented anywhere. I started documenting by adding docstring to the virtual functions and a few others within the plugin context.

The basic design how I see it is:
 - Any setup/tear-down necessary for the plugin to work should be done in on_load and on_unload. This excludes any interaction directly with bluez like manipulating properties, this should be setup in on_manager_state_changed
 - on_manager_state_changed indicates whether bluez dbus api is available or not. We should add/remove watches, (dis)connect bluez object signals and setup/clean-up anything bluez related.
 - Then we have several the default functions that are run on the plugins which should be used to update state, change the ui etc.

ps: also two commits removing unused import and an unused variable

### DIFF
--- a/blueman/plugins/AppletPlugin.py
+++ b/blueman/plugins/AppletPlugin.py
@@ -4,13 +4,9 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-from gi.repository import GObject
-
 import gi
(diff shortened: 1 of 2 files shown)


cschramm:
No objections of course. :smiley:

Issue #415, opened 2015-12-17
PIN code databse


cschramm:
As built-in PIN codes come up over and over again, I'll implement a PIN code database. I will steal [gnome-bluetooth's](https://github.com/GNOME/gnome-bluetooth/blob/master/lib/pin-code-database.xml) as a starting point.

rworkman:
Assuming I understand this, it seems like something that should probably be added to the udev hwdb and be easily query'able.

cschramm:
Sounds reasonable. bluedevil maintains basically the same file (they do not use the vendor attribute): https://quickgit.kde.org/?p=bluedevil.git&a=blob&f=src%2Fwizard%2Fpin-code-database.xml

rworkman:
Hrm... maybe the hwdata project maintained by RH is a better option: https://git.fedorahosted.org/git/hwdata.git
Are you going to use the same format, or maybe a better question is, is the bluedevil file's format the same as gnome-bluetooth's format? If they're in two different formats, I doubt they would embrace any effort to push this in to hwdata, and thus it's just as well to go ahead and do it here. I just hate the idea of trying to keep it in sync with changes that happen to gnome-bluetooth and bluedevil. Everyone would win by keeping a single copy updated somewhere neutral, maybe even as part of bluez itself.

rworkman:
Nevermind - I see the answers to the format issue by looking at the files :-)
Sooo... how about petitioning bluez upstream to include this information at e.g. /usr/share/bluetooth/pin-code-database.xml ?

infirit:
There is a big downside of having this in a project you do not control. Changes may take forever to trickle down to the distro's and having changes accepted may not be so easy. And that is after all agreed on the format with lot's of discussion between the projects, bikeshedding, format stability etc.. It probably is not worth the effort.

However what we can do is contribute back to gnome if we add new devices.

rworkman:
It looks like the format and even the file itself is pretty much identical, which makes me wonder how often things are added/changed/etcetera, and so maybe some of that's not an issue. Valid points nonetheless.

cschramm:
[...]

Issue #10, opened 2013-10-12
Feature: Compose and send notes


cschramm:
[Debian #549274](http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=549274)

PR #554, opened 2016-07-14
Update CHANGELOG and time for an alpha?

infirit:
Doing a PR so we can have a quick discussion whether it is time to do a first alpha release. So much has changed so it would be good if it got some more exposure.

If not what do you think should be done before that can happen?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 * Fix building with musl libc
 * Add generic device-added/removed functions for plugins
 * Drop legacy NetworkManager and ModemManager support
+* Port AppletService proxy to GDBus
 
 
 ### Bugs fixed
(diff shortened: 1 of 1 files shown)


infirit:
Ah, crap merged it somehow, oh well, we can still have the discussion :smile:

cschramm:
:smile: I agree we can target a prerelease version sometime soon.

I started working on #10 a couple of weeks ago, but did not find the time to continue yet. I really want to get that feature in, just because it was kind of the very first item I scheduled for a 2.1 version, long before 2.0 was ready. :laughing:  Another interesting improvement for 2.1 is #415 I think.

I'll try to finish #10 asap.

cschramm:
Sorry for blocking this so long... Can you get the changelog up to date?

infirit:
Updated the changelog :smile:

PR #557, opened 2016-07-18
Functions: Fix error formatter function e_()

infirit:
>We now split and always get the last item as this holds the error message.
>Also call strip() on the string as it may contain a new line and the
>MessageArea will show a "More" button.
>
>This got broken with the GDBus migration of various parts in blueman as
>GLib's exceptions do not match dbus-glib's.

------

There is, imo, a weird implicit behaviour in the ``MessageArea`` where a string containing a newline will trigger the ``More`` button to show up, see [1]. The newline is added by the ``e_`` function when it got an instance of ``Exception``. I would really like to drop this and explicitly send a tuple instead.

Any objections to this? If not I'll add it to this PR.

[1] https://github.com/blueman-project/blueman/blob/master/blueman/gui/MessageArea.py#L168

### DIFF
--- a/apps/blueman-manager.in
+++ b/apps/blueman-manager.in
@@ -193,7 +193,7 @@ class Blueman(Gtk.Window):
             self.List.DiscoverDevices()
         except Exception as e:
             prog.finalize()
-            MessageArea.show_message(e_(e))
+            MessageArea.show_message(*e_(e))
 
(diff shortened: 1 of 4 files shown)


cschramm:
Fine with that.

infirit:
Updates, will merge in a couple of days if nothing comes up :smile:.

PR #560, opened 2016-07-25
WIP: GDBus applet and mechanism

infirit:
Updated FP.
- Both services start and run properly
- Methods are being added by plugins do not yet show up

The plugin methods are added to the object but are not exported. We likely will need to unexport/export.


infirit:
Update: both services come up but all plugin methods are not showing up yet. Added a test method in main.DbusService will make them show up so I know the services are exported and responding properly :tada:.

infirit:
#617

PR #564, opened 2016-08-02
blueman-mechanism: Port Polkit client code to GDBus

infirit:
Also added a small commit that ``dprint``'s the exception when the proxy creation failed.


cschramm:
Not sure why you dropped #559, but this looks good as well. Compared to #559 it saves a small dependency. Is that your reason?

infirit:
I initially had trouble with GDBus (all my fault btw) and used pygi. Now that I figured out my mistakes I used dbus again. I don't mind which one we use in the end. I'll merge this one in a couple of days unless I hear back with objections :smile:.

Issue #549, opened 2016-07-11
ManagerDeviceMenu: Networkaccess does not update when disconnecing


infirit:
Documenting so we do not forget.

In current master, the menu entry does not update and keeps showing "Connected". It appears when we regenerate the menu the service is returning itself as connected but in fact it is not physically.

From a quick check it appears that bluez.Network return Connected True but further investigating is needed.

edit: 2.0.4 is fine but I also do not see a property changed when disconnecting.

infirit:
Well, this is weird, below is a idle session where I connect and disconnect but I only ever see the property change when connecting. I am not sure we are at fault..

```python
>>> from gi.repository import GLib, Gio
>>> proxy = Gio.DBusProxy.new_for_bus_sync(Gio.BusType.SYSTEM,
            Gio.DBusProxyFlags.NONE, None, 'org.bluez',
            '/org/bluez/hci0/dev_6C_83_36_B9_C6_D6', 'org.bluez.Network1')
>>> def on_g_properties_changed(proxy, changed, invalid):
	print(changed, invalid)

	
>>> proxy.connect('g-properties-changed', on_g_properties_changed)
9
>>> loop = GLib.MainLoop()
>>> loop.run()
{'Connected': <true>, 'Interface': <'bnep0'>, 'UUID': <'00001116-0000-1000-8000-00805f9b34fb'>} []
{'Connected': <true>, 'Interface': <'bnep0'>, 'UUID': <'00001116-0000-1000-8000-00805f9b34fb'>} []
```

cschramm:
I *do* see property changed in every case, i.e. 2.0.4, in master, and with your test code (different object path of course):

```
{'Connected': <true>, 'Interface': <'bnep0'>, 'UUID': <'00001116-0000-1000-8000-00805f9b34fb'>} []
{'Connected': <false>} ['UUID', 'Interface']
```

Did you replace your Gentoo with that strange, broken Fedora thingy? :wink:

infirit:
Haha, wel, I do run unstable version of Gentoo and I do tend to forget I have to restart my session/machine once in a while :smile:  so things may have gotten confused. Thanks for checking you do see the update :+1:

infirit:
I like to keep this one open for now to make sure we verify all is well when it is alpha/beta time.

infirit:
This is not a problem for me anymore, closing.

PR #567, opened 2016-08-04
NMPANSupport: port to gdbus

infirit:
This one works now but please test and let me know if there are any issues.

Some improvements in the future which I'll leave for the future.

 * The device holds both the active connection and available connections so no need to traverse all connections.
 * The device also hold the bluetooth device address in the interface property. We could use this to check against instead of retrieving all properties from the NM bluetooth device.
 * The only reason we need the NM bluetooth device is to check whether it is a PAN or a DUN device

### DIFF
--- a/blueman/plugins/applet/NMPANSupport.py
+++ b/blueman/plugins/applet/NMPANSupport.py
@@ -5,8 +5,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 from blueman.Functions import dprint
-from gi.repository import GLib
-import dbus
+from gi.repository import Gio, GLib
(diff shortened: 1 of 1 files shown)


cschramm:
For some reason this version still shows a disconnect item instead of a connect item after disconnecting from the NAP, but rebasing on master makes it work, so it's probably just an issue of the base commit (maybe 9a4b0e8 fixed it?).

infirit:
>still shows a disconnect item instead of a connect item after disconnecting from the NAP

Yes, I saw this and have an issue open for this (#549) but it also happened before this though. I'll merge it now and update the changelog.

PR #569, opened 2016-08-05
Use DBus for sending notifications

infirit:
This drops the libnotify dep :smile:. Still need lots of testing in the various places, especially actions.


infirit:
Fixed some bugs and added python2 compatibility (str on dbus signatures).

cschramm:
I could not find any issue. Everything works as before. :+1:

PR #572, opened 2016-08-10
Bluez: Subclass from Gio.DBusProxy and properly handle properties

infirit:
This is what I actually wanted from the start but could not make it work back then. Figured it out now :smile: . I'll likely do the same for the object manager client (our manager classes) and add it to this PR or a new one.

Like we did for the ``_call`` method we can also use asynchronous call for the properties. I am not entirely sure if this makes sense though for bluez and blueman so left them all synchronous for now.

~~I'll probably also make ``_call`` asynchronous regardless of reply/error_handlers and have the callback handle both cases.~~ - Added this to the PR

``` 
* Use GDBusProxy ability to call methods on a different interface.
 * Override do_g_properties_changed and emit our signals there so
   no more connecting to signals.
 * Add property handling to Base and drop PropertiesBase.
```


cschramm:
Looks a lot cleaner now. :+1:

PR #575, opened 2016-08-17
Generic connect and disconnect

infirit:
> TODO: Handle connect error by showing an error in the message area

Some feedback needed on this one :smile:.
- Which icon should we use for the generic connect? We used to use the icon from the service but this is not an option anymore. Used blueman for now.
- Are there any services outside of serial and netwok we need to handle?


blueman/plugins/manager/Services.py
```diff
@@ -53,12 +53,8 @@ def add_menu_item(manager_menu, service):
             item.show()
 
         for service in get_services(device):
-            try:
+            if service.group == 'network':
```
cschramm: I think we should drop the Input and Audio service classes as they are not actually used now and make `connect`, `disconnect`, `connected` abstract in `blueman.Service`. No need to check for specific service types here then as `get_services` will only return relevant ones.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -213,6 +224,18 @@ def Generate(self):
             self.append(item)
             return
 
+        # Generic (dis)connect
+        if not row["connected"]:
+            connect_item = create_menuitem(_("_Connect"), get_icon("blueman", 16))
```
cschramm: It would be nice to list the affected services in this menu item as you cannot see what you'll get now anymore except for the info dialog.

blueman/plugins/manager/Services.py
infirit: Makes sense, pushed commits to this branch for this.

blueman/gui/manager/ManagerDeviceMenu.py
infirit: How about a tooltip that lists the typical services it would cover, input an audio I know of but are there more that you know of?

cschramm: Profiles that are connected in the generic connect method have the `auto_connect` attribute set in their `btd_profile` struct. Those are A2DP source, A2DP sink, and HID (It looks like our HandsfreeGateway service did not really make sense) . Another improvement would be to not show the generic connect item in case none of them are available.

cschramm:
We should still drop the implementations for `connected`, `connect`, and `disconnect` in `blueman.Service` as they are not actually used anymore.

infirit:
We run into problems in ``RecentConns`` plugin which does not show the generic connected devices.

infirit:
So yesterday I went a slightly different route with this. Instead of handling everything manually I now used the already existing ``connect_service`` dbus method. For this to work I had to add a generic connect to Sdp with id 0.

I also added a tooltip on the connect menu entry. However I can't find a way in dbus to figure out which profiles are auto connectible, am I missing something?

infirit:
I now also implemented not showing the generic connect based of the 3 hardcoded (for now) uuid16 values. Also LE devices are somewhat problematic as in they do not expose UUIDs unless connected to at least once. I also handle this case and assume when the device UUIDs returns an empty list we should provide generic connect.

That leaves me with only two open items. Number one being the tooltip text is awful and can use better wording. And if possible a programmatic way to figure out which uuids can be auto connected to,

And of course the overall approach being acceptable or now 😺

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -209,6 +242,29 @@ def Generate(self):
             self.append(item)
             return
 
+        # Generic (dis)connect
+        # LE devices do not appear to expose certain properties like uuids until connect to at least once.
+        # show generic connect for these as we will not show any way to connect otherwise.
+        device_uuids = self.SelectedDevice['UUIDs']
+        show_generic_connect = False if device_uuids else True
```
cschramm: `not device_uuids` ;)

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -209,6 +242,29 @@ def Generate(self):
...
+        for uuid in device_uuids:
+            if uuid128_to_uuid16(uuid) in (0x110a, 0x110b, 0x1124):
```
cschramm: Why not use constants from `blueman.Sdp`?

blueman/plugins/manager/PulseAudioProfile.py
```diff
@@ -115,8 +110,13 @@ def generate_menu(self, device):
             self.item.show()
 
     def on_request_menu_items(self, manager_menu, device):
+        audio_source = False
+        for uuid in device['UUIDs']:
+            if uuid128_to_uuid16(uuid) == AUDIO_SOURCE_SVCLASS_ID:
+                audio_source = True
+
+        if device['Connected'] and audio_source:
```
cschramm: What about sinks?

blueman/services/__init__.py
```diff
@@ -4,12 +4,8 @@
 from __future__ import division
 from __future__ import unicode_literals
 
-from blueman.services.AudioSink import AudioSink
-from blueman.services.AudioSource import AudioSource
 from blueman.services.DialupNetwork import DialupNetwork
 from blueman.services.GroupNetwork import GroupNetwork
-from blueman.services.HandsfreeGateway import HandsfreeGateway
```
cschramm: You're dropping HandsfreeGateway but not considering it for the generic connect. See my comment on the `auto_connect` properties in BlueZ.

cschramm:
Yes, I don't think that's exposed in D-Bus. The C struct btd_profile has an auto_connect property which is set for HID_UUID, A2DP_SOURCE_UUID, and A2DP_SINK_UUID. struct default_settings also has an equal property set for HFP_AG_UUID, HFP_HS_UUID, and HSP_HS_UUID. I'm not sure if both fields have the same effect.

blueman/gui/manager/ManagerDeviceMenu.py
infirit: Actually no :). The logic is that if we have no UUIDs we set this to True so to show generic connect. If not we set it to False and iterate over the UUIDs below this snippet. I can reverse the logic if this makes things clearer 
```python
show_generic_connect = True if not device_uuids else False
```

blueman/gui/manager/ManagerDeviceMenu.py
infirit: Will update.

blueman/services/__init__.py
infirit: I'll add all of the ones you listed including this one.

blueman/plugins/manager/PulseAudioProfile.py
infirit: Will add AUDIO_SINK_SVCLASS_ID

infirit:
Updated per comments and dropping WIP from the subject.

blueman/gui/manager/ManagerDeviceMenu.py
cschramm: `True if not device_uuids else False` is just `not device_uuids` or am I missing something?

infirit: Haha, you meant it like that, completely misunderstood 😸. I'll fix that 👍

infirit:
Rebased against master.

cschramm:
I guess we should get this into master or is there any actual blocker?

infirit:
No blocker so far I can see, added commit 303f9a4617f2632a577155d64ebafcd25023e8fb for the LE devices uuid. Please check if I understood you properly on https://github.com/blueman-project/blueman/issues/648#issuecomment-276891136. I'll leave the separate commit so you can review easier but I'll squash it together before pushing.

cschramm:
Yes, that's what I meant there.

infirit:
Squashed and rebased against master. If nothing else comes up I think this is good to go.

infirit:
Totally forgot about this 😢. Rebased one more time and will merge in a day or so unless objections raised 😄

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -213,6 +250,34 @@ def Generate(self):
...
+        # LE devices do not appear to expose certain properties like uuids until connect to at least once.
+        # show generic connect for these as we will not show any way to connect otherwise.
+        device_uuids = self.SelectedDevice['UUIDs']
+        show_generic_connect = not device_uuids
+        for uuid in device_uuids:
+            if uuid128_to_uuid16(uuid) in (AUDIO_SOURCE_SVCLASS_ID, AUDIO_SINK_SVCLASS_ID,HANDSFREE_AGW_SVCLASS_ID,
+                                           HEADSET_SVCLASS_ID, HID_SVCLASS_ID):
+                show_generic_connect = True
+                break
+            elif int(uuid.split("-")[0][4:6], 16) == 0x18:
```
cschramm: While this matches all currently assigned GATT service UUIDs there does not seem to be such a restriction in general.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -213,6 +250,34 @@ def Generate(self):
...
+                # LE devices
+                show_generic_connect = True
```
cschramm: Hm. This does not work. Testing with a paired Logitech MX Master the Connect D-Bus method just results in `org.bluez.Error.Failed` after a while. The same happens when I try to connect in `bluetoothctl`. I guess an LE device is just never "Connected" in terms of the BlueZ D-Bus API. We should be fine If we drop this `elif` branch as we never show a "Connect" or "Disconnect" item for them then.

cschramm (changes requested):

blueman/gui/manager/ManagerDeviceMenu.py
infirit: When in ``bluetoothctl`` do you see the connected state change for you mouse? I can drop it just worried that we then also have no way to disconnect.

I really need to get me a BT LE mouse but most are pretty expensive 😿

cschramm: Forget what I wrote. I think the mouse was just sleeping when I tried to connect. 😆 `Connect` works when I make sure it's awake and changes the `Connected` and `ServicesResolved` properties.

blueman/gui/manager/ManagerDeviceMenu.py
cschramm: After a deep dive in the BlueZ code and the Bluetooth Core Specification I have the following picture:

There seems to be no way to detect an LE device via the BlueZ API.
A GATT server with a GAP service is mandatory for LE devices.

We should check for the GAP service with our `GENERIC_ACCESS_SVCLASS_ID`.

I'm not sure what would happen for non-LE devices with a GAP service and non of the auto-connectable audio and input services but I don't expect such a device to exist and if it does it would be confusing anyway, so a connect option that does not have an effect would fit. 😄

infirit: So, add ``GENERIC_ACCESS_SVCLASS_ID`` to the list of class ids ``AUDIO_SOURCE_SVCLASS_ID, AUDIO_SINK_SVCLASS_ID,HANDSFREE_AGW_SVCLASS_ID, HEADSET_SVCLASS_ID, HID_SVCLASS_ID`` and drop the elif. Much nicer 👍 

I'll rebase soon and update with this change.

infirit:
Made the last change requested and rebased against master.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -213,6 +252,32 @@ def Generate(self):
...
 
+        # Generic (dis)connect
+        # LE devices do not appear to expose certain properties like uuids until connect to at least once.
+        # show generic connect for these as we will not show any way to connect otherwise.
+        device_uuids = self.SelectedDevice['UUIDs']
+        show_generic_connect = not device_uuids
+        for uuid in device_uuids:
+            if ServiceUUID(uuid).short_uuid in (
+                AUDIO_SOURCE_SVCLASS_ID, AUDIO_SINK_SVCLASS_ID, GENERIC_ACCESS_SVCLASS_ID,
+                HANDSFREE_AGW_SVCLASS_ID, HEADSET_SVCLASS_ID, HID_SVCLASS_ID
```
cschramm: I'm sorry but I have to throw in yet another discovery. 😆

It probably does not make sense to provide this for *every* LE device but only if there are auto-connectable services.

See https://github.com/blueman-project/blueman/pull/575#issuecomment-270059859 on those. I think we're still missing `HANDSFREE_SVCLASS_ID` here for this entry in `struct default_settings`:

```
           {
                .uuid           = HFP_AG_UUID,
                .name           = "Hands-Free Voice gateway",
                .priority       = BTD_PROFILE_PRIORITY_HIGH,
                .remote_uuid    = HFP_HS_UUID,
                .channel        = HFP_AG_DEFAULT_CHANNEL,
                .authorize      = true,
                .auto_connect   = true,
                .get_record     = get_hfp_ag_record,
                .version        = 0x0105,
        }
```
Beside audio sink and source and HID `auto_connect` is now set in `struct btd_profile hog_profile` and `struct btd_profile midi_profile`.

The short UUID for `struct hog_profile` is 0x1812 (`HOG_UUID16` in the BlueZ code). I did not find this is my initial analysis as it used a different auto-connect method before BlueZ 5.43, see [this commit](https://git.kernel.org/pub/scm/bluetooth/bluez.git/commit/profiles/input/hog.c?id=7d9718cfcc11eaa9d8059e721301cdc00ef8c82e). I checked that there are no other profiles that use that anymore.

The `MIDI_UUID` is `03B80E5A-EDE8-4B33-A751-6CE34EC4C700` and as it's no reserved UUID there is no 16-bit version. The profile was added in BlueZ 5.44. I see three possible solutions for this:
- Ignore it completely. This means that somebody with BlueZ >= 5.44 will not see the generic connect entry for a device with a MIDI service as the only auto-connecting service.
- Enable generic connect for it. This means that somebody with BlueZ < 5.44 will see the generic connect entry for a device with a MIDI service but no (other) auto-connecting service but nothing will get connected when he uses it.
- Enable generic connect for it depending on the version of BlueZ.

infirit: >I'm sorry but I have to throw in yet another discovery. 😆

Now I am sure, you are doing this intentionally 😜 

>I think we're still missing HANDSFREE_SVCLASS_ID here

You are correct, missed this one. Fixed locally.

>The short UUID for struct hog_profile is 0x1812 (HOG_UUID16 in the BlueZ code).

Which translates to the name "Human Interface Device" but we do not have it listed otherwise in the Sdp.py. Will add to the list as ``0x1812``.

>The MIDI_UUID is 03B80E5A-EDE8-4B33-A751-6CE34EC4C700 and as it's no reserved UUID there is no 16-bit version. The profile was added in BlueZ 5.44. I see three possible solutions for this:

Fourth option, add a FIXME and wait for someone to report it not working 😸. It would be best if we had a full dump of the properties of one of these devices so we can see.

infirit: Ok, Added a new commit that adds the missing audi gateway, hog and midi services. Will squash with the main commit before merging.

My reasoning for adding midi is that we never provided this service before as we had no knowledge of the proprietary uuid. If the users runs a recent enough bluez version connect will succeed if not no loss compared to what we would have provided before.

The midi service check is imho a bit ugly but I am not sure how else to best handle this.

Awaiting further investigation that adds more work for me 😝.

cschramm: We should drop `GENERIC_ACCESS_SVCLASS_ID`. We're only interested in 0x1812, not GATT in general.

infirit: New commit, will squash that as well with main commit.

cschramm (approved):

infirit:
squashed and merged

PR #576, opened 2016-08-21
Bluez object manager

infirit:
This changes a couple of things to make more sense.

 * Rename list_* functions to get_* as we are returning proper Adapter and Device objects.
 * Move device object handling to the manager, this is the job of the object manager not the adapter.

To then make it possible to subclass from ``Gio.DBusObjectManagerClient``. Tested this quickly and things appear to work well but feedback and more testing appreciated :smile: .

### DIFF
--- a/blueman/bluez/Adapter.py
+++ b/blueman/bluez/Adapter.py
@@ -5,9 +5,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 from gi.repository import GObject, Gio, GLib
-from blueman.Functions import dprint
 from blueman.bluez.Base import Base
-from blueman.bluez.Device import Device
 from blueman.bluez.AnyBase import AnyBase
 
 class Adapter(Base):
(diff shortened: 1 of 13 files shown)


infirit:
Oops, merged it by accident :cry:

blueman/bluez/Adapter.py
```diff
@@ -16,27 +14,6 @@ class Adapter(Base):
     def _init(self, obj_path=None):
         super(Adapter, self)._init(self._interface_name, obj_path=obj_path)
 
-        self._object_manager = Gio.DBusObjectManagerClient.new_for_bus_sync(
-            Gio.BusType.SYSTEM, Gio.DBusObjectManagerClientFlags.NONE,
-            'org.bluez', '/', None, None, None)
-
-    def find_device(self, address):
```
cschramm: This is moved to Manager, but the usages are not updates, so they still expect `Adapter` to have `find_device`.

blueman/bluez/Manager.py
```diff
@@ -33,16 +33,16 @@ def __new__(cls, *args, **kwargs):
...
-            Gio.BusType.SYSTEM, Gio.DBusObjectManagerClientFlags.NONE,
-            self.__bus_name, '/', None, None, None)
+    def _init(self, **kwargs):
+        super(Manager, self).__init__(
+            bus_type=Gio.BusType.SYSTEM,
+            flags=Gio.DBusObjectManagerClientFlags.NONE,
+            name=self.__bus_name,
+            object_path='/',
+            **kwargs)
+        self._init()
```
cschramm: ?! Infinite recursion.

blueman/bluez/Adapter.py
infirit: Sigh, I thought I did that :cry:, fixed now in master.

blueman/bluez/Manager.py
infirit: Yes, fixed.

infirit:
There is some weird signalling issue in the applet which makes them not being send, the manager is fine so I am confused. But this is not nice to have in the first alpha so reverted the sublcass for now.

infirit:
So, this is a problem only under python2, please can we drop python2 in 2.2 :pray: :smile_cat:

infirit:
Ok, now it seems to have fixed itself, rebased against master and pushed again as c0d7626b37f29d4a9cc32b074dbb2a20e4466df2

PR #578, opened 2016-08-26
Implement a way to compose vNotes

### LINKED ISSUE #10: Feature: Compose and send notes

cschramm:
Closes #10

### DIFF
--- a/blueman/plugins/manager/Notes.py
+++ b/blueman/plugins/manager/Notes.py
@@ -1,0 +1,57 @@
+import datetime
+import os
+from gettext import bind_textdomain_codeset
+from tempfile import NamedTemporaryFile
+
+from blueman.Functions import create_menuitem, get_icon, launch, UI_PATH
+from blueman.plugins.ManagerPlugin import ManagerPlugin
+
+import gi
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk
+
+
+def send_note_cb(dialog, response_id, device_address, text_view):
+    text = text_view.get_buffer().props.text
+    dialog.destroy()
+    if response_id == Gtk.ResponseType.CANCEL:
+        return
+
+    date = datetime.datetime.now().strftime('%Y%m%dT%H%M00')
+    data = ('BEGIN:VNOTE \n'
+            'VERSION:1.1 \n'
+            'BODY;CHARSET=UTF-8: %s \n'
+            'DCREATED:%s \n'
+            'LAST-MODIFIED:%s \n'
+            'CLASS:PUBLIC \n'
+            'X-IRMC-LUID:000001000000 \n'
+            'END:VNOTE \n' % (' '.join(text.splitlines()), date, date))
+
+    tempfile = NamedTemporaryFile(delete=False)
+    tempfile.write(data.encode('utf-8'))
+    tempfile.close()
+    launch('blueman-sendto --device=%s' % device_address, tempfile.name, False, 'blueman')
+    os.unlink(tempfile.name)
+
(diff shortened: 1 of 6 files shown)


blueman/plugins/manager/Notes.py
```diff
@@ -0,0 +1,57 @@
...
+    data = ('BEGIN:VNOTE \n'
+            'VERSION:1.1 \n'
+            'BODY;CHARSET=UTF-8: %s \n'
+            'DCREATED:%s \n'
+            'LAST-MODIFIED:%s \n'
+            'CLASS:PUBLIC \n'
+            'X-IRMC-LUID:000001000000 \n'
+            'END:VNOTE \n' % (' '.join(text.splitlines()), date, date))
+
+    tempfile = NamedTemporaryFile(delete=False)
```
infirit: Not an issue but the ``tempfile`` module supports the ``with`` statement, which is convenient as it cleans up after itself :smile: .

```python
with NamedTemporaryile(delete=False) as tempfile:
    do_stuff
```

cschramm: I don't think that's applicable here as I need to close and unlink separately (before and after running blueman-sendto), is it?

infirit: Wait, ``launch`` does not block so by the time ``blueman-sendto`` is started the file is already unlinked. At least this is what happens on my machine when I tested this. We probably either need to wait for ``blueman-sendto`` to finish (I see no way with launch) or initiate the transfer through the bluez classes and have the callback clean things of on error and success.

cschramm:
I think I found a nice solution now adding a new feature to blueman-sendto as a side product. :smile:

infirit:
After some fixes from my moving things around it does not work for me. See below the applet output. Normal files work fine, is there something I need to do on the phone (Galaxy nexus with custom rom)?

Also, you will want to set the ``activates-default`` on the GtkEntry and then on the ok GtkButton set ``can-default`` so that the enter key will trigger the ok button, at least that is what should happen :smile:.

```
do_g_properties_changed (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Transfer.py:29)
/org/bluez/obex/client/session1/transfer11 Filename /tmp/noteIJHTUL.vnt 
_________
do_g_properties_changed (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Transfer.py:29)
/org/bluez/obex/client/session1/transfer11 Session /org/bluez/obex/client/session1 
_________
do_g_properties_changed (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Transfer.py:29)
/org/bluez/obex/client/session1/transfer11 Size 168 
_________
do_object_added (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Manager.py:49)
/org/bluez/obex/client/session1/transfer11 
_________
do_g_properties_changed (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Transfer.py:29)
/org/bluez/obex/client/session1/transfer11 Status error 
_________
_on_transfer_completed (/usr/lib64/python2.7/site-packages/blueman/bluez/obex/Manager.py:70)
/org/bluez/obex/client/session1/transfer11 False 
```

cschramm:
I improved the dialog.

Android does not support vNotes. There are some applications for Android that do and might enable the phone to receive notes. Unfortunately I do not know of a way to detect if vNotes are supported or not (other than trying to send one and fail or succeed).

PR #579, opened 2016-08-31
Move application classes out of the scripts installed in /usr/bin

infirit:
>This allows us to easily import them in other places and use them.

I have been thinking of doing this but there was no real need for it but now with the notes sending plugin we do. This allows the sendto to be imported and used throughout blueman.


infirit:
Oh, we can debate where they should live and their names :smile:.

cschramm:
I didn't review this carefully, but it should just be cut & paste, right? The change makes sense. :+1:

infirit:
>but it should just be cut & paste, right

Pretty much yes.

Issue #518, opened 2016-05-09
Status icons in bottom right flicker when using headset


stuaxo:
![flickery status icons](http://i.imgur.com/k6fF8Wr.png "Flickery Status Icons")

This happens when I'm using a bluetooth speaker.  I guess, the icons are being updated every there is some sort of event, instead the events should probably set some sort of flag, so the icons can stay animated for some short of time (another one coming in will keep just make it go on longer).

infirit:
That is what is supposed to happen, it indicates the date rate going up and down to the remote device. The faster it blinks the more data is being transmitted.

stuaxo:
Hm, after a certain amount of speed it's just really distracting, a kind of flickery blur between blue and grey...

stuaxo:
+ here's a video  https://youtu.be/IsAZybL2w9k

infirit:
Yes, that is normal to me. The only thing we can have a look at is to see if we need to change the scale. @cschramm opinions?

cschramm:
What scale?

infirit:
The scale on the [blinker speed](https://github.com/blueman-project/blueman/blob/master/blueman/gui/manager/ManagerStats.py#L100). We could add in another step or reduce the rate.

cschramm:
Yeah, why not? Maybe multiples of 2 up to 8?

infirit:
Here is a patch that adds a new level and reduces the rate to be multiples of 2.
https://gist.github.com/infirit/2178914f243c18227fc4c6f7d9bc4c41

cschramm:
:+1:

stuaxo:
I'll have a go with the patch, is it worth having a maximum speed as well ?    Since it's python, might well have a poke @ it..

PR #585, opened 2016-09-07
ManagerStats: Reduce blink rate

infirit:
I still had this one lying around. The blink rate is a bit high with ad2p audio being streamed or files transfered. This reduces the rate but also adds a step for a more fluent progression.


cschramm:
Reference: #518 
Still :+1: :smile:

infirit:
Thx, merged, but with a rebase so need to close manually.

PR #587, opened 2016-09-09
WIP: NMDUnSupport: Port to GDBus

infirit:
Completely untested but this is a good base to start with.


blueman/plugins/applet/NMDUNSupport.py
```diff
@@ -18,8 +17,12 @@ def __init__(self, parent, service, reply, err):
         self.rfcomm_dev = None
         self.timeout = None
 
-        self.parent.bus.add_signal_receiver(self.on_interfaces_added, "InterfacesAdded",
-                                            "org.freedesktop.DBus.ObjectManager")
+        self.signal_sub = self.parent._bus.signal_receiver("org.freedesktop.ModemManager1",
```
cschramm: This should be `signal_subscribe`.

blueman/plugins/applet/NMDUNSupport.py
```diff
@@ -18,8 +17,12 @@ def __init__(self, parent, service, reply, err):
...
+                                                           "org.freedesktop.DBus.ObjectManager",
+                                                           "InterfacesAdded",
+                                                           "/org/freedesktop/ModemManager1",
```
cschramm: There's a `None` argument missing between the path and the flags.

infirit:
Rebased and fixed the comments. It will need more changes for the icon and the weird timeout handling which I will do now.

cschramm:
With the mentioned changes this works as far as I can test it with an Android phone that exposes a DUN service.

I'm confused on whether our DUN handling works at all (no matter if via ModemManager or our `PPPSupport` plugin). First of all, our `get_rfcomm_channel` function is bound to the serial port service class and does not find a channel in my case. Also if I change the implementation to the DUN service class I do not find any channel. I do find a couple of channels (number 2 to 9) if I use `PUBLIC_BROWSE_GROUP` (`0x1002`). I can connect to them but they do not actually seem to have something to do with DUN and thus ModemManager does not pick them up. For some of them, my phone asks me to grant access to messages, contacts, or the SIM.

So I could not test the `on_interfaces_added` method and beyond and am left puzzled why my phone should offer a DUN service but no RFCOMM channel for it... :thinking:

blueman/plugins/applet/NMDUNSupport.py
```diff
@@ -18,18 +17,20 @@ def __init__(self, parent, service, reply, err):
         self.rfcomm_dev = None
         self.timeout = None
 
-        self.parent.bus.add_signal_receiver(self.on_interfaces_added, "InterfacesAdded",
-                                            "org.freedesktop.DBus.ObjectManager")
+        self.signal_sub = self.parent._bus.signal_subscribe("org.freedesktop.ModemManager1",
+                                                            "org.freedesktop.DBus.ObjectManager",
+                                                            "InterfacesAdded",
+                                                            None,
+                                                            "/org/freedesktop/ModemManager1",
```
cschramm: I'm pretty sure `None` and `"/org/freedesktop/ModemManager1"` need to appear the other way around.

infirit:
I just came across an old phone, the n900 which may have proper dun. It may take a while to get it up and running as the battery is most likely dead. But who knows and I am lucky

infirit:
Using the bluez tools I can find the sevice and the rfcomm channel. But we are not finding the rfcomm channel.

```
 # sdptool browse 3C:F7:2A:07:BF:E3
Browsing 3C:F7:2A:07:BF:E3 ...
Service Name: Dial-Up Networking
Service RecHandle: 0x10000
Service Class ID List:
  "Dialup Networking" (0x1103)
  "Generic Networking" (0x1201)
Protocol Descriptor List:
  "L2CAP" (0x0100)
  "RFCOMM" (0x0003)
    Channel: 1
Profile Descriptor List:
  "Dialup Networking" (0x1103)
    Version: 0x0100
```

cschramm:
Looks much better than for me. Finding the channel cannot work as our code is bound to the serial service class. Check https://github.com/cschramm/blueman/commit/86b695217f9be9d5064ab646e6cd8f502845ed14 where I pass the UUID as an additional parameter instead.

I just checked my phone again with `sdptool` and could identify all RFCOMM services. None of them is a DUN service. No idea why it offers one then...

infirit:
>Check cschramm@86b6952 where I pass the UUID as an additional parameter instead.

I get the correct channel for the n900 so I'll do some testing and fixup any problems I see with this phone.

infirit:
I took your commit for the port number and added it to my branch, works perfectly thanks! I can keep it in the branch but if you want feel free to push it to master :smile:.

Pushed a few more commits to fix various things I came across today. Too hot for my brain to continue today.

infirit:
I have the NMDUNSupport plugin working. The changes I had to make makes me to think it was broken before but did not bother to go back and check. However it did not show up in NetworkManager as a modem to be used. This may be because it has no sim but I don't have a full sized one in my current phone. I may have the broken off sim border somewhere.. but in any case it functions as it was designed whether it is actually useful I can't say right now.

With PPPSupport connecting in blueman-manager is timing out a (no idea which) dbus call. When I figure that out DUN I can continue fixing that, just not today :smile:.

My suggestion is to merge the bits here open an issue for the PPPSupport plugin so it can be added to the 2.1 milestone.

PR #591, opened 2016-09-21
PowerManager: Auto power adapter in on_adapter_added

### LINKED ISSUE #571: Blueman assuming wrong powered state
blueman-git from Sunday (should be the latest)
bluez-5.39
Gentoo Linux

When blueman-applet starts, it tends to power on my Bluetooth adapter (BCM20702).

However if the applet is already running when the adapter appears (e.g. after a suspend, or after disabling/enabling it via hardware switch), the adapter defaults to being powered off. Blueman will pretend that it's on ("Bluetooth enabled").

This can be worked around by restarting blueman-applet or by issuing "power on" in bluetoothctl.

infirit:
We only run on_manager_state_changed when the dbus name (dis)appears. Which
means any adapter added later will not be auto powered. Fixes #571

### DIFF
--- a/blueman/plugins/applet/PowerManager.py
+++ b/blueman/plugins/applet/PowerManager.py
@@ -51,6 +51,7 @@ class PowerManager(AppletPlugin):
 
         self.adapter_state = True
         self.current_state = True
+        self.user_power_override = False
 
         self.request_in_progress = False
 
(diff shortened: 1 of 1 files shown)


cschramm:
See my comment in #571. I'm fine with that if nobody else objects.

Issue #595, opened 2016-09-28
Dialogue box for choosing a device misleads


CottonEaster:
It seems to me that there is something rather confusing about the window/dialogue box for choosing a device as a recipient. Specifically, I have in mind the bar at the bottom. That bar gave me the impression that had to wait, or perhaps that, no, I should not wait, because there will be a timeout. In reality, the bar indicates - I think - a search for as-yet-unlisted bluetooth devices. The fact that Android has a similar bluetooth interface doesn't make it less confusing. How about an explanatory label?

Also, any chance of a 'send to' in the context-menu of file managers such as Nemo?

cschramm:
Regarding the "send to": We once shipped a nautlius-sendto plugin until they dropped plugin support. We discussed basically the same thing for caja in #220. There's a feature request at https://github.com/mate-desktop/caja-extensions/issues/12. We will look into that for blueman 2.1. The repository for nemo would be https://github.com/linuxmint/nemo-extensions. I guess this is very similar except that we do not have the GTK+ 2 barrier.

infirit:
So, this is what I like to do.

 * Turn the progress bar into a spinner and have a label with a text, something like "Discovering devices".
 * Do not timeout on discovery  any more and stop discovery when device is selected or app is closed.
 * Add a small delay for adapters that can not communicate properly when discovering

Objections?

ACinnamonUserInUK:
Sounds good to me, though for the moment I've given up on Bluetooth / replaced it with wifi technologies.

infirit:
First WIP PR  #744

infirit:
#744 was merged, closing.

PR #597, opened 2016-10-04
Simple sendto plugin for caja, nemo and nautilus

infirit:
Needs a little cleanup still but is functional. Comments, criticism etc welcome.

Instead of doing a very complicated extension to use the DeviceSelectorWidget just call ``blueman-sendto`` with the appropriate argument for the files. ref #220

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,7 +3,8 @@ SUBDIRS = 		\
 	blueman		\
 	data		\
 	module		\
-	po
+	po		\
+	sendto
 
 DIST_DOCS =			\
 	CHANGELOG.md		\
--- a/configure.ac
+++ b/configure.ac
@@ -137,6 +137,36 @@ AM_CONDITIONAL([HAVE_POLKIT], [test "x$have_polkit" = "xyes"])
 
 AC_SUBST([have_polkit])
 
+AC_ARG_ENABLE([caja-sendto],
+  AS_HELP_STRING([--enable-caja-sendto], [Enable caja-sendto installation]),
+  [enable_caja=$enableval], [enable_caja=yes])
+
+AS_IF([test "x$enable_caja" = "xyes"],
+  [PKG_CHECK_EXISTS([caja-python], [have_caja_python=yes], [have_caja_python=no])],
+  [have_caja_python=no])
+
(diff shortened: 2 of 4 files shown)


infirit:
Screenie

cschramm:
I cannot confirm that it works as caja is currently segfaulting for me, but I can confirm the same snippet works for nemo (#595) if I just replace `Caja` with `Nemo`. :smiley: It also works for nautilus by using `Nautilus` but the icon is missing and `get_background_items` needs to be implemented and return a (empty) list to avoid a warning.

infirit:
>I cannot confirm that it works as caja is currently segfaulting for me

Ouch, I'll try it in a VM, Debian testing?

The python bindings have not changed between the various forks and the original. The extension should work in all of them with little modification. However, Gnome (thus nautilus) decided in their infinite wisdom that menu's shouldn't have icons, that is why its missing.

I'll add a stub for ``get_background_items`` so that if people want to use it under Nautilus their logs are not spammed.

cschramm:
Segfault resolved. My GTK+ was too recent / caja was too old. caja 1.16 was added to unstable just yesterday and works fine. :smiley: 

Regarding Gnome's infinite wisdom: Right, other items do not have icons as well, just the main entry ("Open with [default app]") does.

So it looks like we can just put in `Caja`, `Nemo`, and `Nautilus` with some autotools voodoo and cover all three with just one script.

infirit:
>with some autotools voodoo and cover all three with just one script.

voodoo added :smiling_imp:

Gentoo does not come with [nemo-python](https://github.com/linuxmint/nemo-extensions/blob/master/nemo-python/nemo-python.pc.in#L7) but checking the source it is identical to caja and nautilus.

I still want to use Gio to launch ``blueman-sendto`` instead of ``os.popen``

infirit:
Now using Gio to start ``blueman-sendto``. We could at some point use the our ```Manager``` class to add a submenu with the paired devices. But for now this should do I think :smile:.

cschramm:
Awesome. Installs and works fine for me for all three. There's a typo in the help strings, but who cares. :laughing:

infirit:
>There's a typo in the help strings

Oops, I see them in configure.ac. Fixed those. unless something comes up I'll merge this in a couple of days.

PR #599, opened 2016-10-06
Delay hiding the status icon when ...

### LINKED ISSUE #314: Blueman icon briefly vanishes when turning bluetooth on
If Bluetooth is off and you turn it on using a right-click on the icon and selecting "Turn Bluetooth On", the icon will vanish for a quarter of a second and then reappear.  The icon should change from the off type to the on type without doing this vanishing act.

cschramm:
...bluetooth gets re-enabled.

Right after re-enabling bluetooth no adapters are available so the icon vanishes for a short moment until adapters are presented. To avoid that, just wait a second for adapters to show up before hiding the icon if none are present at all.

Fixes #314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * Icon disappeared when switching off bluetooth
 * Revert "bluez manager: Subclass from GDBusObjectManagerClient"
+* Icon briefly vanished when turning on bluetooth
 
 ## 2.1.alpha1
 
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -104,7 +104,7 @@ class KillSwitch(AppletPlugin):
 
         logging.info("State: %s" % self._enabled)
 
-        self.Applet.Plugins.StatusIcon.QueryVisibility()
+        self.Applet.Plugins.StatusIcon.QueryVisibility(delay_hiding=not self._hardblocked)
         self.Applet.Plugins.PowerManager.UpdatePowerState()
 
         return True
--- a/blueman/plugins/applet/StatusIcon.py
+++ b/blueman/plugins/applet/StatusIcon.py
@@ -82,6 +82,7 @@ class StatusIcon(AppletPlugin, Gtk.StatusIcon):
 
     def on_visibility_timeout(self):
         GLib.source_remove(self.visibility_timeout)
+        self.visibility_timeout = None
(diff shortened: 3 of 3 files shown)


cschramm:
I couldn't test the case where bluetooth gets enabled but actually no adapters are present as my current system activates (i.e. modprobes the driver for) the internal adapter when I turn on bluetooth. Probably an enabled state would be shown for a moment before the icon vanishes which is a little odd but I guess that's much better than the issue we have on every re-enabling.

infirit:
I have 2 usb adapters so i can test that. Other then the no adapter scenario are there any others you like me to test?
-- 

infirit:
Reporting on my experience with this patch:

This happens with and without this patch. When starting blueman-appler the icon still briefly shows and then disappears when no adapter is present.  Perhaps we should assume adapter_state  ``False`` in the [PowerManager](https://github.com/blueman-project/blueman/blob/master/blueman/plugins/applet/PowerManager.py#L52) plugin?


But more importantly, stopping bluetooth (bluez) does not make the icon go away, have not looked why but this should cause ``on_manager_state_changed`` to run with a ``False`` state. However it does stop the icon from disappearing so we are partly there :smile:.

cschramm:
> Perhaps we should assume adapter_state False in the PowerManager plugin?

I guess that makes sense. Did you try it?

> stopping bluetooth (bluez) does not make the icon go away

Hm. Works for me...

infirit:
[Here is the log](https://gist.github.com/infirit/30b54527311dca6dd6d796185e006d1d), this goes from applet start -> insert bt dongle -> stop bluetooth service

I tried setting the adapter_state to False but made no difference. I suspect we always show the StatusIcon regardless of state. I'll poke at that later :smile:.

cschramm:
Resolved. Depended on the sequence of things you do as the timeout state never got reset.

Also fixed another issue where the hidden icon got visible again when the icon got changed. I reproducibly ran into that by starting and then immediately stopping the bluetooth service. As one would expect the icon showed up and vanished accordingly, but then showed up again with a disabled state as PowerManager noticed a state change and called StatusIcon.IconShouldChange. It looks like setting the icon-name property makes an invisible GtkStatusIcon visible. Although I could not find anything about that in the documentation, setting the visibility again after (there's potential for a short blink here :worried:) setting the icon-name makes it work.

infirit:
I tried with the latest changes, works ok now. Just needs a rebase against master before merging.

PR #601, opened 2016-10-17
Use pgrep to check for running processes

infirit:
Untested, see https://github.com/blueman-project/blueman/issues/524#issuecomment-220964106

### DIFF
--- a/blueman/plugins/applet/PPPSupport.py
+++ b/blueman/plugins/applet/PPPSupport.py
@@ -13,7 +13,7 @@ from blueman.main.Config import Config
 from gi.repository import GObject
 
 from blueman.Sdp import uuid128_to_uuid16, DIALUP_NET_SVCLASS_ID
-import os
+import subprocess
 
 
 class Connection:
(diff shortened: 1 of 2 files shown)


infirit:
Added another commit for PPPSupport to change the binary name to ModemManager.

infirit:
Tested now, seems to work nicely. As usual, I have no DUN phone to test the modem manager changes.

cschramm:
Adds procps as a dependency. Fine with that.

Why `modem-manager` -> `ModemManager`?

infirit:
The binary is named like that on my system and my debian vm. It was done already in [2012](https://cgit.freedesktop.org/ModemManager/ModemManager/commit/?id=7d9d003324bb6a466104c7d935758ca29aeac9e5)

infirit:
>Adds procps as a dependency

btw, I can also fix the grep without ``pgrep``. Simply adding brackets around a single letter and it will exclude the grep process. Either way is fine with me :smile:

cschramm:
> The binary is named like that

Looks like it, yes. I did not run into that as #526 drops the name anyway. I guess if we land that, there's no reason for this patch anymore, right?

infirit:
>I guess if we land that, there's no reason for this patch anymore, right?

Yep, but I do like to put this in 2.0 which will not get #526. I'll open a new PR without adding the procps dep.

infirit:
Closing, #526 gets rid of these bits.

PR #602, opened 2016-10-22
DBusService: make QueryAvailablePlugins return a list of keys instead…

infirit:
… of a dict

 The signature and usage elsewhere indicates we should return a list of plugin names similar to
 QueryPlugins. Python-dbus appears to extract the keys so it never causes issues but will late
 when moving to GDBus.


cschramm:
:+1:

PR #608, opened 2016-10-31
Recentconns: Do not use service object

### LINKED ISSUE #607: Reset RecentConns
In [Debian #841869](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=841869) Jelmer Vernooĳ reports a broken headset entry in the recent connections menu. This is due to the removal of `blueman/services/Headset.py` in #442.

To avoid such issues /org/blueman/plugins/recentconns/recent-connections should probably be reset on every `make install`.

infirit:
untested but this is what I was thinking of to fix #607.


infirit:
Tested now and also added some small cleanups

cschramm:
:+1: :smiley:

PR #614, opened 2016-11-06
AppletPlugin: Remove redundant _applet variable

infirit:


cschramm:
Looks like TransferService uses it.

infirit:
Updated the TransferService the use ``AppletPlugin.Applet`` which is the same object and already used everywhere else. There are still references to ``_applet`` in the Agent class which is correct as it gets it passed. Commit and PR updated with that.

cschramm:
:+1:

PR #616, opened 2016-11-14
A couple cleanups and one small change

infirit:


cschramm:
Looks good. I guess we can remove libnotify from https://github.com/blueman-project/blueman/blob/master/Dependencies.md then. 😃

cschramm:
On second thought, I think we must not remove the libnotify dependency, but replace it with something like "A notification deamon that complies with the Desktop Notifications Specification" as blueman isn't prepared to not find the `org.freedesktop.Notifications` service.

infirit:
>as blueman isn't prepared to not find the org.freedesktop.Notifications service.

What do you mean, do we crash if the dbus name is not provided by anything? I hope not as that is then a bug :smile:.

When merging (in the next few days) I'll add something to the deps list. Any suggestion on how we deal with the deps being different between 2.0 and master?

cschramm:
Ah, no. I didn't really try it but of course it should *not* crash as we have our fallback for that case. Then we can just remove the dependency. 🙂 

I'd keep the Dependencies.md in sync with master. Anybody packaging or building should be clever enough to consult the Dependencies.md for (and bundled in) the version he's working with. Or do you mean something else?

infirit:
Ah, I thought Dependencies was a wiki page but its part of the repo. All clear and updated the file to reflect the changes wrt to notifications :smile:.

PR #617, opened 2016-11-15
Gdbus services

infirit:
This is the latest on the DBus service porting to Gio. I am fairly happy with this now although I still need to work on mechanism a bit. But for the most part things work well now.

This is largely based of the [pithos implementation](https://github.com/pithos/pithos/blob/master/pithos/plugins/dbus_util/DBusServiceObject.py) with an additional object to handle sender, path and async callbacks. I had to modify the implementation a bit so we can easily re-export the methods and signals.

Appreciate feedback, criticism etc :smile: 

edit: Two things broke
 * The signal wrapper from Pithos folks does not work. Need to modify the wrapper so it calls a new function on the ApplerPlugin to emit the signal. Or we use said function directly.
 * RecentConns plugin calls the ``connect_service`` directly and now fails dues to the changes wrt callback handling.

### DIFF
--- a/apps/blueman-mechanism.in
+++ b/apps/blueman-mechanism.in
@@ -13,10 +13,6 @@ if 'BLUEMAN_SOURCE' in os.environ:
     timeout = 9999
     os.environ["GSETTINGS_SCHEMA_DIR"] = os.path.join(_dirname, "data")
 
-import dbus
-import dbus.service
-from dbus.mainloop.glib import DBusGMainLoop
-
 from gi.repository import Gio, GLib
 
 loop = GLib.MainLoop()
@@ -79,8 +75,6 @@ sys.stderr = sl
 
 logging.info("Starting blueman-mechanism")
 
-DBusGMainLoop(set_as_default=True)
-
 os.environ["PATH"] = "/usr/bin:/bin:/usr/sbin:/sbin"
 
 dhcp_pids = []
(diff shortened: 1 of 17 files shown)


infirit:
Added a commit that changes the ``EnableNetwork`` mechanism dbus method, https://github.com/blueman-project/blueman/pull/617/commits/e6c425f1a8fce92c9fc08fc683012d4ae05840e2.

cschramm:
> The signal wrapper from Pithos folks does not work.

This only affects `BluetoothStatusChanged` at the moment, right? What's the problem?

infirit:
>What's the problem?

It makes assumption on where the function is called from. It expects to be on the [DBusService](https://github.com/blueman-project/blueman/pull/617/files#diff-cd501e51f5c1449a2cfcf82a2c9449a0R201) object but with our plugins this assumption is false. It should be easy enough to point to the correct place but have not tried yet.

infirit:
Rebased against current master and added workaround for signals. I'll be running this branch myself and keep rebasing it against master.

Overall I'm not too happy with the workaround but they "work" and I see no other solution at the moment. If people are ok with how things are now then lets move forward if not let me know.

infirit:
Thought of a better way to handle signals. Instead of dropping the wrapper add the necessary bits to the AppletPlugin class so that the assumption in the wrapper is covered.

I have not actually tested properties but as we do not use them at the moment not sure how important this is for blueman.

infirit:
Rebased against master.

Also properties do not work on plugins as I suspected. Is that important enough to delay this?

cschramm:
So https://github.com/blueman-project/blueman/pull/617/commits/ecc278903c3a1ff77a0f4e05a0e5a26d301cb412 does not have the desired effect?

infirit:
If I remember correctly they don't even show up on the dbus interface when I add a property through the decorator.

I'll poke at it some more this week to figure out what is going on.

PR #622, opened 2016-12-01
Makes it possible to use the gschema from the source tree

infirit:
With this it is possible to use the compiled schema from the source tree instead of the one installed in ``/usr/share/glib-2.0/schemas``. One less thing to install 😄 

At some point though we should probably add a new install target that only installs the bits we can not do without like the dbus services.


cschramm:
👍

PR #626, opened 2016-12-10
Avoid using get_properties

infirit:
Typically we do not need all properties and at most we use two, retrieve them directly instead.

Along the way I also found some unused variables, eccentricities (like get_device_alias) and checking for non optional properties. I put those in separate commits for easier review and debugging.


cschramm:
Looks good.

Issue #620, opened 2016-11-26
UnicodeDecodeErrors since a9357f30fded4f5934d99e07e23f19d74da18952


infirit:
I have not run python 3 for a while (to make sure we do not break 2.7) but a9357f30fded4f5934d99e07e23f19d74da18952 introduced UnicodeDecodeError under python 3 but it is not immediately obvious why. They show up for me when searching for devices just when discovery is stopped. Relevant bits from the logs below.
blueman-applet:

```
on_adapter_property_changed (/usr/lib64/python3.5/site-packages/blueman/plugins/applet/DiscvManager.py:89)
prop Discovering True 
_________
do_g_properties_changed (/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py:81)
/org/bluez/hci0 Discovering True 
_________
do_g_properties_changed (/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py:81)
/org/bluez/hci0/dev_7C_2F_80_3B_AF_AB RSSI -79 
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb8 in position 1: invalid start byte
_________
on_adapter_property_changed (/usr/lib64/python3.5/site-packages/blueman/plugins/applet/DiscvManager.py:89)
prop Discovering False 

```

blueman-manager:
```
_________
do_g_properties_changed (/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py:81)
/org/bluez/hci0 Discovering True 
_________
do_g_properties_changed (/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py:81)
/org/bluez/hci0/dev_20_D4_E4_F6_C1_D6 RSSI -77 
_________
row_update_event (/usr/lib64/python3.5/site-packages/blueman/gui/manager/ManagerDeviceList.py:269)
row update event RSSI -77 
opacity 159
_________
do_g_properties_changed (/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py:81)
/org/bluez/hci0/dev_D0_03_4B_E2_83_97 RSSI -78 
_________
row_update_event (/usr/lib64/python3.5/site-packages/blueman/gui/manager/ManagerDeviceList.py:269)
row update event RSSI -78 
opacity 156
_________
do_g_properties_changed (/usr/lib64/python3.5/site-packages/blueman/bluez/Base.py:81)
/org/bluez/hci0/dev_7C_2F_80_3B_AF_AB RSSI -79 
_________
[...]

PR #627, opened 2016-12-11
Too broad exception handling

infirit:
This is a WIP branch to get rid off too broad exception handling. Due to #620 I needed to review exception anyway so took the opertunity to fix them and document the ones not fixed yet. This is not all of them but will update as I go through the code.

edit: All except one sorted out now.


infirit:
Pushed some of these already as part of other commits and added some more to leave only one. Removed WIP and waiting for a quick review 😄

PR #628, opened 2016-12-13
Bluez Base: WOrkaround python3 UnicodeDecodeError

infirit:
The invalidated properties in do_g_properties_changed seems to be broken.


cschramm:
Can't see any downside. 👍

PR #629, opened 2016-12-14
Small rework and bug fix

infirit:
I have LE devices being picked up by my machine, that are not mine, which do not expose the device class. BlueZ then sends us a zero.

The question is, should we skip it or should we just show zero? This PR does the latter.

I also reworked the properties a bit and added it to the PR.


infirit:
Ok, I have one LE device which is my Playstation3 controller. It does not expose UUIDs, Class and probably some others __until__ we have connect to it. So I changed my approach here to skip properties that fail to be added to the store and print the name.

cschramm:
Seems necessary. 👍

PR #632, opened 2016-12-24
Fixes for file transfer and new feature

Rudd-O:
Two things:

* Fixes so file transfer completes and doesn't leave resources dangling in case of failure.
* Optional command processing of received files (suitable for example to be used with https://github.com/Rudd-O/organizer )

Ready to be merged in master.

### DIFF
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -254,6 +254,7 @@ class TransferService(AppletPlugin):
             now = datetime.now()
             filename = "%s_%s" % (now.strftime("%Y%m%d%H%M%S"), filename)
             logging.info("Destination file exists, renaming to: %s" % filename)
+            dest = os.path.join(dest_dir, filename)
 
         try:
             shutil.move(src, dest)
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -256,9 +256,10 @@ class TransferService(AppletPlugin):
             logging.info("Destination file exists, renaming to: %s" % filename)
             dest = os.path.join(dest_dir, filename)
 
+        e = None
         try:
             shutil.move(src, dest)
-        except Exception:
+        except Exception as e:
             logging.error("Failed to move files", exc_info=True)
             success = False
 
@@ -271,10 +272,12 @@ class TransferService(AppletPlugin):
             self._add_open(n, "Open", dest)
             n.show()
         elif not success:
+            strerror = ": " + escape(str(e)) if e is not None else ""
             n = Notification(_("Transfer failed"),
-                         _("Transfer of file %(0)s failed") % {
+                         _("Transfer of file %(0)s from %(1)s failed%(2)s") % {
                              "0": "<b>" + escape(filename) + "</b>",
-                             "1": "<b>" + escape(attributes['name']) + "</b>"},
+                             "1": "<b>" + escape(attributes['name']) + "</b>",
+                             "2": strerror},
                          **self._notify_kwargs)
             n.show()
             if attributes['size'] > 350000:
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -17,6 +17,7 @@ from blueman.Functions import get_icon, launch
 from blueman.gui.Notification import Notification
 from blueman.plugins.AppletPlugin import AppletPlugin
 from blueman.main.Config import Config
+from blueman.main.FileProcessor import FileProcessor
 
 if sys.version_info.major == 2:
     from cgi import escape
@@ -286,6 +287,55 @@ class TransferService(AppletPlugin):
                 self._silent_transfers -= 1
 
         del self._agent.transfers[transfer_path]
+
+        if success:
+            command_to_run = self._config["process-command"]
+            # No need to decode bytes into UTF-8.
+            # That functionality is covered in the FileProcessor.
+
+            if command_to_run:
+                try:
+                    processor = FileProcessor(command_to_run, filename, dest)
+                    # The handler _on_process_ended will take care of notifying
+                    # the user about success or particular errors during the
+                    # execution of the command that the user configured.
+                    processor.connect(
+                        "started",
+                        lambda _, fn: logging.info(
+                            "Now processing %s with command %s" % (
+                                fn, processor.get_command_line_as_utf8()
+                            )
+                        )
+                    )
+                    processor.connect("ended", self._on_process_ended)
+                    processor.connect("failed", self._on_process_ended)
+                    processor.start()
+                except Exception as e:
+                    if isinstance(e, ValueError):
+                        # Oops.  There is a configuration error.  We must inform the user.
+                        e = ("The command configured to run after receiving a file "
+                             "has a syntax error and cannot be executed as is.")
+                    n = Notification(_("Cannot process received file"),
+                                _("Processing of file %(0)s failed: %(1)s") % {
+                                    "0": "<b>" + escape(filename) + "</b>",
+                                    "1": "<b>" + escape(str(e)) + "</b>"},
+                                **self._notify_kwargs)
+                    n.show()
+                    logging.error("Failed to move files", exc_info=True)
+
+    def _on_process_ended(self, unused_processor, filename, exception):
+        if exception is None:
+            n = Notification(_("File processed"),
+                             _("File %(0)s successfully processed") % {
+                                 "0": "<b>" + escape(filename) + "</b>",},
+                             **self._notify_kwargs)
+        else:
+            n = Notification(_("Processing failed"),
+                         _("Processing of file %(0)s failed: %(1)s") % {
+                             "0": "<b>" + escape(filename) + "</b>",
+                             "1": "<b>" + escape(str(exception)) + "</b>"},
+                         **self._notify_kwargs)
+        n.show()
 
     def _on_session_removed(self, _manager, _session_path):
         if self._silent_transfers == 0:
--- a/data/ui/services-transfer.ui
+++ b/data/ui/services-transfer.ui
@@ -69,6 +69,68 @@
             <property name="width">2</property>
           </packing>
         </child>
+        <child>
+          <object class="GtkLabel" id="process_label">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">start</property>
+            <property name="label" translatable="yes">&lt;b&gt;File Processing&lt;/b&gt;</property>
+            <property name="use_markup">True</property>
+          </object>
+          <packing>
+            <property name="left_attach">0</property>
+            <property name="top_attach">3</property>
+            <property name="width">2</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkLabel" id="process_title_label">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">start</property>
+            <property name="xalign">0.0</property>
+            <property name="margin_left">12</property>
+            <property name="label" translatable="yes">Command to run after file is received:</property>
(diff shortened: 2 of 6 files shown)


Rudd-O:
Sorry, I do not know how to add the new translations.  But I can be educated about it.

Rudd-O:
I have made some updates to address the remaining issues that didn't get fixed on master.  I can confirm that master works properly with your fixes and better with these fixes of mine.  Just merge.

infirit:
I took the first fixes commit and split them up and pushed them to master. I also took the new fixes and did the same. Thanks for them.

I'll comment later on the new feature.

Rudd-O:
Happy to hear your comments.

Rudd-O:
I do think it belongs on the project, because a more generic solution (example: an inotify-based solution) is not as flexible, does not integrate well with *received files via Bluetooth*, and is obviously more error-prone.

Bear in mind the generic solution is what I started this patch with, and I tried, and I saw it was not reliable, so I scrapped it and I decided to hack on this particular enhancement instead.

I will address the commit comments now, and remove the other commits that already made their way into master.  BRB.

blueman/plugins/applet/TransferService.py
```diff
@@ -23,6 +25,33 @@
     from html import escape
 
 
+@GObject.type_register
+class CommandRunner(threading.Thread, GObject.Object):
```
Rudd-O: That's not correct.  `idle_add` will cause the execution of this code to freeze the Python interpreter in `subprocess.check_call` because that code will run on the main application thread.  That code needs to run asynchronously and in parallel to other tasks that the TransferService may be doing.

This is the correct solution to the problem — spawn a thread to monitor the process, reap the process (all done by `subprocess.check_call`) and then use `idle_add` to signal the return value of the process to the event handler subscribed to this class' outcome.

blueman/plugins/applet/TransferService.py
```diff
@@ -23,6 +25,33 @@
...
+        threading.Thread.__init__(self)
+        GObject.Object.__init__(self)
+        self.setDaemon(True)
+        self.command = command
+        self.filename = filename
+        self.path = path
+
+    def run(self):
+        try:
+            subprocess.check_call([self.command, self.path])
```
Rudd-O: In the scenario you posit, from the standpoint of TransferService and blueman-applet, nothing adverse happens.  All that happens is that this thread remains active until blueman-applet exits.  That's about 64K memory use for the thread stack.  That is the cost we pay to show a nice notification to the user when the command completes either successfully or with an error, rather than swallowing the outcome and leaving the user in limbo.

blueman/plugins/applet/TransferService.py
```diff
@@ -23,6 +25,33 @@
...
+            GObject.idle_add(lambda: self.emit("ended", self.filename, None))
```
Rudd-O: Roger.  Updated commit incoming.

blueman/plugins/applet/TransferService.py
```diff
@@ -252,6 +281,7 @@ def _on_transfer_completed(self, _manager, transfer_path, success):
             now = datetime.now()
             filename = "%s_%s" % (now.strftime("%Y%m%d%H%M%S"), filename)
             dprint("Destination file exists, renaming to: %s" % filename)
+            dest = os.path.join(dest_dir, filename)
```
Rudd-O: It's not a duplicate.  `filename` changes in between this line and the prior one.  This is a bug that I am fixing, meant to actually enact the wishes of the three lines of code above it.

I will spin this off to a separate commit, so it won't be mixed with the feature commit.

data/ui/services-transfer.ui
```diff
@@ -69,6 +69,63 @@
...
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">start</property>
+            <property name="xalign">0.0</property>
+            <property name="margin_left">12</property>
+            <property name="label" translatable="yes">Command to run after file is received:</property>
+          </object>
+          <packing>
+            <property name="left_attach">0</property>
+            <property name="top_attach">4</property>
```
Rudd-O: Done.  Thanks. Update incoming.

data/ui/services-transfer.ui
```diff
@@ -69,6 +69,63 @@
...
+          <object class="GtkEntry" id="process-command">
+            <property name="visible">True</property>
+            <property name="can_focus">True</property>
+            <property name="halign">start</property>
+            <property name="margin_left">12</property>
+            <property name="text"></property>
+          </object>
+          <packing>
+            <property name="left_attach">0</property>
+            <property name="top_attach">5</property>
```
Rudd-O: Done.  Thanks. Update incoming.

data/ui/services-transfer.ui
```diff
@@ -69,6 +69,63 @@
...
+            <property name="label" translatable="yes">&lt;i&gt;Command runs with full path to the file as first parameter.&lt;/i&gt;</property>
+            <property name="halign">start</property>
+            <property name="xalign">0.0</property>
+            <property name="margin_left">12</property>
+            <property name="use_markup">True</property>
+            <property name="wrap">True</property>
+          </object>
+          <packing>
+            <property name="left_attach">0</property>
+            <property name="top_attach">6</property>
```
Rudd-O: Done.  Thanks. Update incoming.

blueman/plugins/applet/TransferService.py
infirit: Ah yes, you are correct on idle_add however... I rather than like to see ``GLib.spawn_async`` used and utilize glibs threading and scheduling capabilities instead of adding subprocess plus threading. Something like below should be just fine for this use case.

```python
from gi.repository import GLib

def on_child_exit(pid, *args, **kwargs):
    print("Closing child with pid %s" % pid)
    GLib.spawn_close_pid(pid)

command = ["/usr/bn/sleep", "10"]

try:
    pid, stdin, stdout, stderr = GLib.spawn_async(
        command, flags=GLib.SPAWN_DO_NOT_REAP_CHILD)
    GLib.child_watch_add(pid, on_child_exit)
except GLib.GError as e:
    print(e.message)
```

Rudd-O:
I will think about your request.  Right now I gotta hit the sack.

Consider this for a minute: one thing to note about your solution is
that the spawn_async function is not clear as to what error it returns
if the program the user configured does not exist, whereas in the
subprocess code it's 100% clear that you will get an OSError which will
then be displayed to the user properly on screen.

Consider this too: we know the current code runs reliably in all Python
versions.

infirit:
>the spawn_async function is not clear as to what error it returns if the program the user configured does not exist

It will raise an exception, GLib.Error, any failure to start the command will. The GLib.Error will contain the error message and code. When started successfully an exit result is sent to the callback where you can apply the same rule that any non zero result is a failure. 

>we know the current code runs reliably in all Python versions.

And GLib.spawn_async will run just as reliably without adding threading. I really do not want to introduce another threading system when we already use GLibs.

cschramm:
Sorry for the delay. I'm not sure if it should go into blueman or not. "a more generic solution [...] does not integrate well with received files via Bluetooth" seems to make sense. I don't get the "not as flexible, [...] more error-prone" points. In what way is the (inotify-based?) solution you tried unreliable?

Rudd-O:
I understand.

The first barrier I noticed was integration — which folder should be monitored for incoming files, and what files should be processed by the command?  Given that my first solution was external to blueman.  It did not work very well, it did not discriminate between files which were placed in the dest folder by blueman and files that were placed there by other programs.  It also required oodles of code just to (a) find out which folder to monitor, which is tied to blueman's config, and is by default the home directory, where many other files are created and modified by other programs (b) react appropriately and correctly to inotify events.

It's why I abandoned that approach, and went for the far cleaner approach of letting blueman be the program that starts the processing — which makes logical sense, because only blueman knows which files it has placed in the dest folder.  That was, in fact, all I wanted from the start.

My patch is working very well on my computer — I use it to process received files with a program that moves them to a different Qubes VM — and I think more people could benefit from this feature.

Given that it's an opt-in thing, and it's just a few lines of code, it doesn't affect anyone who doesn't change the default configuration, and it won't create an undue maintenance burden, so I see very little harm in including it.

Rudd-O:
I will very shortly follow the instructions to update the code to the new upstream, and to replace the normal exceptions with the GLib.Error code that was suggested earlier.  Later today, probably, as a matter of fact.

cschramm:
I'm fine with adding it. 👍

Rudd-O:
Actually, I just tried to code a solution that uses `g_spawn_async` as requested, and now I just realized that the point of having the thread is to supervise the process ending.

I was misled by your earlier comments thinking that `g_spawn_async` would do that for me — both run a background thread to reap the child process, and emit a signal when the process was done — but it clearly doesn't as per the documentation, so I still end up needing a damn thread.  In other words — it's not that I'm "adding *another* threading system" here, rather I am adding *a* thread here, which is necessary to supervise and reap the end of the called process.  That's **necessary**, rather than a frill.

`g_spawn_async` is simply the wrong solution here.  The code is fine as it is.  I shouldn't have second-guessed myself.  All I will do is add startup notification code, for those cases when the code actually runs a GUI program that supports desktop startup notifications.

You will receive a push with updated code very soon.

infirit:
No, it is more you not understanding what are the potential issues with combining python threading with glib mainloop/threading. last attempt with an example and you can explain what part does not work for the feature.

```python
from gi.repository import GLib

def on_child_exit(pid, result):
    print("Closing child with pid %s"  % pid)
    if result != 0:
        print("Process ended with error status %s" % result)
    GLib.spawn_close_pid(pid)

command = ["/usr/bin/sleep", "-10"]

try:
    pid, stdin, stdout, stderr = GLib.spawn_async(
        command, flags=GLib.SPAWN_DO_NOT_REAP_CHILD)
    GLib.child_watch_add(pid, on_child_exit)
except GLib.GError as e:
    print(e.message, e.code, dir(e))
    raise

GLib.MainLoop().run()
```

Rudd-O:
1. There's exactly zero problem — and it is a fully supported use case — to use Python threading and GLib mainloop simultaneously, *so long as the threaded code never touches any shared data structures*.  GLib.idle_add is 100% thread safe and can be used in the manner that my code is using it.  I have been doing GTK programming since *version 1*.  This thing that I have been doing, I have been doing it for fifteen years.
2. If you really want to, I will laboriously remove the use of Python threading altogether.  I could not find any documentation on something like `child_watch_add`.  But this sort of thing you're asking me will make maintenance of the code *harder*, because the code will be *harder* to read as you've outlined above (it's nonlinear).  If you are *fine* with that burden, I will go right ahead and do it.

Meanwhile, right now I'm refining the code to be more robust.  I await a response from you.

infirit:
>so long as the threaded code never touches any shared data structures.

Exactly, now it doesn't but it may at some point. I do not want to set a precedent and have people fall into the trap.

> Before, I could not find any documentation on something like child_watch_add

I gave you this exact example 25days ago, https://github.com/blueman-project/blueman/pull/632#discussion_r95095811 and have not seen any argument what was wrong with it.

I am done with this discussion, change it or not, I stopped caring after your last comment.

Rudd-O:
I was busy refactoring the code and making the user interface nicer.  It's done, and I'm using the GLib spawn functions as per your desire.

We lose proper Python exceptions by not using subprocess, but for the purposes of displaying to the user what will happen, it appears to be adequate in any case.

I tested the majority of the code on both Python 3 and Python 2.  It seems to be okay.

Rudd-O:
Kind ping?

Rudd-O:
Kind ping.

Rudd-O:
OK.  I guess file transfers will continue to be broken in `blueman`, or hopefully fixed by some other PR (that'd be the better outcome).  Feel free to reuse whatever code you can use, under whatever license is appropriate, from this PR.  :hugs:

PR #636, opened 2017-01-03
bluez.Base: print all changed properties in one go

infirit:
Be a little less verbose and not spam logs so much which improves
readability.


cschramm:
Obviously conflicts with the removal of `dprint`, but the logic is fine. 👍

Issue #610, opened 2016-11-01
GtkStatusIcon removed from Gtk+


infirit:
GtkStatusIcon was [removed](https://github.com/GNOME/gtk/commit/d2a8667f835ddce4ac88b9bee143556f9efa93c4) in Gtk+ master in September 2016 ~~so very likely it will be gone in Gtk+ 3.24. We need to plan what to do as we rely on it.~~ The new release will likely be Gtk4 not 3 anymore. But we'll have to prepare anyway.

* Use Appindicator plugin as the default?
* Which Gtk+ version will we target for 2.1?
* More things we need to look at?

Using appinficator we loose [tooltip](https://bugs.launchpad.net/indicator-application/+bug/527458) and opening/closing blueman-manager as left click is not under our control.

cschramm:
We have #306 lying around. I didn't really have a look at the GNotification stuff and what's necessary for that.

> Which Gtk+ version will we target for 2.1?

Well, we'll definitely need to support newer versions. Using GNotification would raise the minimum GLib version to 2.39. Do you see other reasons to raise the minimum version?

infirit:
I edited the original text, the new version will likely be 4 and 3 will be the left alone. See https://blogs.gnome.org/desrt/2016/06/13/gtk-4-0-is-not-gtk-4/ and https://wiki.gnome.org/Hackfests/DeveloperExperience2014/GtkRoadmap. But we still need to fix this eventually though.

Using GNotification still leaves us without an icon under __all__ desktops. It is not meant to replace the trayicon. It is similar to the current notification spec but with extra feature like persistence across sessions.

>raise the minimum GLib version to 2.39

The issues is with the slack users, everyone else is most likely already on 2.39 when 2.1 is released.

cschramm:
> I edited the original text, the new version will likely be 4 and 3 will be the left alone.

Right, it'll be 3.90 which will be called `gtk+-4.0` in pkg-config. So GtkStatusIcon won't get dropped in GTK+ 3 and as 2.24 is still receiving patches I don't see a reason to migrate away from it any time soon. :smiley: 

> Using GNotification still leaves us without an icon under all desktops.

Alright, rifled through gnome-shell and now starting to understand the situation: GNotification does not provide any indicator icon functionality. GNOME still has indicators as provided by GtkStatusIcon, but they are implemented in GNOME Shell internally (with Gjs and [the internal St module](https://developer.gnome.org/st/stable/index.html)).

[...]

PR #637, opened 2017-01-04
Tray application

cschramm:
Making my work on #610 visible. Far from done.

### DIFF
--- a/blueman/main/Tray.py
+++ b/blueman/main/Tray.py
@@ -1,0 +1,51 @@
+from importlib import import_module
+
+import logging
+
+from blueman.Functions import check_single_instance
+from blueman.main.AppletService import AppletService
+
+import gi
+gi.require_version('Gtk', '3.0')
+from gi.repository import GLib, Gio
+
+
+class BluemanTray(object):
+    def __init__(self):
+        check_single_instance("blueman-tray")
+
+        applet = AppletService()
+
+        main_loop = GLib.MainLoop()
+
+        Gio.bus_watch_name(Gio.BusType.SESSION, 'org.blueman.Applet',
+                           Gio.BusNameWatcherFlags.NONE, None, lambda _connection, _name: main_loop.quit())
+
+        indicator_name = applet.GetStatusIconImplementation()
+        logging.info('Using indicator "%s"' % indicator_name)
+        indicator_class = getattr(import_module('blueman.main.indicators.' + indicator_name), indicator_name)
+        self.indicator = indicator_class(applet.GetIconName(), self._activate_menu_item, self._activate_status_icon)
+
+        applet.connect('g-signal', self.on_signal)
+
+        self.indicator.set_text(applet.GetText())
+        self.indicator.set_visibility(applet.GetVisibility())
+        self.indicator.set_menu(applet.GetMenu())
+
+        main_loop.run()
+
+    def _activate_menu_item(self, *indexes):
+        return AppletService().ActivateMenuItem('(ai)', indexes)
+
+    def _activate_status_icon(self):
+        return AppletService().Activate()
+
+    def on_signal(self, _applet, sender_name, signal_name, args):
+        if signal_name == 'IconNameChanged':
+            self.indicator.set_icon(*args)
+        elif signal_name == 'TextChanged':
+            self.indicator.set_text(*args)
+        elif signal_name == 'VisibilityChanged':
+            self.indicator.set_visibility(*args)
+        elif signal_name == 'MenuChanged':
+            self.indicator.set_menu(*args)
--- a/blueman/main/indicators/AppIndicator.py
+++ b/blueman/main/indicators/AppIndicator.py
@@ -1,0 +1,34 @@
+# coding=utf-8
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import gi
+
+gi.require_version('AppIndicator3', '0.1')
+from gi.repository import AppIndicator3
+from blueman.main.indicators.GtkStatusIcon import build_menu
+
+
+class AppIndicator(object):
+    def __init__(self, icon_name, on_activate_menu_item, _on_activate_status_icon):
+        self._on_activate = on_activate_menu_item
+        self.indicator = AppIndicator3.Indicator.new('blueman', icon_name,
+                                                     AppIndicator3.IndicatorCategory.APPLICATION_STATUS)
+        self.indicator.set_status(AppIndicator3.IndicatorStatus.ACTIVE)
+
+    def set_icon(self, icon_name):
+        self.indicator.set_icon(icon_name)
+        self.indicator.set_status(AppIndicator3.IndicatorStatus.ATTENTION)
+        self.indicator.set_status(AppIndicator3.IndicatorStatus.ACTIVE)
+
+    def set_text(self, text):
+        self.indicator.props.title = text
+
+    def set_visibility(self, visible):
+        status = AppIndicator3.IndicatorStatus.ACTIVE if visible else AppIndicator3.IndicatorStatus.PASSIVE
+        self.indicator.set_status(status)
+
+    def set_menu(self, menu):
+        self.indicator.set_menu(build_menu(menu, self._on_activate))
--- a/blueman/plugins/applet/DiscvManager.py
+++ b/blueman/plugins/applet/DiscvManager.py
@@ -27,20 +27,18 @@ class DiscvManager(AppletPlugin):
     }
 
     def on_load(self, applet):
-        self.item = create_menuitem(_("_Make Discoverable"), "edit-find")
-        applet.Plugins.Menu.Register(self, self.item, 20, False)
+        self.item = applet.Plugins.Menu.add(self, 20, text=_("_Make Discoverable"), icon_name="edit-find",
+                                            tooltip=_("Make the default adapter temporarily visible"),
+                                            callback=self.on_set_discoverable, visible=False)
 
         self.Applet = applet
         self.adapter = None
         self.time_left = -1
 
-        self.item.connect("activate", self.on_set_discoverable)
-        self.item.props.tooltip_text = _("Make the default adapter temporarily visible")
-
         self.timeout = None
 
     def on_unload(self):
-        self.Applet.Plugins.Menu.Unregister(self)
+        self.Applet.Plugins.Menu.unregister(self)
         del self.item
 
         if self.timeout:
@@ -56,12 +54,12 @@ class DiscvManager(AppletPlugin):
 
     def on_update(self):
         self.time_left -= 1
-        self.item.set_label(_("Discoverable... %ss") % self.time_left)
-        self.item.props.sensitive = False
+        self.item.set_text(_("Discoverable... %ss") % self.time_left)
+        self.item.set_sensitive(False)
 
         return True
 
-    def on_set_discoverable(self, item):
+    def on_set_discoverable(self):
         if self.adapter:
             self.adapter.set("Discoverable", True)
             self.adapter.set("DiscoverableTimeout", self.get_option("time"))
@@ -111,10 +109,10 @@ class DiscvManager(AppletPlugin):
     def update_menuitems(self):
         if self.adapter is None:
             logging.warning("warning: Adapter is None")
-            self.item.props.visible = False
+            self.item.set_visible(False)
         elif (not self.adapter["Discoverable"] or self.adapter["DiscoverableTimeout"] > 0) and self.adapter["Powered"]:
-            self.item.props.visible = True
-            self.item.set_label(_("_Make Discoverable"))
-            self.item.props.sensitive = True
+            self.item.set_visible(True)
+            self.item.set_text("_Make Discoverable")
+            self.item.set_sensitive(True)
         else:
-            self.item.props.visible = False
+            self.item.set_visible(False)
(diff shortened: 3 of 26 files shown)


blueman/main/Tray.py
```diff
@@ -0,0 +1,44 @@
+from importlib import import_module
```
infirit: importlib was added in python 3.1, I'm ok with dropping python 2 support 👍

blueman/main/Tray.py
```diff
@@ -0,0 +1,44 @@
...
+import gi
+gi.require_version('Gtk', '3.0')
+from gi.repository import Gtk
+
+
+class BluemanTray(object):
+    def __init__(self):
+        check_single_instance("blueman-tray")
+
+        logging.basicConfig(level=logging.INFO)
```
infirit: This should eventually (when logging stuff is merged) be moved to app/blueman-tray and add the arguments needed for setting log level.

blueman/main/indicators/AppIndicator.py
```diff
@@ -0,0 +1,33 @@
+# coding=utf-8
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import gi
+
+gi.require_version('AppIndicator3', '0.1')
```
infirit: This will cause an exception if the typelib is not installed, if you are expecting this ignore this. If not see the current Appindicator plugin how I solved it.

blueman/plugins/applet/DiscvManager.py
```diff
@@ -31,21 +31,18 @@ class DiscvManager(AppletPlugin):
...
         self.time_left = -1
 
-        self.item.connect("activate", self.on_set_discoverable)
-        self.item.props.tooltip_text = _("Make the default adapter temporarily visible")
-
         self.timeout = None
 
     def on_unload(self):
-        self.Applet.Plugins.Menu.Unregister(self)
+        self.Applet.Plugins.Menu.unregister(self)
```
infirit: Adding to the menu was changed from ``Register`` to ``add``, do the same here and use ``remove``?

blueman/plugins/applet/StandardItems.py
```diff
@@ -24,53 +24,34 @@ class StandardItems(AppletPlugin):
...
+        self.devices = self.Applet.Plugins.Menu.add(self, 60, text=_("_Devices") + "...", icon_name="blueman",
+                                                    callback=self.on_devices)
 
-        self.Applet.Plugins.Menu.Register(self, Gtk.SeparatorMenuItem(), 51)
+        self.adapters = self.Applet.Plugins.Menu.add(self, 70, text=_("Adap_ters") + "...", icon_name="blueman-device",
+                                                     callback=self.on_adapters)
 
-        self.devices = create_menuitem(_("_Devices") + "...", get_icon("blueman", 16))
-        self.devices.connect("activate", self.on_devices)
+        self.Applet.Plugins.Menu.add(self, 80, text=_("_Local Services") + "...", icon_name="preferences-desktop",
```
infirit: We should probably use the Unicode ellipsis instead of three dots. Example ``…``. It was like that before but if you are changing things anyway 😄

cschramm:
Submenus (recent connections), activation (opening blueman-manager), and exit (closing blueman-tray when the applet service becomes available) work now and the tray itself no longer depends on GTK+. The GtkStatusIcon indicator of course depends on GTK+ 3 and the AppIndicator indicator does as well as it uses shared code but it should be easy to extract that and use GTK+ 4 for the AppIndicator menu instead.

I'll go through your comments next.

A major issue is that the AppIndicator implementation does not work. The icon just isn't displayed for me (and the `connected` property stays `False`). I do see `Gdk-CRITICAL **: gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0' failed` when calling `set_menu` but I see that when using the AppIndicator plugin in blueman-applet in master as well and the icon *does* work there, so that does not seem to be related.

infirit:
yup, the critical has been there for a while. I will try this later today and see if we can figure out the missing icon.
-- 

infirit:
Couple observations:

If I run ``blueman-tray`` with ``blueman-applet`` not running I only have a single entry in the menu (Turn bluetooth on/off). When the applet is already running the menu is fine.

I always get the statusicon even when I enabled the appindicator plugin under xfce4.

blueman/main/Tray.py
```diff
@@ -0,0 +1,54 @@
...
+
+        main_loop.run()
+
+    def _activate_menu_item(self, *indexes):
+        return AppletService().ActivateMenuItem('(ai)', indexes)
+
+    def _activate_status_icon(self):
+        return AppletService().Activate()
+
+    def on_signal(self, _applet, sender_name, signal_name, args):
```
infirit: args here is a GVariant which has to be unpacked first. So my sugestion would be

```def on_signal(self, _applet, sender_name, signal_name, params):
    args = params.unpack()
    ....
```

blueman/main/Tray.py
```diff
@@ -0,0 +1,54 @@
...
+        if signal_name == 'IconNameChanged':
+            self.indicator.set_icon(*args)
+        elif signal_name == 'TextChanged':
+            self.indicator.set_text(*args)
+        elif signal_name == 'VisibilityChanged':
+            self.indicator.set_visibility(*args)
+        elif signal_name == 'MenuChanged':
+            self.indicator.set_menu(*args)
```
infirit: The reason the exititem is not changed is because the signal MenuChanged is never receieved here. I do see others like ``VisibilityChanged`` and ``VisibilityChanged`` so it appears to be on the dbus service side.

blueman/main/Tray.py
```diff
@@ -0,0 +1,54 @@
...
+
+        applet = AppletService()
+
+        main_loop = GLib.MainLoop()
+
+        Gio.bus_watch_name(Gio.BusType.SESSION, 'org.blueman.Applet',
+                           Gio.BusNameWatcherFlags.NONE, None, main_loop.quit)
+
+        indicator_name = applet.GetStatusIconImplementation()
+        indicator_name = 'GtkStatusIcon'
```
infirit: Here is why I always get the status icon 😄

blueman/main/Tray.py
infirit: now that logging has been merged this should be dropped.

blueman/plugins/applet/Menu.py
```diff
@@ -24,50 +100,52 @@ class Menu(AppletPlugin):
...
 
     def on_plugins_loaded(self):
         self.__plugins_loaded = True
         self.__sort()
-        self.__load_items()
 
-    def get_menu(self):
-        return self.__menu
+    def on_menu_changed(self):
+        self.MenuChanged(self.GetMenu())
```
infirit: Why send the whole menu? Would be simpler if the the client (Tray) just calls ``AppletService.GetMenu`` instead?

blueman/plugins/applet/Menu.py
```diff
@@ -4,18 +4,94 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-from blueman.Functions import *
+import dbus
```
infirit: ```import dbus.service``` still not signal after this though, functions are being called properly just no signal emission 😞

infirit: I checked with ``gdbus monitor -e -d org.blueman.Applet`` and I do see it.

```
The name org.blueman.Applet is owned by :1.885
/: org.blueman.Applet.MenuChanged ([{'tooltip': <'Laat netwerkverkeergebruik zien'>, 'markup': <false>, 'sensitive': <true>, 'text': <'Netwerk _gebruik'>, 'icon_name': <'network-wireless'>}],)
```

very confused why it does not show up in ``on_signal``.

cschramm:
Turns out DBus signal stuff was broken all the time. #703.

blueman/main/Tray.py
cschramm: Looks like like `*args` works fine, doesn't it?

blueman/main/Tray.py
cschramm: Ah, sorry, forgot to mention that. 🙈 

You can easily verify that it's the AppIndicator item by left-clicking on it. The GtkStatusIcon should open blueman-manager while AppIndicator would probably show the menu. Also, blueman-tray tells you which indicator it is using.

blueman/plugins/applet/Menu.py
cschramm: Well, it will need the menu anyway when it gets that signal.

blueman/main/Tray.py
cschramm: Hm. Shall we? Probably makes sense.

blueman/main/Tray.py
infirit: I know the difference, the problem is forcing ``indicator_name`` to be ``GtkStatusIcon``. Line 27 should be dropped 😸.

blueman/plugins/applet/Menu.py
infirit: Signals work with your fix 🎉, still we should add ``import dbus.service`` as it contains the decorators, we should not rely on other modules for them 😄

blueman/plugins/applet/Menu.py
infirit: Just my preference here. Either way is fine with me 👍

blueman/main/Tray.py
infirit: Indeed it does, just wondering why though as in previous code I wrote it caused issues. Anyway if it works 👍

blueman/main/Tray.py
infirit: If we do then we should add a follow up commit to change configure.ac. We can then also drop the major/minor python version bits.

blueman/plugins/applet/StandardItems.py
```diff
@@ -79,53 +60,53 @@ def change_sensitivity(self, sensitive):
...
+        button.connect("clicked", lambda _button: self.on_plugins())
 
         about.action_area.pack_start(button, True, True, 0)
         about.action_area.reorder_child(button, 0)
 
         about.run()
         about.destroy()
 
-    def on_plugins(self, button):
+    def on_plugins(self):
```
infirit: This broke the dbus method used by manager. We should drop ``None`` from [here](https://github.com/cschramm/blueman/blob/13381dc017e84b8dc816e8732681a5c52180286b/blueman/plugins/applet/DBusService.py#L127)

infirit:
>Gdk-CRITICAL **: gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0'

On this, was experimenting with AppIndicators, I can make this happen with a very simple example like https://gist.github.com/infirit/e05131eb5414b036a51acb8e0af47a9c. It happens whenever it falls back to XEmbed and while annoying its most likely not something we can fix.

cschramm:
Looks mostly fine now on my side. Even the AppIndicator implementation started working for some reason.

One small issue I have is that I'm getting a wait cursor in the Xfce panel for a while after launching blueman-tray. Not sure what `GAppInfo` expects from blueman-tray to stop that...

> If I run ``blueman-tray`` with ``blueman-applet`` not running I only have a single entry in the menu (Turn bluetooth on/off). When the applet is already running the menu is fine.

Didn't try that yet. Might still be an issue.

cschramm:
Running `blueman-tray` with `blueman-applet` not running (invoked via D-Bus) works for me. I could not produce such an incomplete menu you described.

However, there's problems with the AppIndicator implementation again. If it's used the icon only shows up for me if `blueman-tray` gets launched from `blueman-applet`. If I start it manually, nothing shows up. The GtkStatusIcon implementation works fine. 🤔

infirit:
>One small issue I have is that I'm getting a wait cursor in the Xfce panel for a while after launching blueman-tray. Not sure what GAppInfo expects from blueman-tray to stop that...

Its not GAppInfo afaics. I see his happen when the starting the tray directly when blueman-applet is not running. If blueman-applet is already running and I start the tray I see nothing. This makes sense as the tray uses a plain glib mainloop which afaik does not support startup notfication.

I looked at the docs and when using Gtk.main() the first GtkWindow will mark startup complete. My guess is that with the tray in a separate process the applet now never creates a GtkWindow. Couple of things I would (and will this weekend) try to make BluemanApplet use a plain glib mainloop. This hopefully will stop it from using startup notification. Also try to sprinkle ```Gdk.notify_startup_complete()``` in several places to see what is causing this.

>Running blueman-tray with blueman-applet not running (invoked via D-Bus) works for me. I could not produce such an incomplete menu you described.

Good, neither can I now 😄 

>However, there's problems with the AppIndicator implementation again. If it's used the icon only shows up for me if blueman-tray gets launched from blueman-applet.

I can not reproduce this here. Can you let me know your steps and appindicator/gtk version?

cschramm:
Never mind the last one. Looks like there's some issue with showing the AppIndicator icon if `blueman-tray` is run as a child of PyCharm. 😉

So the only issue left is the wait cursor. I think that's a minor one we can work on afterward. Shall we get the current version into master?

infirit:
>So the only issue left is the wait cursor.

I like to add one, it does not notice plugin changes so switching between the two needs a restart.

>Shall we get the current version into master?

Sure, I'll turn it into a G(tk)Application. Also I think it should probably not autostart applet but instead watch for the name to appear (I'll take care of that with G(tk)Application work).

PR #638, opened 2017-01-05
Argparse fixes and fix a bug in blueman-sendto

infirit:
We do not handle when a non existing device is used with bluemna-sendto. ``Manager.find_device`` returns None and when it does exit.


cschramm:
👍

Issue #606, opened 2016-10-29
interface problem


rwst:
Trying to connect with a Samsung Soundbar via the main Setup button will open the Bluetooth Authentication box and then take some time until presenting me with "Device added successfuly(sic) but failed to connect". However, if I after that use the right mouse button on the entry and Setup I quickly get "Device added and connected successfuly". This is the only method I can connect. It seems that the main Setup button, as well as the code I start via the Device->Setup menu entry is missing something which only the right-click combo gets right.

remborg:
I have exactly the same problem with an HK bluetooth speaker on linux mint.
I ran the blueman-applet from the command line and here is what I get when creating the device for the first time (resulting a "Device added successfully but failed to connect")

`__del__ (/usr/lib/python2.7/dist-packages/blueman/main/Device.py:67)
deleting device None 
_________
Destroy (/usr/lib/python2.7/dist-packages/blueman/main/Device.py:89)
invalidating device None 
_________
Run (/usr/lib/python2.7/dist-packages/blueman/main/PluginManager.py:203)
Function on_request_menu_items on Services Failed 
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/blueman/main/PluginManager.py", line 207, in Run
    ret = getattr(inst, function)(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/blueman/plugins/manager/Services.py", line 54, in on_request_menu_items
    add_menu_item(manager_menu, service)
  File "/usr/lib/python2.7/dist-packages/blueman/plugins/manager/Services.py", line 29, in add_menu_item
    if service.connected:
  File "/usr/lib/python2.7/dist-packages/blueman/Service.py", line 54, in connected
    return self._legacy_interface.GetProperties()['Connected']
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 70, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
DBusException: org.freedesktop.DBus.Error.UnknownMethod: Method "GetProperties" with signature "" on interface "org.bluez.Handsfree" doesn't exist

_________
Generate (/usr/lib/python2.7/dist-packages/blueman/gui/manager/ManagerDeviceMenu.py:165)
[...]

PR #647, opened 2017-01-22
Device selector widget re-implment and cleanups

infirit:
While looking at the code a bit for #606 I did not like what I saw. The ui xml seemed overly complex and could be fairly simple done from the class instead. To me this makes thing much more understandable and readable. No functional changes and I could not spot any but other may.

Another PR will follow later with the actual ui changes.


infirit:
Added some commits for the DeviceSelectorDialog, cleanups and removal of deprecated code.

PR #656, opened 2017-02-02
Add GATT service names

cschramm:


infirit:
Wondering if it is wise to mark (comment or something) these as belonging to GATT. Otherwise 👍 😄

Issue #648, opened 2017-01-22
Logitech MX Anywhere 2 UI glitches


akostadinov:
Hello, I've got Logitech MX Anywhere 2 mouse to hopefully find a good mouse finally. It connects perfectly well so far which is beyond awesome (unlike the lenovo thinkpad bluetooth keyboard which seems to be doing legacy pairing, connects unreliably and is laggy /btw can I somehow force disable legacy pairing?/).

issues I see:

- blueman applet doesn't show a mouse icon
- blueman applet does not show connection strength indicators
- battery power level is not shown through the power manager applet (for the lenovo keyboard it was shown)
- a number of unknown UUIDs in the `info` section

Here is a screenshot: http://pasteboard.co/p8ewixAgO.png

I suspect you will tell me that some of these issues are outside control of blueman. I'll move any of these to other projects if you advise so.

-- blueman: 2.1, BlueZ: 5.43, 4.8.16-200.fc24.x86_64

infirit:
>blueman applet doesn't show a mouse icon

It uses the ``Class`` property for this which is empty for you. I think we may want to use the ``Icon`` property as a fallback. @cschramm, opinions, sugestions? Drop icons based on Class entirely and solely use ``Icon``?

>blueman applet does not show connection strength indicators

Can you post the console output from ``blueman-manager`` it should have useful info.

>battery power level is not shown through the power manager applet (for the lenovo keyboard it was shown)

That is out of our control, probably best to report this to the people from the power manager applet.

> a number of unknown UUIDs in the info section

These are LE uuids not part of SDP from what I could quickly find. We will need to handle this at some point but as long as it does not crash I'm ok with this for now. Also reff  #625

akostadinov:
It is very interesting. The icon of blueman applet shows green as if device is connected. But if I right click on manager over the mouse, it doesn't show the option to disconnect as if device is not connected. But the mouse is in fact connected and working. If I turn it off through button, then blueman applet icon shows as if no devices are connected. So to me it seems that the manager app doesn't recognize the mouse is connected presently. Also doesn't show any "Connect to" options with right-click. 

This log was taken like:

1. turn off mouse
2. start manager in console
3. turn on mouse
4. right click on mouse in manager

The mouse just starts to work so I didn't notice manager didn't show "Connect" option. For some reason I can't take a screenshot though. It is just "Info" at the top, then grayed out "Send FIle", etc.

[...]

PR #658, opened 2017-02-09
ManagerDeviceList: Add fallback for device icon

infirit:
See https://github.com/blueman-project/blueman/issues/648#issuecomment-277299335


infirit:
The Mi Bluetooth Speaker and SL910 were blueman icons before.

screenshot_2017-02-09_12-35-11

PR #659, opened 2017-02-11
Fix: tray icon whithout icons with right click on mate desktop

sc0w:
revert https://github.com/blueman-project/blueman/commit/762fb161360031a28b93722f74549d7aa12a8089

Fixes https://github.com/mate-desktop/mate-panel/issues/521

### DIFF
--- a/blueman/plugins/applet/AppIndicator.py
+++ b/blueman/plugins/applet/AppIndicator.py
@@ -23,7 +23,7 @@ class AppIndicator(AppletPlugin):
     def on_load(self, applet):
         
         self.indicator = girAppIndicator.Indicator.new("blueman",
-                                                       self.Applet.Plugins.StatusIcon.props.icon_name,
+                                                       applet.Plugins.StatusIcon.get_option("icon"),
                                                        girAppIndicator.IndicatorCategory.APPLICATION_STATUS)
         
         self.indicator.set_status(girAppIndicator.IndicatorStatus.ACTIVE)


infirit:
This is very wrong, did you even test this properly? And it only "fixes" the issues by failing to load the AppIndicator plugin. Check the output for blueman-applet (run from a terminal).

infirit:
Or check the plugins dialog and try to enable AppIndicator plugin.

sc0w:
weird, If I unchecked the AppIndicator plugin, it works, and if I checked it and restart session it doesn't work

infirit:
Yes, because the StatusIcon takes over from AppIndicator. The problem is with the libdbusmenu as I described in the other issue.

sc0w:
but, with blueman 1.99~alpha1 it works with AppIndicator

infirit:
Correct, all the details are in the other issue. But it comes down to Gtk deprecating GtkImageMenuItem, projects try to workaround the deprecation as suggested by the Gtk api docs (Blueman release 2.0). But libdbusmenu has not caught up with the suggestion and can not find the icon, just the label. Now that libdbusmenu caught up (release 16.04) things are back to normal as I reported on the other issue.

infirit:
btw you do realize that the new notification area for mate-panel now uses LibAppIndicator instead if the older GtkStatusIcon (now also deprecated). Before it would show our StatusIcon plugin but now it uses the AppIndicator if enabled. This is why you only noticed this issues after the new notification area was merged, it has been like that on ubuntu desktop for a while now.

sc0w:
thanks for the clarifications :)

PR #686, opened 2017-04-30
Implement a ServiceUUID class

cschramm:
This is @infirit's idea from #648 to fix an inaccuracy in our service class handling where we just took the last 16 bits of the time_low field as a 16-bit short UUID without checking the full UUID is actually a reserved one. The consequence was that we did not differentiate between unknown reserved service classes and prorietary ones and especially that we could possibly confuse proprietary ones with known reserved ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * blueman-adapters is now (Xfce-)pluggable
 
+### Changes
+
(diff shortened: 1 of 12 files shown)


infirit:
Have not tested but looks good to me 👍 . Nice use of the UUID module 😄

PR #687, opened 2017-05-02
invoke _NotificationDialog.callback with 1 argument

dakkar:
The elements in `_NotificationBubble._callbacks` are invoked with a single argument (the action name), but `_NotificationDialog.callback` was invoked with two arguments: the `_NotificationDialog` instance, and the action name.

Some places expected their callbacks to be invoked one way, some the other.
I've normalised everything to always expect a single argument.


cschramm:
Thank you! 😃

PR #689, opened 2017-05-06
WIP: KillSwitch: Add support for connman

infirit:
**Untested** (tested now, works ok so far) work in progress but this is what I am thinking of to add connmans support.

>Connman blocks the rfkill device for writing and we need to use it to
>set the power state.

### DIFF
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -6,12 +6,11 @@ from __future__ import unicode_literals
 
 import os
 import weakref
-from gi.repository import GLib
+from gi.repository import GLib, Gio
 import struct
 import logging
 
 from blueman.main.Mechanism import Mechanism
-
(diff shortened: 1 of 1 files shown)


blueman/plugins/applet/KillSwitch.py
```diff
@@ -126,7 +149,13 @@ def reply(*_):
         def error(*_):
             cb(False)
 
-        Mechanism().SetRfkillState(str('(b)'), state, result_handler=reply, error_handler=error)
+        if self._have_connman:
```
cschramm: You could just check for `self._connman_proxy` and save that other variable.

progandy:
That seems to work fine.

blueman/plugins/applet/KillSwitch.py
infirit: Good one, will do that before merging 👍

infirit:
Thanks for testing @progandy, merged and pushed to master.

Issue #688, opened 2017-05-03
RFCOMM channel hard coded to 1


pideu-mh:
blueman: 2.0.4
BlueZ: 5.37
Distribution: Ubuntu 16.04.2 LTS
Desktop environment: Xubuntu

I am using a CSR8510 based Bluetooth dongle which is basically working fine.

Now I'd like to connect to a remote Bluetooth Classic device that uses RFCOMM on channel number 5.

Unfortunately this is not working correctly, a connection can never be established.

The root cause of the problem can be found in the _connect()_ method here:
[https://github.com/blueman-project/blueman/blob/master/blueman/services/meta/SerialService.py](url)
```
# TODO: Channel?
port_id = create_rfcomm_device(Adapter(self.device["Adapter"])['Address'], self.device["Address"], 1)
```

The comment already indicates that the author was well aware that hard coding the channel number to 1 is probably going to be a problem.

Anyway, in my local installation of _blueman_ I changed the 1 to 5 and restarted the _blueman-applet_. Afterwards I was able to connect to my remote Bluetooth Classic device that uses RFCOMM on channel number 5.

I have no idea of the inner workings of _blueman_, but my suggestion is to properly implement the _service discovery protocol_ (SDP) handling in order to find out the channel number the RFCOMM service on the remote device is using.

On the command line, that handling can be triggered using _sdptool_
```
$ sdptool records aa:bb:cc:dd:ee:ff
Service Name: Serial Port
Service RecHandle: 0x10000
Service Class ID List:
  "Serial Port" (0x1101)
Protocol Descriptor List:
  "L2CAP" (0x0100)
  "RFCOMM" (0x0003)
    Channel: 5
```

infirit:
>I have no idea of the inner workings of blueman

The low level bluetooth code is in https://github.com/blueman-project/blueman/blob/master/module/libblueman.c which needs some love to extract the channel. Then it is wrapped into a module for python with cython https://github.com/blueman-project/blueman/blob/master/module/_blueman.pyx.in

pideu-mh:
Thank you for your comment. 

Hm, in the files you mentioned the handling for the RFCOMM channel seems to be implemented correctly already.

In _create_rfcomm_device()_ a "channel" parameter can already be provided and in _def rfcomm_list():_ channel is appended as well.

It seems that the problem is in the higher layers instead.

[...]

PR #694, opened 2017-05-10
Use correct RFCOMM channel number by retrieving the RFCOMM channel nu…

pideu-mh:
…mber from RFCOMM list

This is a change proposal in order to fix the issue addressed in https://github.com/blueman-project/blueman/issues/688

As already said, from my perspective it looks like everything is already there.

Please check out _rfcomm_list()_ from __blueman.pyx.in_. It generates a list of known devices and adds the address, state, flags *and* the channel to that list.

It seems that in _SerialService.py_ this information is simply not extracted at all.

Please check out this patch. I added some code to the _connect()_ function of _SerialService.py_. 

It will use _rfcomm_list()_ to retrieve the list of devices and then iterate over it to find the entry matching this connection attempt. It will then use the channel number that was stored for that rfcomm entry when the RFCOMM channel is finally set up.

Probably this is not the cleanest possible solution. 

On the one hand the list may be large and on the other hand the other information for the connection attempt by _create_rfcomm_device_ comes from _self.device[]_. So maybe the channel information should be stored as _self.device["Channel"]_ instead. But this is out of my scope to accomplish.

Anyway, with this change I can successfully connect to my RFCOMM device which uses the RFCOMM channel 5. The channel number is successfully retrieved using _rfcomm_list()_.

### DIFF
--- a/blueman/services/meta/SerialService.py
+++ b/blueman/services/meta/SerialService.py
@@ -51,8 +51,11 @@ class SerialService(Service):
 
     def connect(self, reply_handler=None, error_handler=None):
         try:
-            # TODO: Channel?
-            port_id = create_rfcomm_device(Adapter(self.device["Adapter"])['Address'], self.device["Address"], 1)
+            channel = 1
+            for dev in rfcomm_list():
+                if dev["dst"] == self.device['Address']:
+                    channel = dev["channel"]
+            port_id = create_rfcomm_device(Adapter(self.device["Adapter"])['Address'], self.device["Address"], channel)
             filename = '/dev/rfcomm%d' % port_id
             logging.info('Starting rfcomm watcher as root')
             Mechanism().open_rfcomm(str('(d)'), port_id)


infirit:
>So maybe the channel information should be stored as self.device["Channel"] instead. But this is out of my scope to accomplish.

``self.device`` is a [DBus Proxy object](https://github.com/blueman-project/blueman/blob/master/blueman/bluez/Device.py) that proxies the bluez dbus interface ``org.bluez.Device1``. We added dict like symantics to set/get properties which we could extend it I would prefer to leave it as a pure dbus proxy and keep it cleanly separate :smile:.

The commit subject is a bit long so I would suggest something like ``SerialService: use correct RFCOMM channel to connect``

Other than the commit subject this looks ok to me, @cschramm may have some things.

cschramm:
> The comment already indicates that the author was well aware that hard coding the channel number to 1 is probably going to be a problem.

Yes, I was aware this was going to bite me... 😄

Actually I'm not sure why this solution works for you.

I have a device that provides a serial service with channel 1. On first invocation, `rfcomm_list()` returns an empty list for me as there are no RFCOMM device files yet, so `SerialService.connect` still uses channel 1. This works for me but you said it does not work for you. If I try to connect a second time (first connection is still alive) I do find the first connection in `rfcomm_list()` as there is a device file for it.

If `rfcomm_list()` is non-empty for you even during the very first invocation, there must already be an RFCOMM device file (not created by blueman). 🤔

pideu-mh:
I understand what you mean and I agree with you.

I restarted the machine and added some debug prints and the returned list is empty for me as well now. So it seems that due to my previous experiments there were left-over RFCOMM device files, due to whatever reasons.

I understand now that this functionality is only there to retrieve the list of devices *after* a connection has been successfully established before.

So let's scrap that idea. :-) 

I am wondering: in the dialog as implemented by _blueman-assistant.in_ it will display all remote devices by their Bluetooth device address, sometimes with their name and later has the information which profiles are implemented. 

I have not yet understood how the service discovery procedure is triggered, who is then doing that service discovery procedure, how the information about the found devices then reaches _blueman_ exactly and why the channel number does not seem to be part of that information.

cschramm:
-> #688

PR #695, opened 2017-05-10
Fix name for auto connect profiles

cschramm:
RecentConns uses the name in the submenu and notifications and currently shows "Proprietary".


infirit:
I don't like the long name but can't think of a better one so :+1:

PR #700, opened 2017-05-14
ManagerDeviceMenu: Always update row with Connected property

infirit:
If we fail to get connection info, which is the case currently with LE
devices, we never update the row connected item. Our ManagerDeviceMenu
is not able to correctly show a Disconnect menu item.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -271,6 +271,10 @@ class ManagerDeviceList(DeviceList):
             self.row_update_event(tree_iter, "Paired", device['Paired'])
         except Exception as e:
             logging.exception(e)
+        try:
+            self.row_update_event(tree_iter, "Connected", device["Connected"])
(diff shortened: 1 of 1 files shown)


blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -271,6 +271,10 @@ def on_finished(fader):
             self.row_update_event(tree_iter, "Paired", device['Paired'])
         except Exception as e:
             logging.exception(e)
+        try:
+            self.row_update_event(tree_iter, "Connected", device["Connected"])
+        except Exception as e:
```
cschramm: What are you catching here? 🤔

infirit: I just kept it the same as the other two, Trusted and Paired. But to be honest, no idea why they are there. These were blanket exceptions before and in the logging commit I at least have the logging module print the exception so we can figure out why they are there.

I can drop them if you like.

PR #703, opened 2017-05-19
Fix DBus signal emission

cschramm:
As DbusService's self.locations returns an iterator, signal emission in instance works only once and no more locations to send to are found after that.


infirit:
Oops, we never noticed 😲. Tray works fine now so 👍

PR #708, opened 2017-05-31
Small Adapters tab rework

infirit:
I was playing around with Gtk.Application and GAction in blueman-adapters and noticed we pretty much retrieve all the objects from the glade ui xml. Plus that the tab is very simple so I reimplemented it in python.

I'll test later this week or tomorrow (too late now) but putting up already for discussion or comments.


infirit:
Commit https://github.com/blueman-project/blueman/pull/708/commits/6fef6cc20dc5aab84acb611aae7134387369240f is wrong and I will re add the ellipsize and add a tooltip if alias of the adapter exceeds 10 characters with full name. But as I said too late here time for bed 😴

infirit:
OK fixed my silliness from last night and tested things. Works nicely here.

infirit:
I was not happy with how the tab width was set and redid this. It now requests a width which appears to at least have "normal" tabs. Also reworked some more things and fixed some bugs.

We still fail to set the format-value on the GtkScale properly as we never change its value to 0 (zero). I have added a FIXME as I have not been able to get this to work properly.

infirit:
Closing for now, not happy with the page label changes. Will open a new one soon 😄

PR #709, opened 2017-06-02
Discussion: Have a header in the assistant?

infirit:
There is no header any more in Gtk3 however we can replicate one as part of the page. It will require a little work to convert boxes to grids but not too major.

This commit drops the now completely useless setting of images to the header.


cschramm:
I think it would not be that useful. Do you?

infirit:
The current design is not like in gtk2. I have no issue with the current look and feel and have no strong preference for the old look. If you are fine with how it looks now then so am I :)

I'll merge this in the next couple of days.
-- 

PR #712, opened 2017-06-10
icon ui work

infirit:
We had some inconsistencies in the toolbar in particular wrt to icon sizes. This PR added the necessary icons and lets Gtk handle the icon loading for the menus and toolbars.

Along the way one encounters lots of small things to fix like unused imports, dead code etc which are also in this PR.


infirit:
One drawback with this is that certain icon themes do not provide the correct sizes for icons. And we end up with some larger icons. I'll try to figure something out.

infirit:
We can force the pixel size, I'll add this soon for the menus and toolbars etc.

infirit:
Done setting pixel size everywhere. Next is using cairo surfaces to create the device icon but that is for another pull request.

infirit:
Also added some deprecated properties work (I was in the ui files anyway) and a fix for a separator in the network services ui.

blueman/Functions.py
```diff
@@ -244,8 +247,8 @@ def create_menuitem_box(text, pixbuf, orientation=Gtk.Orientation.HORIZONTAL, si
     return item_box
 
 
-def create_menuitem(text, pixbuf):
-    box = create_menuitem_box(text, pixbuf)
+def create_menuitem(text, icon_name=None, pixbuf=None):
```
cschramm: I don't see how this can work for the cases in Services.py that provide a pixbuf as second argument. It would be treated as the `icon_name`.

cschramm:
That all makes sense to me, I just don't see how the `get_x_icon` is supposed to work now.

blueman/Functions.py
infirit: Thanks, I'll fix that by using a keyword argument.

infirit:
Fixed Services.py and found some more things. 

>I just don't see how the get_x_icon is supposed to work now.

I am not sure what you mean, are you referring to Services.py?

PR #719, opened 2017-07-02
(Manager)DeviceList scaling work

infirit:
Posting for interested people. The goal here is to prepare to use cairo surfaces to paint the trusted and paired icons over the device icon. Then when the cairo surface device scaling functions are in a released version of pycairo we can have nice crisp device icons when scaling. The GdkCellRenderePixbuf has a surface property so when I figure out how to use that we can drop all pixbufs.

I'll probably push some commits like the clean-ups of fake device and unused import in the next week.

ps: I have a silly example that uses a patched pycairo that shows how this is done.
https://gist.github.com/infirit/22b147bcdcdc1ed4584f1f513fd1855e

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -227,12 +227,17 @@ def format_bytes(size):
     return (ret, suffix)
 
 
-def create_menuitem(text, icon_name=None, pixbuf=None):
+def create_menuitem(text, icon_name=None, pixbuf=None, surface=None):
     image = Gtk.Image(pixel_size=16)
     if icon_name:
         image.set_from_icon_name(icon_name, Gtk.IconSize.MENU)
+    elif surface:
+        image.set_from_surface(surface)
     elif pixbuf:
         image.set_from_pixbuf(pixbuf)
+    else:
+        raise ValueError("At least provide one of, icon name, surface or pixbuf")
+
     item = Gtk.ImageMenuItem(label=text, image=image, use_underline=True)
     child = item.get_child()
     child.set_use_markup(True)
--- a/blueman/gui/Animation.py
+++ b/blueman/gui/Animation.py
@@ -3,23 +3,21 @@ from gi.repository import GLib
 
 
 class Animation:
-    def __init__(self, image, images, rate=1, rev=False):
-        self.pixbuffs = []
+    def __init__(self, icon, icons, rate=1, rev=False):
+        self.icon_names = []
         self.timer = None
         self.current = 0
-        self.image = None
-
-        self.image = image
+        self.icon = icon
         self.rate = 1000 / rate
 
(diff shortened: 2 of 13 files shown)


infirit:
Getting somewhere now 😄. I just noticed the trusted and paired icons are slightly transparent which I have to fix.
manager_gdk_scale_2

Before
manager_gdk_scale_2_before

infirit:
Fixed transparent trusted and trusted icons. The main problem remaining is our icons and pixmaps. The  rssi, tpl, lq and device icons a low quality and can't be scaled properly. Some icon themes like vertex (see below) do provide nice device icons for blueman.

I am no artist but this can be fixed if we had svg version of these. So any artists lurking 😉.

image

infirit:
I got all of blueman now using icon names or cairo surfaces except for NMDUNSupport plugin. I have no means of testing this. The risk is that there is no icon to disconnect.

infirit:
Rebased against master.

infirit:
Rebased against master

cschramm:
Seems like those icons look much cleaner now. Shall we merge this already?

infirit:
Sure, the last thing I had on my list of problems is the device icons blueman provides. However this is not something we can fix right now. I'll merge the GtkImageMenuItem bits first then rebase and make sure things are still working properly.

infirit:
Rebased and things work as they should.

cschramm:
Unfortunately https://github.com/blueman-project/blueman/pull/719/commits/8ba486598c3be34b3b49857ac833a6d4d04c2f67 does not work on my system.
I get a lot of `AttributeError: 'cairo.ImageSurface' object has no attribute 'set_device_scale'`.

Checking the documentation, it seems like cairo_surface_set_device_scale` is new in 1.14. We should mention that in the dependencies. However, I do have 1.14.10, so I do not really see why that happens to me. Maybe the change did not get reflected in my version of pycairo yet... :thinking:

```
libcairo-gobject2:amd64                                     1.14.10-1
libcairo2:amd64                                             1.14.10-1
python3-cairo                                               1.10.0+dfsg-5+b3
python3-gi-cairo                                            3.22.0-2+b1
```

cschramm:
Ah, yeah, that's it. I need 1.14 to have https://github.com/pygobject/pycairo/commit/498dc027a0ab067471392e1de0e472a77453a415...

infirit:
>Ah, yeah, that's it. I need 1.14 to have pygobject/pycairo@498dc02...

Yup but we don't need to set the device scale. I forgot to remove this as get_icon_info now handles setting the scale which sets it properly for us. Dropped it in 98878a3da1e861a658877c7bb6befeec9e11b78c so you we should be good with older pycairo again.

PR #721, opened 2017-07-04
The larger blueman bluetooth icon are horrible

infirit:
The improvement in quality is huge for the large icon sizes. It is less noticeable or equal on the small ones. Size of all icons together went from 486.7kB to 97.2kB 😲


parkerlreed:
Oh my god, so much better. This is the exact reason I kept turning off the indicator plugin. :D :+1:

infirit:
>No idea why they've been so bad.

My guess is that originally they were created from small icon (~32px) then resized larger with some blurring.

>This is the exact reason I kept turning off the indicator plugin

Unintended but pleased to hear this got improved 😸.

PR #728, opened 2017-07-08
Prepatory work for using cairo surfaces for device icons

infirit:
Using dictionaries makes it much easier to add new features. And a nice side effect is that it makes the code much more readable 😨 

At the moment there is no way to add a cairo.Surface to a GtkListStore and I will use a cell data function to retrieve the surface instead.

### DIFF
--- a/blueman/gui/GenericList.py
+++ b/blueman/gui/GenericList.py
@@ -16,31 +16,26 @@ class GenericList(Gtk.TreeView):
         self.ids = {}
         self.columns = {}
 
-        types = []
-        for i in range(len(data)):
-            types.append(data[i][1])
-
-        types = tuple(types)
+        types = [row["type"] for row in data]
 
         self.liststore = Gtk.ListStore(*types)
         self.set_model(self.liststore)
 
-        for i in range(len(data)):
+        for i, row in enumerate(data):
+            self.ids[row["id"]] = i
 
-            self.ids[data[i][0]] = i
+            if "renderer" not in row:
+                continue
 
-            if len(data[i]) == 4 or len(data[i]) == 5:
+            column = Gtk.TreeViewColumn()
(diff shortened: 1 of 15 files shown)


blueman/gui/GenericList.py
```diff
@@ -29,15 +29,15 @@ def _load(self, data):
 
             self.ids[data[i][0]] = i
 
-            if len(data[i]) == 5 or len(data[i]) == 6:
+            if len(data[i]) == 4 or len(data[i]) == 5:
 
-                column = Gtk.TreeViewColumn(data[i][4])
+                column = Gtk.TreeViewColumn(data[i][3])
```
cschramm: Instead of what seems to have been `None` in all cases you're now passing the "renderer_attributes" `dict` as what should be a `Gtk.CellArea`. I wonder how does not cause an error.

blueman/gui/GenericList.py
```diff
@@ -16,31 +16,26 @@ def _load(self, data):
...
+        for i, row in enumerate(data):
+            self.ids[row["id"]] = i
 
-                column = Gtk.TreeViewColumn(data[i][3])
+            if "renderer" not in row:
+                continue
 
-                column.pack_start(data[i][2], True)
-                column.set_attributes(data[i][2], **data[i][3])
+            column = Gtk.TreeViewColumn(row["renderer"])
```
cschramm: This argument still does not make sense to me (you now changed it from "renderer_attrs" to "renderer").

infirit: Lol, the first argument to the Gtk.TreeViewColumn is being used to set the title. I'll just drop the argument as we do not use the title anywhere.

infirit:
Updated and forced pushed the update.

PR #733, opened 2017-07-24
Errordialog

infirit:
We create several error dialogs in blueman some with backtrace selectable (from an expander) and some without. Keeping it WIP as for some reason obex transfers did not work for me during testing and I have no clue why yet 😭. But still make it available so people can comment, suggest things etc.


infirit:
The obex issue was unrelated. Tested the new dialog and it works in all cases.

Also added the url to the issue tracker when a plugin fails to load. The text can use some love in general but here is how it looks now.
afbeelding

infirit:
Rebase against master

cschramm:
Not sure if like that request to submit it. It might not necessarily be a blueman issue but e.g. some misconfigure or broken DBus, polkit, BlueZ etc., missing dependencies etc.

infirit:
>Not sure if like that request to submit it. It might not necessarily be a blueman issue but e.g. some misconfigure or broken DBus, polkit, BlueZ etc., missing dependencies etc.

Agreed. Except for adding the link to the issue tracker this PR does not change any of the messages.

We can (probably should) do a review of all the errors we produce and make them more clear what the user should be doing. I can open an issue for that for 2.2?

PR #735, opened 2017-07-28
many PEP8 fixes

infirit:
This is currently already in the fixes PR but that got too large so split these off to a new PR. No functional changes.

it also drop lots of commented code that either was for debugging or belonged to features not present or used any more.

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 # coding=utf-8
-
 import os
 import sys
 import signal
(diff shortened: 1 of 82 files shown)


infirit:
If there are things that bugs you let's just skip them for now and revisit them later (I promise I will not be offended so don't hold back 😄).

What I'll do is merge the simple things like white-space and comment formatting first. I'll leave the renames of functions and classes out for now.

cschramm:
> I'll leave the renames of functions and classes out for now.

Well, I guess just testing with those changes is the only good way to discover issues...

cschramm:
Shall we get this into 2.1.alpha2?

infirit:
Its very out of sync and rebasing will take some time.

infirit:
See #810

PR #736, opened 2017-07-29
2-0-stbale: Fixes the network services ui

infirit:
Backported from master.

Also, can someone verify using isc-dhcpd works on jessie? its broken for me.

### DIFF
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -166,14 +166,14 @@ class DhcpdHandler(object):
             f.write(subnet)
             f.close()
 
-            cmd = [have("dhcpd3") or have("dhcpd"), "-pf", "/var/run/dhcpd-server/dhcp.pan1.pid", "pan1"]
+            cmd = [have("dhcpd3") or have("dhcpd"), "-pf", "/var/run/dhcpd.pan1.pid", "pan1"]
             p = Popen(cmd)
 
             ret = p.wait()
 
             if ret == 0:
                 dprint("dhcpd started correctly")
-                f = open("/var/run/dhcp-server/dhcpd.pan1.pid", "r")
+                f = open("/var/run/dhcpd.pan1.pid", "r")
                 self.pid = int(f.read())
                 f.close()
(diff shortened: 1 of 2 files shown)


cschramm:
For me the PID file does not get written and thus cannot be read.

The reason is that the containing directory, /var/run/dhcpd-server, does not exist.

You changed that from `/var/run/dhcpd3-server` in https://github.com/blueman-project/blueman/commit/a65d527b6de672bb15a8b192c568391ace2f9403. I can only find references to such a subdirectory in Ubuntu's `/etc/init/isc-dhcp-server.conf` (which also creates it). Upstream uses `/var/run/dhcpd.pid` as the default PID file: https://source.isc.org/cgi-bin/gitweb.cgi?p=dhcp.git;a=blob;f=includes/site.h;hb=f883cc903171b43537434d81076a7bb6de7ada62#l137

infirit:
>The reason is that the containing directory, /var/run/dhcpd-server, does not exist.

Neither does ```/var/run/dhcpd3-server``` for me on jessie and wheezy. Isn't creating this directory the job of the blueman package?

>You changed that from /var/run/dhcpd3-server in a65d527

Yeah, I made it more broken than it already was 😨. Also notice hat the pid filename also does not match ... sigh I really had to test this better.

Ok, is there still a point in supporting dhcpd3? Its not in wheezy afaics, checking fedora now

cschramm:
> Isn't creating this directory the job of the blueman package?

Well, it could but never did. If we want to that, I think the name should better refer to blueman. Anyway, I don't see a good reason to use that subdirectory. We don't do it for the dnsmasq counterpart as well.

> Yeah, I made it more broken than it already was

I don't think so. The Internets have no reference to `/var/run/dhcpd3-server` at all but do for `/var/run/dhcpd-server`.

> Ok, is there still a point in supporting dhcpd3?

I think that's gone for at least five years now, so no, probably not.

infirit:
Added the two commits from #757 and one extra to filter ipv6 name servers.

PR #737, opened 2017-07-30
Network and Serial Service: Add property for availability

infirit:
For NetworkService we now handle the missing dbus interface gracefully.

For SerialService the devices I have either don't expose the uuid until
paired or ask to be paired when we attempt to connect.

Also set menu entry sensitive based of this.


infirit:
Rebased against master

PR #738, opened 2017-07-30
mechanism: Add option to stop timer

infirit:
This is useful for debugging so we can keep it running in a terminal.


infirit:
oh and I care very little about the name so feel free to suggest another name for it 😸

PR #742, opened 2017-08-07
Send bytes to os.write

infirit:
PyCharm flagged this but I have no devices that use PPP. Needs a tester but it has been broken for some time now so not sure how many folks still use this..


infirit:
Python2 will indeed just convert between the bytes and codepoints.

PR #743, opened 2017-08-09
Store an array of dicts for recent-connections

infirit:
I wanted to do this for some time. It could use some more work but that is for another time.

Instead of storing a binary blob as string we now store

```
[{'adapter': '00:02:72:C5:BB:B4', 'address': '00:07:61:48:7A:2C', 'alias': 'Logitech MX1000 mouse', 'icon': 'input-mouse', 'name': 'Auto connect profiles', 'uuid': '00000000-0000-0000-0000-000000000000', 'time': '1502298305.8234344', 'device': '', 'mitem': ''},
{'adapter': '00:02:72:C5:BB:B4', 'address': 'D2:0C:FD:5A:21:12', 'alias': 'M720 Triathlon', 'icon': 'input-mouse', 'name': 'Auto connect profiles', 'uuid': '00000000-0000-0000-0000-000000000000', 'time': '1502298328.266433', 'device': '', 'mitem': ''}]
```

### DIFF
--- a/blueman/plugins/applet/RecentConns.py
+++ b/blueman/plugins/applet/RecentConns.py
@@ -233,18 +233,18 @@ class RecentConns(AppletPlugin):
     def on_item_activated(self, menu_item, item):
         logging.info("Connect %s %s" % (item["address"], item["uuid"]))
 
-        item["mitem"].props.sensitive = False
+        menu_item.props.sensitive = False
 
         def reply(*args):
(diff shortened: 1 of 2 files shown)


cschramm:
I guess the intention was to add some compression but I don't really see a reason for it, so this looks good. 👍

infirit:
Thanks, rebased to master and merged.

PR #744, opened 2017-08-09
Sendto: Replace progresbar with spinner and always discover devices

infirit:
The logic has been added just need to make the spinner and combobox look nicer. See #595 for the details on this.

afbeelding

### DIFF
--- a/apps/blueman-sendto.in
+++ b/apps/blueman-sendto.in
@@ -107,7 +107,7 @@ class SendTo:
             return []
 
     def select_device(self):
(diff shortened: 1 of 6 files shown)


infirit:
OK looks ok now imo.

afbeelding

infirit:
I also added some other commits to,
* Fix a bug when a source adapter was provided but not a device, we would display device on the wrong adapter.
* Use the last adapter name stored in gsettings if no source adapter was provided.

infirit:
rebased against master

PR #748, opened 2017-08-14
ManagerDeviceList: Fix logic to determine rss and tpl state

infirit:
>For example, if rssi was exactly 30 rssi_state wouldn't be set. Revised
after correction by @cshramm

This should cover everything (I hope 😨)

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -453,32 +453,24 @@ class ManagerDeviceList(DeviceList):
 
                     if rssi < 30:
                         rssi_state = _("Poor")
-
-                    if 40 > rssi > 30:
+                    elif rssi < 40:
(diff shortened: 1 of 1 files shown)


blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -442,34 +442,28 @@ def tooltip_query(self, tw, x, y, kb, tooltip):
...
-                    elif 40 < rssi < 60:
+                    elif rssi < 60:
                         rssi_state = _("Optimal")
-
-                    elif rssi > 60:
+                    elif rssi < 70:
                         rssi_state = _("Much")
-
-                    elif rssi > 70:
+                    elif rssi >= 70:
```
cschramm: That's just an else.

infirit: Changed that and merged it.

PR #766, opened 2017-09-05
Drop adding methods to plugin classes

infirit:
This is not meant to be merged, it is more to help understand.

I see no reason to add the methods to the BasePlugin class when we can just check for it when looping over all instances. Is there a benefit to doing it like this compared to how it is done in this PR?

### DIFF
--- a/blueman/main/PluginManager.py
+++ b/blueman/main/PluginManager.py
@@ -205,28 +205,25 @@ class PluginManager(GObject.GObject):
     def Run(self, func, *args, **kwargs):
         rets = []
         for inst in self.__plugins.values():
-            try:
-                ret = getattr(inst, func)(*args, **kwargs)
-                rets.append(ret)
-            except Exception:
-                logging.error("Function %s on %s failed" % (func, inst.__class__.__name__), exc_info=True)
+            if hasattr(inst, func):
+                try:
+                    rets.append(getattr(inst, func)(*args, **kwargs))
+                except:
+                    logging.error("Function %s on %s failed" % (func, inst.__class__.__name__), exc_info=True)
 
         return rets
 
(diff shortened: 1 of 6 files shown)


cschramm:
I don't see any, no.

blueman/main/PluginManager.py
```diff
@@ -205,28 +205,25 @@ def get_plugins(self):
...
-            try:
-                ret = callback(inst, ret)
-            except StopException:
-                return ret
-            except Exception:
-                logging.error("Function %s on %s failed" % (func, inst.__class__.__name__), exc_info=True)
-                return
-
-            if ret is not None:
-                args = ret
```
cschramm: This was obviously meant for chaining. If the callback returns something it shall be used as `*args` for the next method call. I don't know if we actually use that anywhere, though.

cschramm: No, none of the callbacks passed to `RunEx` returns anything. They all just raise `StopException`s.

infirit: Ah, good one, I missed that.

infirit:
Thinking about this some more and taking ``on_status_icon_query_icon`` as an example. Below is hypothetical as ShowConnected and PowerManager can not return values at the same time, but ...

self.__plugins is a dict and we can not assume order [1] in it. And if we had 2 plugins returning values for ``on_status_icon_query_icon`` there is no way to know which plugin will return first.

The same applies for the chaining, how do we know which plugin is going to provide args to the other?

Something to remember for the future when adding new plugins 😄.

Thanks for the feedback 👍 

[1]  cpython in 3.6 does provide order of insertion for dicts but for now should be considered implementation detail.

PR #778, opened 2017-10-04
Rename __parent__ to parent and make it available to all plugins

infirit:
I was looking to always have the parent available for general use [1] and found it was already there as ``__parent__``. 

```
>__parent__ is set in BasePlugin and is actually the Applet, Manager or
Service instances passed in by the PluginManager. However we also pass
these instances to on_load function meaning we may end up adding them
to the plugin twice, once as __parent__ during init and again in on_load.

This also unifies the name "parent" across all plugin bases classes.
```

[1] I am tired of fighting the GDBus service implementation from the pithos folks and make it work with our plugin system. I am looking at doing it differently.

### DIFF
--- a/apps/blueman-tray.in
+++ b/apps/blueman-tray.in
@@ -1,11 +1,5 @@
 #!/usr/bin/env python3
 # coding=utf-8
-
-from __future__ import print_function
-from __future__ import division
-from __future__ import absolute_import
-from __future__ import unicode_literals
-
 import logging
 import os
 import sys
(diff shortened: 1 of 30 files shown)


infirit:
Rebased against master. Will do some testing as the rebase was not trivial due to the tray changes.

infirit:
While I was in the code anyway I did a little cleanup and fixed the busy cursor when launching he tray.

It was indeed the blueman-tray that caused the busy cursor. Adding ``sn=False`` to the launch command solves that problem.

I'll merge the whole lot this weekend unless you have some changes you like made.

cschramm:
Note: This conflicts with #797 as that introduces some new references to `self.Applet`.

infirit:
If you merge that I'll fix anything still left this weekend.

infirit:
Ok, all done and a few commits for little things I ran into while working on this.

I dropped the change from Gtk loop to GLib's as the ExitItem plugin calls Gtk.main_quit.

PR #779, opened 2017-10-09
Cleanups and rework for the blueman/bluez

infirit:
The big change is the use of GObjectMeta to replace the ugly ``__new__`` and ``_init`` code. It should have no functional changes and doesn't in my testing.

Te other big(ish) one is not rely on ``__del__`` to be run. The problem is that ``__del__`` is unreliable for cleanup as can be seen in a PR i'll open in a few minutes. There is also a better solution in the weakref module, also see other PR.

The rest is small things I noticed while doing the GObjectMeta work


blueman/plugins/applet/TransferService.py
```diff
@@ -173,18 +173,15 @@ def on_unload(self):
...
+        if not self._agent:
+            self._agent = Agent(self.Applet)
         self._agent.register()
 
-    @classmethod
-    def _unregister_agent(cls):
-        if cls._agent:
-            cls._agent.unregister()
-
-    def on_manager_state_changed(self, state):
```
cschramm: Why does this method get dropped? 🤔

infirit: When on_manager_state_changed is run with state == False bluez and obex are already gone so it is pointless to try and unregister the agent at that point. It is probably also pointless to do it on ``_on_dbus_name_vanished`` but I'll check and make sure.

infirit: Yeah, no point in unregistering. I'll drop ``_on_dbus_name_vanished`` as well 
```
blueman-applet 22.43.58 INFO     TransferService:191 _on_dbus_name_vanished: org.bluez.obex
blueman-applet 22.43.58 ERROR    AgentManager:28 on_unregister_failed: /org/bluez/obex/agent/blueman  The name :1.5300 was not provided by any .service files
```

infirit:
Split as mentioned in comment and merged 😄

PR #780, opened 2017-10-09
Discussion: Usage of __del__ in blueman

infirit:
I was looking at ``__del__`` usage in plugins. This does not work at all as it is used. Even on proper shutdown we never saw any of the messages printed in ``__del__``.

Commit message explains,
```
This is useful for plugins that need to stop scripts and tear down
things that must always be done. For example kill running scripts like
we do in SerialManager.

We can not rely on __del__ to be run and this is most likely why
SerialManager uses the atexit module. We can howver use weakref to
run a function when the plugin is about tot be deleted.

Now we see the debug message "Deleting plugin instance %s" at shutdown
which never showed for me before.

But, this raises an interesting issue, we never see this message when
disabling a plugin. More work has to be done to properly unload plugins.
```

None of the messages printed in ``__del__`` in all of blueman show up for me. So it is probably good to stop relying on it completely.

Question, are message like https://github.com/blueman-project/blueman/blob/master/blueman/gui/manager/ManagerDeviceMenu.py#L63 really useful as I don't see any? If not Ill just get rid of all of these.

### DIFF
--- a/blueman/bluez/obex/Manager.py
+++ b/blueman/bluez/obex/Manager.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 import logging
+import weakref
 from blueman.bluez.obex.Transfer import Transfer
 from gi.repository import GObject, Gio
 from gi.types import GObjectMeta
@@ -39,7 +40,9 @@ class Manager(GObject.GObject, metaclass=ManagerMeta):
         self.__signals.append(self._object_manager.connect('object-added', self._on_object_added))
         self.__signals.append(self._object_manager.connect('object-removed', self._on_object_removed))
 
-    def __del__(self):
+        weakref.finalize(self, self._on_delete)
+
+    def _on_delete(self):
         for sig in self.__signals:
             self._object_manager.disconnect(sig)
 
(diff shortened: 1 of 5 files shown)


cschramm:
> Question, are message like https://github.com/blueman-project/blueman/blob/master/blueman/gui/manager/ManagerDeviceMenu.py#L63 really useful as I don't see any?

I guess they never were...

cschramm:
Anything stopping us from landing this into master?

infirit:
>Anything stopping us from landing this into master?

Mostly time (busy) to rebase and make sure it still works properly with recent changes. It shouldn't be too difficult so hopefully I can finish it with the other things on the poke list.

infirit:
Got it sorted and fixed all the places __del__ was doing more than just print debugging information.

I can replace the remaining __del__ with weakref.finalize like the one I pointed to earlier but imo logging that an object is deleted is not very useful. In the ManagerDeviceMenu example the only time I would expect to see the message is when I close blueman-manager, I really do not care at that point :smile:.

cschramm:
`DeviceSelectorWidget` calls `self.List.destroy()` in the `__del__` method. Not sure if that's of any use. Probably not...

infirit:
>DeviceSelectorWidget calls self.List.destroy() in the __del__ method. Not sure if that's of any use. Probably not...

Not really, the DeviceSelectorWidget is only created once and goes away when the app is closed.

PR #792, opened 2017-11-20
NetConf.py: explicitly pass "dev" to ip(8) command

rworkman:
This shouldn't actually be a problem, but as long as it's
in there, we may as well set a good example and be pedantic.

This is a minor fixup to #759


cschramm:
👍 Thanks. 🙂

PR #800, opened 2017-12-02
Handle when obexd can not be autostarted

infirit:
Works for me on Gentoo (without systemd) but needs some testing on other ditros.

The TransferService plugin is handled nicely now. It waits for the name ``org.bluez.obex`` to show up before setting up the obex.Manager.

But in blueman-sendto I had to resort to checking the error message for ``StartServiceByName``. The string shows up in the error on a Fedora vm and my gentoo workstation so it should be fairly safe to rely on it.

The best way to test this is to to modify the obex dbus service file. Either set the exec entry to ``Exec=/path/to/obexd`` or ``Exec=/bin/false``.

### DIFF
--- a/blueman/main/Sendto.py
+++ b/blueman/main/Sendto.py
@@ -95,11 +95,6 @@ class Sender(Gtk.Dialog):
         self.num_files = len(self.files)
         try:
             self.client = obex.Client()
-        except obex.ObexdNotFoundError:
-            d = ErrorDialog(_("obexd not available"), _("obexd is probably not installed"))
-            d.run()
-            d.destroy()
-            exit(1)
         except GLib.Error as e:
             if 'StartServiceByName' in e.message:
                 logging.debug(e.message)
--- a/blueman/main/Sendto.py
+++ b/blueman/main/Sendto.py
@@ -14,6 +14,7 @@ gi.require_version("Gdk", "3.0")
 from gi.repository import Gdk
 from gi.repository import Gtk
 from gi.repository import GObject
+from gi.repository import GLib
 
 from blueman.bluez.Adapter import Adapter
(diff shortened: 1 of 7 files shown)


cschramm:
If obexd cannot get started, I run into `TransferService._on_dbus_name_vanished` after 15 seconds where `self._agent.close()` fails as `self._agent` is `None`. Doesn't that happen for you? 🤔

Regarding blueman-sendto, a dialog would be nice. I think we should catch the `GLib.Error` right in `Client.__init__` and raise an `ObexdNotFoundError`. The `Sendto` class will take care about the dialog then.

infirit:
Both fixed now.

blueman/main/Sendto.py
```diff
@@ -94,11 +95,17 @@ def __init__(self, device, adapter_path, files):
         self.num_files = len(self.files)
         try:
             self.client = obex.Client()
-        except obex.ObexdNotFoundError:
```
cschramm: This can get raised in `Client.__init__` in case an old obexd (for BlueZ 4) is found. It looks like I ran into that situation when working on #259. I'm fine with ignoring the case, but then we should remove the related code from `Client.__init__`.

cschramm:
Looks good now on my side but see my comment on `ObexdNotFoundError.

blueman/main/Sendto.py
infirit: Yeah, its not relevant in bluez5 anymore. I'll drop that in a separate commit later tonight (first have something to eat). Not on my best game lately :disappointed: .

infirit: Should be good now. Will be testing a bit more this week. Off to get some :coffee:.

Issue #691, opened 2017-05-07
StatusIcon does not update with rfkill unblock 0


infirit:
When testing connman changes I used out statusicon to disable the adapter. Then I used ``rfkill unblock 0`` and the adapter started working again. However the statusicon remained as disabled.

Using the statusicon to enable and disable however works fine. My question is, are we supposed to handle this situation? I think we should though 😄.

cschramm:
Definitely yes. I'm not sure why we don't. My guess what happens is that something (I suppose udev) brings up the adapter automatically when you unblock it. What should happen then is:

1. `PropertyChanged` D-Bus signal for the `Powered` property of `org.bluez.Adapter1`
2. `BluemanApplet._on_adapter_property_changed`
3. `PowerManager.on_adapter_property_changed`
4. `PowerManager.UpdatePowerState`
5. `StatusIcon.IconShouldChange`
6. `StatusIcon.on_status_icon_resized`
7. `PowerManager.on_status_icon_query_icon` returns `None`
8. `StatusIcon.props.icon_name` gets set to an enabled icon

That chain looks good to me, so my guess is that point one never happens but that would be really strange... 🤔

infirit:
So, I enabled ``AutoEnable=true`` in bluez's main.conf and this is actually powering the adapter for me. I do see the property ``Powered`` change in bluetoothctl so we probably are at fault here. Marking as bug and I will investigate further.

cschramm:
Works for me in f34455f.

infirit:
that is annoying, it doesn't for me.

I'm building a fresh gentoo install and will check if i get the same in there.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

cschramm:
Status?

infirit:
Not sure, will poke at it this weekend to confirm.

infirit:
Tracked it down to ``on_power_state_changed`` not being run on the applet plugins. It should be done in ``UpdatePowerState`` when ``self.current_state != new_state``.

Relevant bit from the debug log,

```
blueman-applet 00.08.02 INFO     KillSwitch:110 io_event  : killswitch changed 0
blueman-applet 00.08.02 INFO     KillSwitch:120 io_event  : State: True
blueman-applet 00.08.02 INFO     PowerManager:168 UpdatePowerState: off True | foff False | on True | current state False | new state False
[...]

PR #801, opened 2017-12-03
PowerManager: Update adapter_state on adapter power changes

infirit:
See github issue #691

### DIFF
--- a/blueman/plugins/applet/PowerManager.py
+++ b/blueman/plugins/applet/PowerManager.py
@@ -101,15 +101,17 @@ class PowerManager(AppletPlugin):
 
         def check(self):
             if self.called == self.num_cb:
+                GLib.source_remove(self.timer)
                 logging.info("callbacks done")
                 self.parent.set_adapter_state(self.state)
-                GLib.source_remove(self.timer)
+                self.parent.UpdatePowerState()
                 self.parent.request_in_progress = False
 
         def timeout(self):
             logging.info("Timeout reached while setting power state")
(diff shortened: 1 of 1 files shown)


cschramm:
I don't think that's 100 % correct. If it's about two adapters that are not powered and you power one of them, you would set `adapter_state = True` while `get_adapter_state` would still return `False`. Maybe we should just use that method? 🤔

infirit:
New approach is to call ``set_adapter_state`` which handles all adapters and it updates adapter_state when done. This does mean that if one is powered all are powered and the other way around.

cschramm:
Not sure about the implications of that. Let's try.

infirit:
I dug a bit deeper today. The actual problem is in ``RequestPowerState`` when ``self.UpdatePowerState()`` is never run after ``cb.check()``. ``cb.check()`` correctly calls ``set_adapter_state`` but for some reason ``GLib.source_remove(self.timer)`` causes problems, maybe ``cb.check()`` blocks, not sure yet but very likely. But I do know that when I comment ``GLib.source_remove(self.timer)`` out ``self.UpdatePowerState()`` is run. Hopefully have a proper fix later tonight.

infirit:
It did not block so that was good to see. However the bug is very subtle and it was just that we ended up calling ``UpdatePowerState`` too early. Moving it into the Callback the ui updates properly now.

No offense to anyone but bl**dy hell, this is one complicated plugin :fearful:

PR #805, opened 2017-12-09
blueman-tray: Port it to use Gio.application

infirit:
This also changes the behavior of how it runs. Instead of shutting down it just hides the tray. It also adds a quit action which can be used over dbus on interface org.freedesktop.Application with the ActivateAction method. One can add more actions later to handle other things like have it reload the indicator implementation on the fly.

Another things that might be interesting is to make it dbus activatable.


blueman/main/Tray.py
```diff
@@ -10,16 +10,18 @@
 from gi.repository import GLib, Gio
 
 
-class BluemanTray(object):
+class BluemanTray(Gio.Application):
     def __init__(self):
-        check_single_instance("blueman-tray")
```
cschramm: I guess `GApplication` takes care of this (via the `application_id`), right?

cschramm: `StatusIcon._on_plugins_changed` does not work without this (as the lockfile does not get written). It needs to use the quit action now.

cschramm:
> Instead of shutting down it just hides the tray.

Not sure if I like leaving blueman-tray running even when the user clicks the exit entry...

blueman/main/Tray.py
infirit: The application id is used to identify the app. Then in ``do_activate`` it can do anything you like one of the options is to only allow one instance.

I'll add a proxy like appletservice and make it easy to quit the tray.

infirit:
I went the ``os.execv`` route and this works well. I'll open a new PR porting everything to G(tk)Application.

PR #810, opened 2017-12-16
Fix most issues reported by pycodestyle

cschramm:
This is a rebase of #735 on current master.

`pycodestyle --max-line-length=120 --ignore=E402,E722 blueman` does not report any issues with it. Not sure if anything broke though... 😆

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -10,7 +10,7 @@ import gi
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
 
-#support running uninstalled
+# support running uninstalled
 _dirname = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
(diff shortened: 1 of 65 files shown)


infirit:
Have not tried but I noticed 24d5dad0586aac1aabebfcf907f90bccf05a6594 brings back the blueman/gui/Dialogs.py which was removed in favor of a common error dialog put in CommonUi.

infirit:
Testing now. Below is what I found and they may also be in my original branch (not checked).

* UpdatePowerState was left in PowerManager on line 107.
* in DBusService plugin GetLoaded and GetClasses have not been renamed
* ``__all__`` was duplicated in Functions.py

If anything else comes up I'll post them here.

infirit:
Have not found anything since last comment. We can take care of any leftovers after merging. agreed ?.

PR #811, opened 2017-12-23
Cleanups and dbusproxy move

infirit:
Some of this was also in another PR and it will need a rebase if this is merged. 

The big move is because the proxy classes have much common code and do pretty much the same thing. I think this is cleaner as it is immediately obvious that we are dealing with dbus proxies

The other big(ish) change is that we now raise an exception when we fail to initialize a proxy (eg failed to autostart). This reduces the number of bare exception which is imo always a good idea :smile:. Feel free to suggest a better name for the exception, I couldn't come up with a better name :disappointed: .

Misc changes:
* Use importlib to load plugins
* Remove a few unused imports


infirit:
Added commits that removes many bare exceptions. Still testing but things look good.

PR #815, opened 2017-12-28
Fix lots of small issues with rfcomm and serial

infirit:
Lots of small fixes for issues I found testing the SerialManager.

Turns out that releasing the rfcomm device as normal user failed with permission denied when using the command line utility. So it looks like we are only allowed to release it as root.

### DIFF
--- a/blueman/plugins/applet/DBusService.py
+++ b/blueman/plugins/applet/DBusService.py
@@ -87,7 +87,7 @@ class DBusService(AppletPlugin):
 
                 self.parent.Plugins.run("on_rfcomm_disconnect", port)
 
-                logging.info("Disonnecting rfcomm device")
+                logging.info("Disconnecting rfcomm device")
             else:
 
                 def cb(_inst, ret):
--- a/blueman/plugins/applet/DBusService.py
+++ b/blueman/plugins/applet/DBusService.py
@@ -83,7 +83,7 @@ class DBusService(AppletPlugin):
             service = get_service(Device(object_path), uuid)
 
             if service.group == 'serial':
-                service.disconnect(port)
+                service.disconnect(port, reply_handler=ok, error_handler=err)
 
                 self.parent.Plugins.run("on_rfcomm_disconnect", port)
 
--- a/blueman/plugins/applet/DBusService.py
+++ b/blueman/plugins/applet/DBusService.py
@@ -110,7 +110,7 @@ class DBusService(AppletPlugin):
     def rfcomm_connect_handler(self, service, reply_handler, error_handler):
         return False
 
-    def on_rfcomm_connected(self, device, port):
+    def on_rfcomm_connected(self, service, port):
         pass
 
     def on_rfcomm_disconnect(self, port):
--- a/blueman/plugins/applet/DBusService.py
+++ b/blueman/plugins/applet/DBusService.py
@@ -23,22 +23,10 @@ class DBusService(AppletPlugin):
         AppletPlugin.add_method(self.rfcomm_connect_handler)
         AppletPlugin.add_method(self.service_connect_handler)
         AppletPlugin.add_method(self.service_disconnect_handler)
-        AppletPlugin.add_method(self.on_device_disconnect)
 
     @dbus.service.method('org.blueman.Applet', in_signature="", out_signature="as")
     def QueryPlugins(self):
         return self.parent.Plugins.get_loaded()
-
-    @dbus.service.method('org.blueman.Applet', in_signature="o", out_signature="", async_callbacks=("ok", "err"))
-    def DisconnectDevice(self, obj_path, ok, err):
-        dev = Device(obj_path)
-
-        self.parent.Plugins.run("on_device_disconnect", dev)
-
-        def on_timeout():
-            dev.disconnect(reply_handler=ok, error_handler=err)
-
-        GLib.timeout_add(1000, on_timeout)
 
     def on_device_disconnect(self, device):
         pass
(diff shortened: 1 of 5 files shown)


blueman/plugins/applet/DBusService.py
```diff
@@ -78,8 +78,12 @@ def cb(_inst, ret):
     @dbus.service.method('org.blueman.Applet', in_signature="osd", async_callbacks=("ok", "err"))
     def disconnect_service(self, object_path, uuid, port, ok, err):
         if uuid == '00000000-0000-0000-0000-000000000000':
+            def on_timeout(dev):
+                dev.disconnect(reply_handler=ok, error_handler=err)
+
             device = Device(object_path)
-            device.disconnect(reply_handler=ok, error_handler=err)
+            self.parent.Plugins.run("on_device_disconnect", device)
```
cschramm: That does not seem right to me. Triggering `SerialManager.on_device_disconnect` here would mean that all connected serial services get disconnected when generic services (audio, HID) get disconnected.

I think what we actually need to do is transform `SerialManager.on_device_disconnect` into an `on_device_removed` method to kill the remaining watchers when the device got removed.

blueman/services/meta/SerialService.py
```diff
@@ -76,5 +76,14 @@ def connect(self, reply_handler=None, error_handler=None):
                 raise e
         return True
 
-    def disconnect(self, port_id):
-        Mechanism().close_rfcomm('(d)', port_id)
+    def disconnect(self, port_id, reply_handler=None, error_handler=None):
+        try:
+            Mechanism().close_rfcomm('(d)', port_id)
```
cschramm: I guess that's due to the `release_rfcomm_device` call failing in mechanism, no?

blueman/plugins/applet/DBusService.py
infirit: Hmm, I'll look into it.

blueman/plugins/mechanism/Rfcomm.py
```diff
@@ -22,3 +23,6 @@ def close_rfcomm(self, port_id):
         for line in out.decode("UTF-8").splitlines():
             if command in line:
                 os.kill(int(line.split(None, 1)[0]), signal.SIGTERM)
+
+        # FIXME for some reason only root can release, revisit in the future
+        release_rfcomm_device(port_id)
```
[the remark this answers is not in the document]
infirit: Is there another process keeping the device open. ModemManager comes up in web searches.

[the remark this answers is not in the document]
infirit: Also, what happens if try to release it by command line? Permission denied or something else?

blueman/services/meta/SerialService.py
infirit: Not specifically but it would get caught. Any failure in the proxy will be caught here and if there is a handler have it shown in the manager.

blueman/plugins/mechanism/Rfcomm.py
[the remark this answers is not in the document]
cschramm: OK, did some more tests. There *is* a difference between root and non-privileged (but in dialout group).

Steps:
1. Open a python terminal
2. Call `create_rfcomm_device`
3. Call `release_rfcomm_device`
4. Call `create_rfcomm_device`
5. Run `blueman-rfcomm-watcher /dev/rfcomm0`
6. Call `release_rfcomm_device`
7. Call `create_rfcomm_device`
8. Run `blueman-rfcomm-watcher /dev/rfcomm0`, then kill it -> `/dev/rfcomm0` gets removed
9. Call `release_rfcomm_device`

Step 3 always succeeds. Step 9 always gives error 13 as the device is gone. If I run the python terminal as an unprivileged user in the dialout group, step 6 also gives error 13. If I run it as root, step 6 succeeds. So, it looks like, other than an unprivileged user, root can release a device that is currently in use.

However, we should have killed `blueman-rfcomm-watcher` at this point and there's probably no point in releasing the device if anything else is still using it as it will get auto-closed after that.

blueman/plugins/applet/DBusService.py
infirit: Thought about it and there is no way we can keep the serial devices connected. Generic disconnect will actually disconnect the device and any services setup will be forcefully disconnected. 

I can however move this out of DBusService and have SerialManager handle when the device Connected property changes to False. It makes more sense to do this as the only user of ``on_device_disconnected`` is SerialManager.

blueman/plugins/mechanism/Rfcomm.py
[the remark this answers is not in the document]
infirit: The problem is that if we are disconnecting the device there is no way other users of the device can keep using it. What I'll do is add a ``rfcomm_release`` method on Mechanism and implement the logic to deal with disconnecting the device and disconnecting the service.

infirit:
There appears to be no negative affect by dropping ``release_rfcomm_device``. Killing the watcher is enough for it to release by itself.

Updated the PR with the changes and tested this with root and user watches.

There is one annoying thing which happens to me regardless of the changes here. For some reason the actual device nodes ``/dev/rfcom*`` are not cleanud properly. So currently I am at rfcomm11 :disappointed: with no rfcomm devices listed with command line util ``rfcomm`` or our ``rfcomm_list``. it doesn't look like it causes problems but I would like to know if this happens to more folks :smile:.

cschramm:
The only way I can currently provoke increased device numbers is by terminating the underlying connection e.g. by just turning the connected device off. The rfcomm device file persists but `rfcomm` also lists the connection (as closed) for me. A device file without a corresponding entry in the `rfcomm` output sounds like a kernel issue to me. Did you check `fuser` / `lsof` for the file?

PR #822, opened 2018-01-12
A few fixes so that we have no more tracebacks stopping bluetoothd

infirit:


infirit:
And when you run it for a little longer it becomes clear why it was not used. The problem I ran into was
that when we emit our ``property-changed`` signal the cached property has not been updated. So when we try to get the Connected property for example we may get the wrong result and not start monitoring power levels in the DeviceList.

Instead when we fail to get a property directly from dbus attempt to get it from the cache and if that fails go ahead and raise an error. Updated the PR and also added a few commits for trivial fixes.

PR #825, opened 2018-01-20
Update travis ci

infirit:
We do not run anything right now which is a shame. This PR updates the code to run flake8 and make it pass.

Where we can not fix the code add a ``noqa`` line comments. If there was no FIXME for these cases I added one.

With this flake8 passes, see https://github.com/infirit/blueman/pull/1


.travis.yml
```diff
@@ -1,12 +1,24 @@
+language: python
+
+python:
+  - "3.4"
+  - "3.6"
+  - "3.7-dev"
+
+install:
+ - python -m pip install flake8
```
cschramm: I don't know if it makes a difference, but I'd use python3.

infirit:
The main reason to drop pylint is false positives like below. They are flagged because pylint does not support dynamic functions names so it will not find ``QueryPlugins``. Also tries not to import modules so it has no way to know that ``GLib.Error`` is a proper Exception.

There is a way to tell pylint of each function through a plugin, see https://www.logilab.org/blogentry/78354. In the long run this may be worth the effort but to have something running on travis __now__ I used flake8 .

```
************* Module blueman.Functions
E: 67,25: Instance of 'AppletService' has no 'QueryPlugins' member (no-member)
E: 68,15: Instance of 'AppletService' has no 'get_bluetooth_status' member (no-member)
E: 82,16: Instance of 'AppletService' has no 'SetBluetoothStatus' member (no-member)
E: 83,23: Instance of 'AppletService' has no 'get_bluetooth_status' member (no-member)
E:135,11: Catching an exception which doesn't inherit from Exception: Error (catching-non-exception)
E:140,15: Catching an exception which doesn't inherit from Exception: Error (catching-non-exception)
```

.travis.yml
infirit: Each run is done in separate virtualenv instances and bin/python is linked to the correct version.

cschramm: Ah, sure. `python-flake8` was the one that confused me.

cschramm:
All good. Let's try. 😃

.travis.yml
infirit: Ooh, forgot to drop python-flake8 from the commit. Will do before merging.

When we re-enable the actual build we will can change to using python3 and drop setting the environment variable :smile:

infirit:
When we re-enable the build we should use something like below travis.yml to split python checks and the build.

```
matrix:
  include:

    - language: python
      python:
        - "3.4"
        - "3.6"
        - "3.7-dev"
      script:
        - python -m flake8
      install:
        - python -m pip install flake8

    - language: c
      env:
        - PYTHON="python3"
      script:
        - ./autogen.sh
        - make
        - make check
      before_install:
        - sudo add-apt-repository ppa:vidplace7/bluez5-backports -y
        - sudo apt-get update -qq
        - sudo apt-get install aptitude
        - sudo aptitude install ~OLP-PPA-vidplace7-bluez libbluetooth-dev
        - sudo apt-get install -qq automake autoconf libglib2.0-dev libtool libgtk-3-dev libstartup-notification0-dev python-gi-dev intltool python3-dev cython python3-dbus dh-autoreconf
```

PR #828, opened 2018-01-26
Fix travis build for stable branch

### LINKED ISSUE #827: Travis and incorrect python cflags PKG_CHECK_MODULE
I have been poking at travis for stable and running into some weird problem. We probably will see the same on master but due to bluez5 missing we never get to build.

PKG_CHECK_MODULES sets PYTHON_CLFAGS to ``-g -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security`` (LIBS is fine) which is wrong. Calling ``pkg-config`` directly I do get the correct values ``-I/usr/include/python3.4m -I/usr/include/x86_64-linux-gnu/python3.4m``.

Trying to build in a xubuntu 14.04 vm (which is the same ubuntu version in travis) it is fine.

example build: https://travis-ci.org/infirit/blueman/jobs/333479403
branch on my fork: https://github.com/infirit/blueman/tree/travis_20

I am very confused.. :cry:

infirit:
So this is what is need to get the build working. Why PYTHON_CFLAGS is is being overwritten in travis I have no idea. Normal VM is fine so it probably has something to do with travis.

closes #827

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
+language: python
+
+virtualenv:
+  system_site_packages: true
+
+python:
+  - "2.7"
+  - "3.4"
+
 script:
   - ./autogen.sh
   - make
   - make check
-
-env:
-  - PYTHON=python2.7
-  - PYTHON=python3.2
 
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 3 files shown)


infirit:
The compiler does not know where to look for the includes which is what I saw as well. So it is likely the same problem.

PR #834, opened 2018-02-01
Allow setting systemd unit dirs separately, and get sane defaults from pkg-config

mgorny:
Currently blueman uses custom-made `--with-systemdunitdir` that sets top directory for systemd units. This doesn't work for split /usr systems because system units need to go into `/lib/systemd` while user units need to go into `/usr/lib/systemd`.

Replace the custom option with quasi-standard `--with-systemdsystemunitdir` and `--with-systemduserunitdir` that allow setting both necessary directories separately. <del>Furthermore, default to querying pkg-config for the correct paths from systemd.pc.</del>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -215,14 +215,23 @@ AC_ARG_ENABLE([appindicator],
 
 AM_CONDITIONAL([HAVE_APPINDICATOR], [test "x$use_appindicator" = "xyes"])
 
-AC_ARG_WITH([systemdunitdir],
-  AS_HELP_STRING([--with-systemdunitdir=PATH],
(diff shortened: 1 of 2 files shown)


mgorny:
Hmm, I see that the PKG_CHECK_VAR macro is not present in older versions of pkg-config, and I don't care enough to implement it manually.

infirit:
I "borrowed" some code (see 8ba8bcbdc4f26c8a5db6344aa967acecd26f8285) to add the pkgconfig function and merged your original commits. 

thx

mgorny:
Thanks a lot!

PR #838, opened 2018-02-03
RFC: Networking: Also allow for busybox binary for udhcpd

infirit:
Untested code just looking for feedback. In gentoo for example we have busybox and can (at own risk) have it create the symlinks but the default is not to. So we have busybox bot not the symlink to udhcpd. There should be no difference using ``busybox udhcpd`` but I am not sure if this is a good idea.


infirit:
Cool, I'll test and merge if it works properly.

infirit:
This caused multiple issues that I had not foreseen. It is not a good idea to do this so closing.

PR #844, opened 2018-02-24
Experiment moving to meson

infirit:
This is not meant to be merged anytime soon but wanted to put my experiment out here for folks to look at. 

I was interested how meson worked and tried to move blueman to use it. Along the way I ran into some problems where autotools supported features are not available in meson and had to workaround them.

The ubuntu image on Travis on travis is too old.
* meson does not have something similar to AM_CHECK_PYMOD. Because of that we need to use pkg-config but python3-dbus has no -dev version thus missing the pkg-config file :disappointed: 
* msgfmt in trusty does not have the --desktop option so meson crashes. I consider this a bug on meson as it should check if it has a recent enough gettext. I can workaround this however but I rather not and till 2.1 is out and we target more modern distros.

Everything works on my machine except translations which I'll look into fixing.


infirit:
>except translations

This was actually a problem with my user that set LANGUAGE to "en". Once unset translations are fine.

meson.build
```diff
@@ -0,0 +1,308 @@
...
+    output: 'blueman-manager.desktop',
+    type: 'desktop',
+    po_dir: 'po',
+    install: true,
+    install_dir: join_paths([datadir,'applications']),
+)
+
+i18n.merge_file(
+    input: 'data/thunar-sendto-blueman.desktop.in',
+    output: 'thunar-sendto-blueman.desktop.desktop',
```
cschramm: !

cschramm:
Thanks for sharing. Very interesting.

> python3-dbus has no -dev version

The pkg-config file is in python-dbus-dev.

meson.build
infirit: Good catch, the old copy paste error :disappointed:

infirit:
>The pkg-config file is in python-dbus-dev.

I tried that and and the check still failed. The problem was that python-dbus-dev depends on libdbus-1-dev and it was not installed (one can wonder why this was not done automatically :disappointed:). 

So the only thing left is the ancient gettext version in 14.04.

infirit:
Gettext supports xml but for polkit it requires files provided by polkit [1]. The version they are in is 0.114 and it got released very recently so nobody for the foreseeable future will have this. We will need to keep using intltool for polkit until then.

The other thing I have not tested is running uninstalled. The paths change and the build dir can be any name so currently this is broken.

[1] meson only supports gettext and it looks like it will never add intltool.

infirit:
There is a big regression with using meson which is running uninstalled. I have tried several things but I have not been able to make it work so I dropped the autotools removal and added the meson.build files to EXTRA_DIST so make dist will include them.

I'll fix the circleci build tonight

infirit:
Fixed the build on circleci

infirit:
Sure, just make sure to remember running uninstalled does not work. I got far making it work but never completely 😞

PR #849, opened 2018-03-08
Make travis use xenial

infirit:
It is a bit flaky and slow at times but it does work. Let's see if this PR passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 cache: pip
+dist: xenial
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


infirit:
It's too flaky. Had to restart jobs several times and it still is not done, I'll leave it open so that we can restart the job in travis to see if things improved.

infirit:
It seems to be pretty good now. I commented out python 3.4 as it failed to download hence the build was marked as failed. There is however no big difference between 3.4 and 3.5 so do we really need 3.4?

There is the occasional stall of the worker but usually a subjob restart sorts it out.

cschramm:
I'm fine with not checking 3.4. The only real thing we should keep an eye on is not to use PEP 448 then.

infirit:
Rebased and merged, let's see how it behaves. If it ends up to cumbersome let's just revert and not get annoyed too much :smiley:

infirit:
All the nightly cron builds failed for master and getting annoyed :disappointed:. Will revert this soon as possible.

Issue #239, opened 2015-04-04
Auto-connect services


cschramm:
Once in a while users want to have an auto-connect feature where a known service is connected automatically blueman is started or the device becomes available. @rsm-imap-cc figured out in #235 there are cases where BlueZ 4 does such an auto-connect, but BlueZ 5 does not, so this becomes more important with the migration.

jod35fan:
...and this Bluez5 migration takes years, if  Debian responsbles do not code and package anything. Meanwhile debian 8 users must use kdevil, where in xfce you must click 3 times button in 3 panels before it works. I think it is a waste of time to average user to have current blueman and bluez5 packages  in testing repositories. Tune blueman a bit and package the latest bluez5 with minor modifications that i have documented. Debian Jessie is otherwise very good, but this  bluetooth problem is not.

cschramm:
A draft for the "backend" pieces is in #853.

I do not have a concept for the GUI pieces yet. Some ideas:

* An option to enable auto-connect for the selected service in the device setup wizard. A list with removable auto-connect entries in the plugin settings.
* Entries to toggle auto-connect for specific services in the device menu, e.g. submenus instead of the immediate connect / disconnect items that contain the actual connect / disconnect item and an additional enable / disable auto-connect item.
* Auto-connect checkboxes next to specific services listed in an extended version of the "info" view.

bth73:
Hi, I'm hoping there's a fix for the auto connect feature. It's the last thing I need fixed  to have a perfect Asus T100 linux mint 19.1 install. My install is a few weeks old. I tried altering main bluetooth config file, 
# AutoEnable defines option to enable all controllers when they are found.
# This includes adapters present on start as well as adapters that are plugged
# in later on. Defaults to 'false'.
AutoEnable=true

The last line was already uncommented. Then I tried this:
# The ReconnectUUIDs defines the set of remote services that should try
# to be reconnected to in case of a link loss (link supervision
# timeout). The policy plugin should contain a sane set of values by
# default, but this list can be overridden here. By setting the list to
# empty the reconnection feature gets disabled.
ReconnectUUIDs=00001112-0000-1000-8000-00805f9b34fb, 0000111f-0000-1000-8000-00805f9b34fb, 0000110a-0000-1000-8000-00805f9b34fb
Removed the comment from last line.
No change with either.
I'm sort of a middle weight, so need complete instructions if there's a workaround or fix you can lay on me. Good at reading, following directions,copy and pasting.

[...]

PR #853, opened 2018-03-21
AutoConnect plugin

cschramm:
Initial draft for #239. It's currently just a "backend" you can configure via gsettings, e.g. `gsettings set org.blueman.plugins.autoconnect services '[("CC:9F:7A:A5:47:34","00000000-0000-0000-0000-000000000000")]'`.

The only useful UUIDs are the above one for auto-connect services (audio, HID) and the specific UUIDs for for other services we can handle (serial, network). Anything else currently causes a crash as `DBusService.connect_service` only expects those UUIDs.

I applied small fixes to the RecentConns plugin on the way. It used a wrapper for the `connect_service` method for no reason and in any case it depends on `DBusService` plugins to be loaded.

### DIFF
--- a/blueman/plugins/applet/AutoConnect.py
+++ b/blueman/plugins/applet/AutoConnect.py
@@ -1,3 +1,5 @@
+from gi.repository import GLib
+
 from blueman.Sdp import ServiceUUID
 from blueman.gui.Notification import Notification
 from blueman.plugins.AppletPlugin import AppletPlugin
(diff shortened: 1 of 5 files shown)


infirit:
I'll have a look this week.

For the ui I think an option is to have a list of devices and a list of profiles (that makes sense to autoconnect). Then have an add / remove button that add to the list, the list could be a liststore with a view and have the view show headers.

The profiles could also be a row of check buttons like PAN, AUDIO etc.

infirit:
One question I have is shouldn't it retry every X minutes to connect these devices? Now it only connects once on startup.

cschramm:
It currently only takes place on startup, yes. Unfortunately BlueZ does not tell us if a device becomes available (at least not via the D-Bus API).

I didn't actually think about such a scheme where we constantly retry. Not sure if we should do that but I guess it should not do any harm... :thinking:

cschramm:
Rebased and verified the functionality once again. :+1:

PR #867, opened 2018-04-28
Add net-tools and iproute to deps and configure checks for the commands

infirit:


cschramm:
Not sure about the check. It's a runtime-only dependency not relevant on the build system.

infirit:
It is in the runtime checks part so if distro folks want to skip it it they can. I can drop it if you like.

cschramm:
Oh, right. 👍

PR #870, opened 2018-04-29
Fix removed adapter handling (again)

infirit:


gfernandop:
Thanks for you support.

Does git have a way to apply the changes and revert if needed?
I work as network analyst having basic git experience.
Should I use a git command to apply the patch or replace each file manually (keeping a backup)?

Best Regards

PR #872, opened 2018-05-03
added test of installation to .travis.yml

krichter722:


krichter722:
Please restart the Travis build since it failed for errors which are unrelated to the changes (Travis seems to have troubles today).

infirit:
When adding installation make sure to set the configure options properly. For example you should add  ``--prefix=/usr/``.

However, I am unclear how running install helps us? What gains do we get?

krichter722:
> However, I am unclear how running install helps us? What gains do we get?

It tests that the installation works since there's a possibility that it might fail or dependencies used by the installation might change. It's nothing that should be left out of a CI script.

I think the prefix to test doesn't matter. If any I would test both with (an explicit) prefix and without (even though 2 installation checks might be overkill).

infirit:
>I think the prefix to test doesn't matter.

Well, I think you are very wrong. Just installing blueman to the default prefix of /usr/local may succeed but tests for a case which will never happen in real world. No distro or user should ever use the default prefix because it will install blueman broken.

>dependencies used by the installation might change

autoconf (meson perhaps in the future) is the place to deal with dependencies, why do you think installing is going to make a difference?

>It tests that the installation works since there's a possibility that it might fail

distcheck will run through all the steps plus cleanup. This may be good to run but  just make install with default prefix is pointless.

krichter722:
>>  I think the prefix to test doesn't matter.
>
> Well, I think you are very wrong. Just installing blueman to the default prefix of /usr/local may succeed but tests for a case which will never happen in real world. No distro or user should ever use the default prefix because it will install blueman broken.

I should have said "shouldn't matter" because a default value should always work (in any imaginable context inside and outside software development). Disclaimer: I don't want to start a big back and forth discussion with the following, but have to point it out so clearly: Distros not, but users would and should definitely use the installation prefix `/usr/local` (that's why the Linux Filesystem Hierarchy standard describes it) and this is why `./configure && make && make install` as well as `python setup.py && sudo python setup.py install` install there and that's what happens after following the build instructions in your `README`. That means that assuming that users can know that installing blueman into `/usr/local` causes a broken installation will cause trouble for everyone.

> autoconf (meson perhaps in the future) is the place to deal with dependencies, why do you think installing is going to make a difference?

To deal with dependencies, but not to test that they're sufficient to install successfully.

> distcheck will run through all the steps plus cleanup. This may be good to run but just make install with default prefix is pointless.

There's no guarantee that `distcheck` will always do that, so that one day the installation might not be tested entirely leaving possibility of unrecognized failure.

cschramm:
As usual with automake, the defaults work fine if the dependencies (like python, dbus, icon themes etc.) are in the (default) prefix. Let me quote @infirit: "[That] will never happen in real world".

> a default value should always work

There is no possible default with that premise. If the user wants to depend on stuff that doesn't reside in the prefix (especially his system's stuff), he has to supply the matching configuration. Report to automake if you think having a default is a bug.

I have no hard feelings about this PR. I guess running the installation does not hurt, although I don't see a good reason either.

infirit:
I don't mind running make install but let me explain a bit how blueman works because you seem to be under the impression it is just another python script (slowly you are figuring out through all the errors it is not).

Blueman has two dbus service files that have to be installed in the right place for dbus to be able to auto start our services. One of them, blueman-mechanism, is used to run low level privileged operation like creating a bridge network and starting a dhcp daemon. Because it runs these privileged operation as root normal users are not allowed to run them by default. We use policykit for this and have to install a policy file in the correct place for policykit to manage this. If this is done policykit will show a dialog every time you run one of these privileged operations. Most distros will install a policykit rule for this so user don't have to type the password all the time.

Next is that we use gsettings. Its schema file has to be installed in a place that gsettings can find it. Also the plain text schema files (not just blueman's, all of them) have to be be compiled into a single binary file before gsettings can use them.

If you want to use dhcpd3 or 4 as dhcp daemon each distro has slightly different names for the configuration file. Many distro have to override our default at configure, my distro gentoo has to do this (but don't and I have a bug report on their tracker).

I probably forgot a few things but you get the idea I hope.

There is no way the default prefix can work with blueman so please trust that we are at least slightly competent when we say that :wink:.

Now, if you add ``make distcheck`` to the list on top of ``make install`` I'll merge this.

infirit:
>Please restart Travis because build stalled.

It error-ed again because of permission denied. I'll fix that when I merge it so don't worry about that.

>I hope you didn't get the impression that I doubted your competence

No to me which is why I put the winking smily at the end :smile:.

>I strongly recommend to change the README

I need to go now and will be busy later, can you open an issue for me as a reminder. I should be able to take care of this over the weekend.

krichter722:
>> no info about the deb packages
>
>Huh? That's what I see on GitHub:

That's just wrong. I wanted to remove that from the comment before posting.

PR #894, opened 2018-07-13
Warn level pedantic

infirit:
Decided to turn up the warning and it came up with these two.

I am not sure about the named variadic macros one, I am not familiar enough with C to find a better solution :disappointed:. I can also drop it as the internets are of various opinions on how to deal with this one and if it is a problem worth dealing with.


cschramm:
I cannot reproduce the variadic macros issue. Maybe your compiler has different defaults that disable GNU extensions (I guess the warning should not show up with `-std=gnu99` but with `-std=c99`). I'm using GCC 7.3.0.

infirit:
I got them while trying different build types and flags with meson. I got both with ``meson --warnlevel 3 --buildtype debugoptimized``. The verbose ninja output is below.

```
[3/11] cc -Imodule/module@@_blueman@sha -Imodule -I../module -I/usr/include/pygobject-3.0 -I/usr/lib64/libffi-3.2.1/include -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/python3.6m -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -O2 -g -fPIC -pthread -DSN_API_NOT_YET_FROZEN  -MD -MQ 'module/module@@_blueman@sha/modem-prober.c.o' -MF 'module/module@@_blueman@sha/modem-prober.c.o.d' -o 'module/module@@_blueman@sha/modem-prober.c.o' -c ../module/modem-prober.c
../module/modem-prober.c:68:26: warning: ISO C does not permit named variadic macros [-Wvariadic-macros]
 #define verbose(fmt, args...) \
                          ^~~
../module/modem-prober.c: In function ‘modem_probe_caps’:
../module/modem-prober.c:364:40: warning: ISO C99 requires at least one argument for the "..." in a variadic macro
   verbose ("GCAP failed, trying ATI...");
                                        ^
../module/modem-prober.c: In function ‘do_probe’:
../module/modem-prober.c:508:3: warning: ‘fd’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   close(fd);
   ^~~~~~~~~
```

cschramm:
OK, got it. I thought `-Werror=variadic-macros` or `-Wall` would be enough but `-Wpedantic` did the trick. GCC only supports the named variable form for backwards compatibility, so I guess it makes sense that it reports an error when it's pedantic.

PR #896, opened 2018-07-20
Polkitauth 2 0

infirit:
Per e-mail discussion. merge and release? Opinions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 2.0.6
+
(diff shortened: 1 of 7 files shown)


cschramm:
Fine for me to release this as 2.0.6. Maintainers should definitely notice the instructions on what to do if it is the only change in the release. Many distributions are on 2.1 snapshot versions anyway, so they are no problem for this. The Debian package is in my own hands. What's left on 2.0.5 then are mainly Arch Linux and Slackware.

PR #897, opened 2018-07-22
Add circle config

infirit:
Running it for python2 and python3. Do we need to test on debian stable?


cschramm:
I'm fine with unstable only.

PR #898, opened 2018-07-23
Use ipaddress module to handle ip and network

infirit:
No problems with dnsmasq but will have to test udhcpd and dhcpcd. Unless I made typos it should work fine for them but better make sure.

Any 32bit packed ip's previously stored should get converted automatically. @cschramm, it would be nice if I could get your ``network.state`` file to test with and double my sample size :smile_cat:.

Too tired and hot to test the other dhcp handlers but opening PR so people can have a look and comment.

ps: I am experimenting with NetworkManager and NAP so it van be setup as normal user. I have it working in a test script which was nice. To integrate it in blueman I have to redesign the current AppletPlugin plus NetConf and passing "normal" strings around makes things so much simpler :smile:


cschramm:
> @cschramm, it would be nice if I could get your `network.state` file to test with and double my sample size

I don't have an interesting one (just the default) at hand, sorry.

Note that this raises the required Python version to 3.3. Nothing wrong with that, just have to update the dependencies.

infirit:
Fixed a couple bugs and bumped the python version to 3.3. Tested with the others and works well.

PR #899, opened 2018-07-27
Merge changelog from 2-0-stable

cschramm:
The changes are not 100% consistent as they are not linear but I think it is a good idea to have the 2.0 changelog entries in master as well.


infirit:
Looks good to me.

PS: circleci did not run which I'll investigate later

PR #905, opened 2018-08-08
Fix generic connect when ServiceUUID.short_uuid is None

### LINKED ISSUE #903: N900 no generic conect
I found my old nokia n900 and it does not show a generic connect option. bluetoothctl does connect the audio bits so i'll need to investigate what we need to add.

```
[Beast]# info 3C:F7:2A:07:BF:E3
Device 3C:F7:2A:07:BF:E3 (public)
	Name: Beast
	Alias: Beast
	Class: 0x005a020c
	Icon: phone
	Paired: yes
	Trusted: no
	Blocked: no
	Connected: yes
	LegacyPairing: no
	UUID: Vendor specific           (00000002-0000-1000-8000-0002ee000002)
	UUID: Dialup Networking         (00001103-0000-1000-8000-00805f9b34fb)
	UUID: IrMC Sync                 (00001104-0000-1000-8000-00805f9b34fb)
	UUID: OBEX Object Push          (00001105-0000-1000-8000-00805f9b34fb)
	UUID: OBEX File Transfer        (00001106-0000-1000-8000-00805f9b34fb)
	UUID: Audio Source              (0000110a-0000-1000-8000-00805f9b34fb)
	UUID: A/V Remote Control Target (0000110c-0000-1000-8000-00805f9b34fb)
	UUID: A/V Remote Control        (0000110e-0000-1000-8000-00805f9b34fb)
	UUID: Headset AG                (00001112-0000-1000-8000-00805f9b34fb)
	UUID: Handsfree Audio Gateway   (0000111f-0000-1000-8000-00805f9b34fb)
	UUID: Phonebook Access Server   (0000112f-0000-1000-8000-00805f9b34fb)
	UUID: Message Access Server     (00001132-0000-1000-8000-00805f9b34fb)
	UUID: Vendor specific           (00005005-0000-1000-8000-0002ee000001)
	UUID: Vendor specific           (00005601-0000-1000-8000-0002ee000001)
	RSSI: -74
```

infirit:
fixes #903

### DIFF
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -251,8 +251,9 @@ class ManagerDeviceMenu(Gtk.Menu):
                 show_generic_connect = True
                 break
             elif not service_uuid.reserved:
-                show_generic_connect = uuid == '03b80e5a-ede8-4b33-a751-6ce34ec4c700'
-                break
+                if uuid == '03b80e5a-ede8-4b33-a751-6ce34ec4c700':
+                    show_generic_connect = True
+                    break
 
         if not row["connected"] and show_generic_connect:
             connect_item = create_menuitem(_("_Connect"), "blueman")


blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -244,6 +244,9 @@ def generate(self):
         show_generic_connect = not device_uuids
         for uuid in device_uuids:
             service_uuid = ServiceUUID(uuid)
+            if service_uuid.short_uuid is None:
```
cschramm: If I get it right, this is true for all UUIDs where `service_uuid.reserved` is false. This means it also catches and thus breaks the handling of 03b80e5a-ede8-4b33-a751-6ce34ec4c700 for BLE devices.

The actual problem in my eyes is that `break` in the `elif` branch below is only right if `uuid == '03b80e5a-ede8-4b33-a751-6ce34ec4c700'`. We must not break the loop as long as `show_generic_connect` is `False`.

infirit: Good one, pushed a new commit for this.

PR #907, opened 2018-08-12
PPP fixes

infirit:
With this I see the commands being send by mechanism. And by increasing a dbus timeout I see proper error showing in blueman-manager.

It should works as before but I have no sim in the n900 so if it is actually useful I cant tell.


infirit:
I found a full size sim bracket so I can actually test this properly. More problems showed up :sob:. I'll investigate and add commits here.

infirit:
Merged with slight modifications. The timeout increase moved to the modemmanager check. And fixed another bytes vs codepoints issue.

PR #910, opened 2018-08-17
Re-implement PANU and DUN connection through NetworkManager

infirit:
Putting my work here as it is mostly done. Still need to clean a few things.

 * make sure error and reply handlers are reaching blueman-manager properly
 * Handle apn/number changes
 * Update the Service manager plugin to show the GsmSettings menu for NMDUNSupport

There is no api exposed over bluez dbus for DUN connection to check if it is actually connected like there is for PANU. It is trivial to get this from NM but afaics we have no functionality to communicate this to blueman-manager. Not sure if it is worth the effort though but if I can figure out a simple way I'll open a separate PR for this.

### DIFF
--- a/blueman/main/Makefile.am
+++ b/blueman/main/Makefile.am
@@ -17,7 +17,8 @@ blueman_PYTHON = \
 	Sendto.py \
 	Services.py \
 	Tray.py \
-	DBusProxies.py
+	DBusProxies.py \
+	NetworkManager.py
 
 if HAVE_PULSEAUDIO
 blueman_PYTHON += PulseAudioUtils.py
--- a/blueman/main/NetworkManager.py
+++ b/blueman/main/NetworkManager.py
@@ -1,0 +1,217 @@
+import gi
+import logging
+import uuid
+gi.require_version('NM', '1.0')
+from gi.repository import GLib, NM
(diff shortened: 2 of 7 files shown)


infirit:
Fixed the point and things work nicely for me on both panu and dun connections.

infirit:
There are apps on android that emulate dun if it is not available. I tried BlueDUN+ trial which appears to work but it is not free. I'll see if there is a free replacement but it may be worth buying to have something to test on recent phones.

infirit:
NM appears to crash bluedun but it works fine with PPPSupport. I'll see if I can get in touch with the dev and perhaps he can fix this.

cschramm:
Any news?

infirit:
I thought I commented here some time ago.. Anywayz

I backported this to stable and the user in #935 tested that. The user confirmed the NM bits works. I tested this as well with my n900 phone and it worked nicely as well. I do have the software solution but it tends to upset NM, not sure why as I haven't contacted the dev.

Now, I would like to see this merged in both master and stable so anything holding this up :question: 

ps: bluemans own implementation works for me with the software solution but not for the user in #935. This will be an ongoing project I fear but should not hold this up :smile:

PR #911, opened 2018-08-19
A few glib fixes for errors that started to annoy me

infirit:
I am not sure why the device-property-changed signal handler is not connected anymore but it is easy to check so do that instead.


infirit:
Added fix for typo in icon name.

PR #939, opened 2018-11-01
2 0 dun

infirit:
using the blueman service for dun is tested. The NM bits need work which I hope to deal with in the next couple of days.


infirit:
The NM parts are working for me now. The code in 2.0 is almost alien to me now :fearful:.

PR #950, opened 2018-12-02
Avoid authorization check

### LINKED ISSUE #948: "You might not be able to connect to the Bluetooth network via this machine"
blueman: 2.0.6
BlueZ: 5.50
Distribution: ArchLinux
Desktop environment: Xorg 1.20.3

Whenever I start `blueman-applet`, I get a popup stating: 

> "You might not be able to connect to the Bluetooth network via this machine"

There are two problems here (though I personally just care about the latter):

1. This tells me there's a problem with something. But I've no clue as to what the problem is, how it affects me, how to fix it, etc. UX-wise, this is really bad, because a typical user can only shrug, and click `OK`, without the slightest clue as to how it should be addressed nor what this means.
2. I don't really know how to get rid of this popup. `blueman` works for everything that I use it for, so, TBH, I just want to get rid of the popup warning everytime I run `blueman`.

Here's the console output for `blueman-applet`: https://gist.github.com/WhyNotHugo/00f92f516e439780994076e7ea5e86a1

cschramm:
We might ask for network setup authorization even if no network settings are configured and we do not actually set anything up. Moving the check in front of the confirmation should avoid that.

Fixes #948

### DIFF
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -349,12 +349,6 @@ class NetConf(object):
         return type(self.dhcp_handler)
 
     def apply_settings(self):
-        if self.ip4_address is None or self.ip4_mask is None:
-            if self.ip4_changed:
-                self.ip4_changed = False
-            self.store()
-            return
-
         if self != NetConf.get_default():
             NetConf.get_default().remove_settings()
             NetConf.default_inst = self
--- a/blueman/plugins/mechanism/Network.py
+++ b/blueman/plugins/mechanism/Network.py
@@ -45,8 +45,13 @@ class Network(MechanismPlugin):
 
     @dbus.service.method('org.blueman.Mechanism', in_signature="", out_signature="", sender_keyword="caller")
     def ReloadNetwork(self, caller):
-        self.confirm_authorization(caller, "org.blueman.network.setup")
         nc = NetConf.get_default()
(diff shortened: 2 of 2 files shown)


pi-theta:
Hi
I was going through your patch which claims to have solved the problem of #948
"You might not be able to connect to the Bluetooth network via this machine"

I am using Arch Linux 5.3.11 with blueman2.1.1-2 and am still facing the issue.
It pops up as a dialog box whenever blueman starts.
Have posted on arch forums[]( https://bbs.archlinux.org/viewtopic.php?id=246755)

Request if you can help resolve this?
Cheers and Regards

infirit:
https://wiki.archlinux.org/index.php/Blueman#Troubleshooting

pi-theta:
Already tried that a long time ago. If I disable the checkered box with the Networking Plugin in within the Blueman plugin window the popup doesn't show on next start of blueman.

But if I check the chekered box next to the plugin when blueman is running, it pops up immediately.

infirit:
@pi-theta sorry, can't help you if you can't read your own distros wiki properly. Let me quote it for you, 

``From version 2.0.6 the official documentation recommends creating polkit rules in order to avoid polkit agents asking for password on every boot, as root user add the following polkit rules:``  https://wiki.archlinux.org/index.php/Blueman#Permissions

If creating the polkit rules does not work raise a new issue and make sure to read the the issue template.

PR #951, opened 2018-12-02
Fix flake8 issues

cschramm:


infirit:
Never mind the other regular expressions, fixed them with something else in #952

PR #956, opened 2018-12-04
Add default polkit rules

cschramm:
At least in Arch it seems to be a policy to not change any sources, so its users currently do not get any rules that allow admin users to use privileged features. We now install a default rules file that targets the standard wheel group.

### DIFF
--- a/data/configs/Makefile.am
+++ b/data/configs/Makefile.am
@@ -19,9 +19,12 @@ endif
 
 if HAVE_POLKIT
 @INTLTOOL_POLICY_RULE@ 
-policykitdir = $(datadir)/polkit-1/actions
-policykit_in_files = org.blueman.policy.in
-policykit_DATA = $(policykit_in_files:.policy.in=.policy)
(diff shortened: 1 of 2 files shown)


cschramm:
Debian does not have `wheel` by default but at https://wiki.debian.org/SystemGroups it's documented for `pam_wheel`. The blueman.rules that I add in my Debian package (which would now override our default one and could also be a patch instead) I target groups `sudo` and `netdev`.

According to https://en.opensuse.org/SDB:Administer_with_sudo OpenSUSE they use `wheel` in sudoers.

Does Gentoo use `plugdev` in sudoers? :thinking:

infirit:
> Does Gentoo use `plugdev` in sudoers?

No, it uses wheel but the rules file the maintainer made uses ``plugdev``. But other than that it is identical so a sed oneliner will be enough to override this.

So the most common group is ``wheel``, lets merge this as is. Please also open a PR for 2.0 :smile:

PR #958, opened 2018-12-08
Use with statement, drop deprecated os.popen and call list on map object

infirit:
* Forward port calling list on map object, it return an iterator
* Use with statement, context managers :straight_ruler: 
* Use subprocess.Popen instead of os.popen
* Make a few methods static

### DIFF
--- a/blueman/main/PPPConnection.py
+++ b/blueman/main/PPPConnection.py
@@ -210,7 +210,7 @@ class PPPConnection(GObject.GObject):
 
         if found:
             lines = filter(lambda x: x != "", lines)
-            lines = map(lambda x: x.strip("\r\n"), lines)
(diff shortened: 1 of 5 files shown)


blueman/main/PPPConnection.py
```diff
@@ -210,7 +210,7 @@ def on_data_ready(self, source, condition, terminators, on_done):
 
         if found:
             lines = filter(lambda x: x != "", lines)
-            lines = map(lambda x: x.strip("\r\n"), lines)
+            lines = list(map(lambda x: x.strip("\r\n"), lines))
```
cschramm: A list comprehension would be nicer. `[x.strip("\r\n") for x in lines]`

cschramm:
```suggestion
            lines = [x.strip("\r\n") for x in lines]
```

PR #962, opened 2018-12-10
Fixes for next stable

infirit:
These commits are partly taken from master and one that needs forward port to master.


infirit:
Will need to forwardport 53695df59e9b89301e8feb7a2b8f49ab11ecce95 to master.

infirit:
Merged and also pushed a commit to update CHANGELOG. Unless you run into problems I think it should be ready for release.

PR #964, opened 2018-12-10
NetworkManager: Raise ImportError if NM python bindings are not avail…

infirit:
…able


infirit:
Also added a commit to add the new rules file to EXTRA_DIST so its included in the tarball.

PR #972, opened 2018-12-25
Class id is not handle in several places so revert for now

infirit:
changes in class_id do not work as I had thought so revert for now

I have been thinking and experimenting with moving ip, netmask and dhcphandler to gsettings. It will move the logic into applet plugins but this is not ready yet for this alpha.


infirit:
Yes, I was under the wrong impression the exception was handled. Looking at the commit history it never was. I can probably have NetConf do some sort of migration but I rather do it as normal user in the applet plugin.

PR #974, opened 2018-12-29
Add elementary-xfce as compatible icon-theme

cschramm:
Some time ago I identified elementary-xfce as an additional compatible icon theme. It actually looks like it's currently the only one that provides all of the icon names that we use (mate-icon-theme and gnome-icon-theme e.g. lack `menu-editor`).

### DIFF
--- a/Dependencies.md
+++ b/Dependencies.md
@@ -8,7 +8,7 @@
 * [GdkPixbuf](http://www.gtk.org/) (GI bindings)
 * [GLib 2, Gio 2](http://www.gtk.org/) (>= 2.32) (GI bindings)
 * [GTK+ 3](http://www.gtk.org/) (GI bindings) [1]
-* [gnome-icon-theme](https://git.gnome.org/browse/adwaita-icon-theme/) or [mate-icon-theme](https://github.com/mate-desktop/mate-icon-theme)
+* [gnome-icon-theme](https://git.gnome.org/browse/adwaita-icon-theme/) or [mate-icon-theme](https://github.com/mate-desktop/mate-icon-theme) or [elementary-xfce](https://github.com/shimmerproject/elementary-xfce)
 * [libappindicator](https://launchpad.net/libappindicator) (optional) (GI bindings)
 * [notification-daemon](https://developer.gnome.org/notification-spec/) that provides dbus name org.freedesktop.Notifications
 * [obexd 5](http://www.bluez.org/)


cschramm:
I identified 20 icon names that we use and do not provide ourselves. A quick survey in the current Debian repository brought up the following packages that provide at least 15 of them and four icon names that are problematic:

||gnome-power-manager|menu-editor|modem|network|
|-|-|-|-|-|
|elementary-xfce-icon-theme|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
|papirus-icon-theme|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
|faenza-icon-theme|:heavy_check_mark:|:heavy_check_mark:||:heavy_check_mark:|
|mate-icon-theme-faenza|:heavy_check_mark:|:heavy_check_mark:||:heavy_check_mark:|
|obsidian-icon-theme|:heavy_check_mark:|:heavy_check_mark:||:heavy_check_mark:|
|gnome-accessibility-themes|:heavy_check_mark:||:heavy_check_mark:||
|gnome-icon-theme|||:heavy_check_mark:|:heavy_check_mark:|
|mate-icon-theme|||:heavy_check_mark:|:heavy_check_mark:|
|adwaita-icon-theme|||:heavy_check_mark:||
|libelementary-data|||:heavy_check_mark:||
|mate-themes|:heavy_check_mark:||||
|numix-icon-theme||||:heavy_check_mark:|
|oxygen-icon-theme|||:heavy_check_mark:||
|breeze-icon-theme|||||

Note that this is not 100 % accurate as packages are inspected separately. For example, if `mate-themes` is installed, the `modem` and `network` icons *are* present as `mate-icon-theme` is a dependency.

Conclusion: Only elementary-xfce and the recent papirus provide everything we need and gnome and mate do not (anymore?). And most importantly: Icon themes are a pain... :unamused:

infirit:
Ok with adding elementry but..

Perhaps we should only rely on icon listed in https://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html? Then we can just refer to it and list a few examples. I don't see a good replacement for gnome-power-manager though..

We should definitely update ``network`` -> ``preferences-system-network`` as there is no icon for it in any theme I have.

Also we do not need to provide device icons as we can reliably use Device.Icon. Device.Icon is one of the standard ones listed in the spec.

cschramm:
I noticed that there's a check in configure.ac that would not work if neither mate-icon-theme nor gnome-icon-theme is present and it does not match the dependencies, so there's more to do regarding icons. :arrow_right: #976

PR #993, opened 2019-02-09
Small 'unknown' class bugfix

sezanzeb:
This bug was preventing the blueman-manager to start in my case because klass was a string. After this fix it shows 'Unknown' in the device list instead of crashing.

### DIFF
--- a/blueman/DeviceClass.py
+++ b/blueman/DeviceClass.py
@@ -232,6 +232,10 @@ gatt_appearance_categories = {
 
 
 def get_major_class(klass):
+    # class might be 'unknown'
+    if(type(klass) == str):
+        return klass
+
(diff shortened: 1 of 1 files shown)


infirit:
Which version is crashing? Because I introduced a bug in the stable version 2.0.7 which was fixed in 2.0.8, see commit ad7a67d3d528c232d41ae31554f9fc70c28839e2.

This bug was not present in master and the 2.1 alphas.

sezanzeb:
2.0.7-1

infirit:
Thanks, please upgrade to 2.0.8 which fixes this. Someone already marked version 2.0.7 as out of date on arch https://www.archlinux.org/packages/community/x86_64/blueman/.

PR #1008, opened 2019-04-23
Port DbusService to GLib

cschramm:


cschramm:
I will prepare a second PR that ports blueman.bluez.Agent and blueman.bluez.obex.Agent to blueman.main.DbusService as well.

cschramm:
...and fix Python 3.4 compatibility. :grimacing:

infirit:
I'll try it out later but some comment for discussion.

* The agents (Auth and Obex) have a predefined interfaces so I see no benefit to use DbusService over how it is done now.
* `/` is usually used for object managers so I was planning on using a nicer path like `/org/blueman/{applet,mechanism}` like in my old PR.
* I do not see anything related to dbus properties (which was what held my PR back). Did you skip this as it is not used if not are you planning on adding this?
* NetUsage is still using the old so it needs the below snippet or it needs to drop the related code using  PppStats (which was removed from NM's dbus api)

```python
from dbus.mainloop.glib import DBusGMainLoop
DBusGMainLoop(set_as_default=True)
```

I have tried three or four attempts to make it work with the implementation from the pygobject bug report / pithos but all attempts did not result in something I was willing to maintain. So this is much better already.

infirit:
Also it now assume we only export one interface, this is true atm. If this is to be implemented later (when required) a comment when registering the object to mention this would be nice.

cschramm:
This was a completely fresh take on the issue that we still depend on dbus-python which keeps us in an unsatisfying alpha-ish state. To be honest I neither looked at the previous work on this before I got started nor did I remember or re-read the issues exactly. :wink: This is just meant to be a migration from what we did with dbus-python to pretty much exactly the same thing with Glib.

> The agents (Auth and Obex) have a predefined interfaces so I see no benefit to use DbusService over how it is done now.

Well, right, didn't really think about it. It just saves some boilerplate (~ 100 lines) and redundant logic. A quickly rebased and completely untested commit is at https://github.com/cschramm/blueman/commit/20d3e4fdbbd17b7b0ecdd8ee461d4ac023e6d937.

> `/` is usually used for object managers so I was planning on using a nicer path like `/org/blueman/{applet,mechanism}` like in my old PR.

I have no strong opinion about that. Would be fine for me.

Surveying my system, there are quite some interfaces at `/`. The specification just says "Using an object path of / is allowed, but recommended against, as it makes versioning of interfaces hard". :man_shrugging: 

> I do not see anything related to dbus properties (which was what held my PR back). Did you skip this as it is not used if not are you planning on adding this?

No plans, I just completely ignored that topic and only ported what we already did before.

> NetUsage is still using the old so it needs the below snippet or it needs to drop the related code using  PppStats (which was removed from NM's dbus api)
> 
> 
> ```python
> from dbus.mainloop.glib import DBusGMainLoop
> DBusGMainLoop(set_as_default=True)
> ```

Right, that still needs its main loop... However, to get rid of dbus-python, we'll need to change that thing anyway and we use `NM-1.0.gir` for the NetworkManager stuff which seems to only be provided by NetworkManager 1.0 and newer which do not have that interface anymore. We can probably just drop the related code, yes.

> Also it now assume we only export one interface, this is true atm.

Yepp, that can easily get extended by adding an interface name argument to the methods that take a method or signal name (and removing it from the constructor).

infirit:
> > The agents (Auth and Obex) have a predefined interfaces so I see no benefit to use DbusService over how it is done now.
> 
> Well, right, didn't really think about it. It just saves some boilerplate (~ 100 lines) and redundant logic. A quickly rebased and completely untested commit is at [cschramm@20d3e4f](https://github.com/cschramm/blueman/commit/20d3e4fdbbd17b7b0ecdd8ee461d4ac023e6d937).

Feel free to add this here as you already started working on it.
 
> > `/` is usually used for object managers so I was planning on using a nicer path like `/org/blueman/{applet,mechanism}` like in my old PR.
> 
> I have no strong opinion about that. Would be fine for me.
> 
> Surveying my system, there are quite some interfaces at `/`. The specification just says "Using an object path of / is allowed, but recommended against, as it makes versioning of interfaces hard". 

Yes please, for some reason it annoys the hell out of me. When I do my actual review I'll add a comment.
 
> > I do not see anything related to dbus properties (which was what held my PR back). Did you skip this as it is not used if not are you planning on adding this?
> 
> No plans, I just completely ignored that topic and only ported what we already did before.

:+1: 
 
> > NetUsage is still using the old so it needs the below snippet or it needs to drop the related code using  PppStats (which was removed from NM's dbus api)
> > ```python
> > from dbus.mainloop.glib import DBusGMainLoop
> > DBusGMainLoop(set_as_default=True)
> > ```
> 
> Right, that still needs its main loop... However, to get rid of dbus-python, we'll need to change that thing anyway and we use `NM-1.0.gir` for the NetworkManager stuff which seems to only be provided by NetworkManager 1.0 and newer which do not have that interface anymore. We can probably just drop the related code, yes.

Approved the PR to remove it from NetUsage just now. I do like to open an feature request issue as NM does provide stats but only when we use NM to handle the connection. I just have not had the time to implement NM bits in NetUsage. But this is for 2.2, lets get this done first as its been too long we had a proper release.

I'll probably do a proper review and test this weekend.

blueman/main/DbusService.py
```diff
@@ -1,56 +1,132 @@
...
+            if signal_signature:
+                node_xml += "<arg type='%s'/>" % signal_signature
+            node_xml += "</signal>"
+        node_xml += "</interface></node>"
+
+        node_info = Gio.DBusNodeInfo.new_for_xml(node_xml)
+
+        regid = self._bus.register_object(
+            self._path,
+            node_info.interfaces[0],
```
infirit: Please add comment that this assumes only a single interface.

blueman/main/DbusService.py
```diff
@@ -1,56 +1,132 @@
...
+            if "sender" in options:
+                args += (sender,)
+            if "async" in options:
+                method(*(args + (ok, lambda exception: self._return_dbus_error(invocation, exception))))
+            else:
+                ok(method(*args))
+        except Exception as e:
+            self._return_dbus_error(invocation, e)
+
+    def _return_dbus_error(self, invocation, data):
```
infirit: Make static?

blueman/plugins/AppletPlugin.py
```diff
@@ -26,28 +24,38 @@ def __init__(self, parent):
 
         self.__overrides = []
 
+        self._dbus_service = parent.DbusSvc
+        self._dbus_methods = set()
+        self._dbus_signals = set()
```
infirit: Using ``set()``  is likely to cause a problem when adding a method/signal name more than once. On export it will result in an error so we better handle this. I suggest check for already existing methods/signals in ``DbusService``. And  in ``AppletPlugin`` catch the error from ``DbusService`` and raise ``ImportError`` for ``PluginManager``.

blueman/plugins/applet/DBusService.py
```diff
@@ -69,14 +61,12 @@ def reply(rfcomm):
                     pass
                 else:
                     logging.info("No handler registered")
-                    err(dbus.DBusException(
-                        "Service not supported\nPossibly the plugin that handles this service is not loaded"))
+                    err("Service not supported\nPossibly the plugin that handles this service is not loaded")
```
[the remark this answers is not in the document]
cschramm: The idea was to let `_handle_method_call` do that auto-magically. There's no reason to use a specific exception type here and `_handle_method_call` will just put the message into a `DbusError`.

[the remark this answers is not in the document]
infirit: Got it. But please make sure the error message is properly shown in the gui's like blueman-manager. See my other comment for when it doesn't  :smile:.

blueman/main/DbusService.py
```diff
@@ -1,56 +1,132 @@
...
+        if isinstance(data, DbusError):
+            name, message = data.name, data.message
+        elif isinstance(data, Exception):
+            tb = "".join(traceback.format_exception(type(data), data, data.__traceback__))
+            name, message = "org.blueman.Exception", "%s: %s\n%s" % (data.__class__.__name__, data, tb)
```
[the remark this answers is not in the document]
cschramm: Right, the traceback should not get added for our `BluezDBusException` classes.

blueman/main/DbusService.py
cschramm: Should be obvious from the method signatures as the constructor takes a (single) interface name while `add_method` and `add_signal` don't, no?

blueman/main/DbusService.py
```diff
@@ -1,56 +1,145 @@
...
+    def _return_dbus_error(invocation, data):
+        if isinstance(data, DbusError):
+            invocation.return_dbus_error(data.name, data.message)
+        else:
+            if isinstance(data, Exception):
+                et, ev, etb = sys.exc_info()
+                if ev is data:
+                    message = "".join(traceback.format_exception(et, ev, etb))
+                else:
+                    message = "".join(traceback.format_exception_only(data.__class__, data))
```
cschramm: This is taken from dbus-python (https://gitlab.freedesktop.org/dbus/dbus-python/blob/master/dbus/service.py#L285).

The message e.g. for our `DBusFailedError` thus matches the old one now (`blueman.bluez.errors.DBusFailedError:  Host is down`).

infirit:
Agents appear to work fine. I'll look at the new changes soon.

blueman/main/DbusService.py
infirit: It is now but will it in a year or two? Also it is very easy to handle it already but I'm not going to hold the PR for only this.

PR #1012, opened 2019-04-27
Handle invalid directory for incoming file transfers

### LINKED ISSUE #998: Error dialog after login when user language changed
blueman:
BlueZ:
Distribution: Ubuntu 18.04
Desktop environment: Mate
Related issues: #914, #496 

# Some facts:
Please correct me if I'm wrong:

- `blueman-applet` tries to load the default transfer directory at [plugins/applet/TransferService.py#L149](https://github.com/blueman-project/blueman/blob/master/blueman/plugins/applet/TransferService.py#L149).
- If `org.blueman.transfer.shared-path` does not exist, then `GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_DOWNLOAD)` is used and somehow that path is stored at `org.blueman.transfer.shared-path`.
- The directory pointed by `GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_DOWNLOAD)` depends on the system language, as given by `$XDG_DOWNLOAD_DIR` in `~/.config/user-dirs.dirs`. 
- After changing the language of my computer, on the first login Ubuntu (and I guess other distributions as well) asks me if I want to move my directories (e.g. from `~/Downloads` to `~/Baixades`, when switching from English to Catalan). And I choose yes.

# The issue:

If blueman has read `GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_DOWNLOAD)` and copied the path to `org.blueman.transfer.shared-path` with my user language in English, blueman will continue using `~/Downloads` even after changing the user language and having moved the Downloads directory. blueman asks me that I manually change the path (as mentioned in the related issues). I can manually change the path and I have done that.

However, I would expect that blueman does not store a copy of the value `GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_DOWNLOAD)` at `org.blueman.transfer.shared-path`, so whenever I change my user language blueman is able to automatically pick up the new Downloads directory through `GLib`.

Thanks for your time, I hope I have helped clarify the other related issues and provided enough information to fix the issue once and for all.

infirit:
Fixes #998

### DIFF
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -237,12 +237,12 @@ class TransferService(AppletPlugin):
         else:
             self._silent_transfers += 1
 
-    @staticmethod
-    def _add_open(n, name, path):
+    def _add_open(self, n, name, path):
         if n.actions_supported:
             logging.info("adding action")
 
             def on_open(*_args):
+                self._notification = None
                 logging.info("open")
                 launch("xdg-open", [path], True)
 
@@ -282,13 +282,13 @@ class TransferService(AppletPlugin):
             success = False
 
         if success:
-            n = Notification(_("File received"),
-                             _("File %(0)s from %(1)s successfully received") % {
-                                 "0": "<b>" + escape(filename) + "</b>",
-                                 "1": "<b>" + escape(attributes['name']) + "</b>"},
-                             **self._notify_kwargs)
-            self._add_open(n, "Open", dest)
-            n.show()
+            self._notification = Notification(_("File received"),
+                                              _("File %(0)s from %(1)s successfully received") % {
+                                                  "0": "<b>" + escape(filename) + "</b>",
+                                                  "1": "<b>" + escape(attributes['name']) + "</b>"},
+                                              **self._notify_kwargs)
+            self._add_open(self._notification, "Open", dest)
+            self._notification.show()
         elif not success:
             n = Notification(
                 _("Transfer failed"),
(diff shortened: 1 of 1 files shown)


cschramm:
Shouldn't we better ask the user for confirmation before resetting the configuration instead of telling him that we just did that? :smile:

zeehio:
At least for the case reported on #998 I would not expect any confirmation (no other software asks for confirmation in that case).

For other cases I don't know.

infirit:
I don't mind asking the user but the transfer will fail if they don't act. So instead I now have,

* A function construct the share path which should be valid, return tuple with path and if stored path was valid
* Use the new function instead of retrieving the path from gsettings
* Show a notification informing the user the stored path is invalid and what we use instead

Updated the commit for comment but is has not been tested yet (which i'll do later).

cschramm:
@zeehio: Right, doesn't make much sense for #998.

@infirit: :+1: I like the solution.

infirit:
@zeehio yeah, the problem I have to deal with is that all users up till now (which includes you) will have the full path stored in out settings. So I took a bit of a (too) drastic measure.

Your comment made me think. I can probably compare the stored path with our default and reset if they are the same. I am very tempted to add this :thinking:

infirit:
Added the reset when stored and default path are identical. Will merge in a couple of days unless someone has objections.

zeehio:
It is admirable that you are putting so much effort to address this properly. Thank you 🙂 and sorry for wasting your time 😅.

If on new installations everything works fine, then in the long term the problem will be solved for everyone and if I were you I would not put much more work into it, but of course do as you please :-)

In the GUI blueman-services, where I am able to choose the download path, would you think it is worth to give the option "Use the system default" vs "Specify a custom directory"? New users "Use the system default" while users coming from older versions would have a "custom directory" set, unless your path comparison trick has reset the setting.

Ideally, if the custom directory does not exist, the user would be warned and ideally he/she would be given the option to restore to the system default (informing what the default is) in the same dialog. In my case I did not even know what blueman-services was when I saw the message and I had to stop and think how to open blueman-services. If I had been given the option "restore to system default" I would have clicked it almost without thinking. Or at least "Click here to open blueman-services" would have felt more natural than "find blueman-services and open it".

If I think about the user who does not know how to open blueman-services, having a notification on each login about some invalid path is less annoying than an error message, because the notification disappears without having to click it. However, having a dialog only once that allows me to solve the issue ("click here to use the default setting") would be much better than the notification on every login.

In any case, thanks again for your work. This PR solves the issue and blueman-services were actually easy to find when I got tired of closing the error message (in my case I made the effort of looking for blueman-services after three days clicking once every day). 😅

blueman/plugins/applet/TransferService.py
```diff
@@ -171,6 +161,32 @@ def on_unload(self):
 
         self._unregister_agent()
 
+    def _make_share_path(self):
+        config_path = self._config["shared-path"]
+        default_path = GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_DOWNLOAD)
+        path = None
+        error = False
+
+        if config_path == '':
```
cschramm: Is this true for the initial configuration or will querying `shared-path` give us `None` instead? :thinking:

infirit: Yes, the default in the schema is an [empty string](https://github.com/blueman-project/blueman/blob/master/data/org.blueman.gschema.xml#L159).

infirit:
@zeehio no need to apologise and you are not. I added an action (think button) to the notification that will allow the user to reset. The notification should stay up for 30 seconds and then go away.

@cschramm also added a commit to fix the open action_cb on completed file transfers.

PR #1014, opened 2019-04-28
blueman-sendto: quit when no file was selected

infirit:


infirit:
Added a commit to deal with deprecation in pygobject

PR #1015, opened 2019-04-28
Set shebang to configured python interpreter

### LINKED ISSUE #887: Use distutils / entrypoints / console_scripts for apps
I have noticed that `blueman-applet` and other scripts from `apps` use a shebang `#!/usr/bin/env python` (or `#!/usr/bin/env python3` on master).

This causes errors like the following when running it from some virtualenv:
```
  File "/usr/bin/blueman-applet", line 12, in <module>
    import dbus
ModuleNotFoundError: No module named 'dbus'
```

By using `console_scripts` entry_points this would be avoided, since then a script would be generated to use the Python version where it was installed into, i.e. the system Python on my Arch Linux system.

On master the scripts in apps appear to be much smaller already, so maybe it's easy to use `console_scripts` from there.

This is a rather minor issue, but I've noticed it when running `blueman-applet` from my current terminal to see why it failed to start (bluetooth.service wasn't running, since I do not use it by default).

cschramm:
Closes #887

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,16 @@ Makefile
 /config.status
 /libtool
 /stamp-h1
+/apps/blueman-adapters
+/apps/blueman-applet
+/apps/blueman-assistant
+/apps/blueman-manager
+/apps/blueman-mechanism
+/apps/blueman-report
+/apps/blueman-rfcomm-watcher
+/apps/blueman-sendto
+/apps/blueman-services
+/apps/blueman-tray
 /blueman/Constants.py
 /data/configs/blueman-applet.service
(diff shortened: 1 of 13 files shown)


cschramm:
For the record: Distros don't like `env` at all, e.g. https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/2PD5RNJRKPN2DVTNGJSBHR5RUSVZSDZI/, but are no problem as they rewrite shebangs anyway. Users who compile on their own usually do not rewrite shebangs but might run into issues with `env` as well (#887).

PR #1017, opened 2019-04-29
Prepare 2.1.beta1

cschramm:
Requires #1008

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
+## 2.1.beta1
+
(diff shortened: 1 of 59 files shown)


infirit:
We changed quite a few strings since 2.0 and I am wondering if we ever
pushed these to transifex? I'll approve regardless when I get home tonight.

cschramm:
> I am wondering if we ever pushed these to transifex

Today. :grimacing: I think it was just five new strings, so I'm confident we'll get enough translations for a final release.

PR #1021, opened 2019-05-05
Ask user for auto-power-on setting

### LINKED ISSUE #877: Promote auto-power setting
The PowerManager plugin has a setting to automatically enable bluetooth as that is useful to some users. On the other hand every once in a while users are irritated by that and do not find the setting to disable it.

One idea to improve that: Ask the user about his preference on first start (maybe providing the way to get to the setting to change it later, but he will forget anyway :wink:).

Another idea: Move all settings to a single dialog. It's much easier to find one's way there instead of the PowerManager plugin settings. Regarding plugins I'm only counting four settings. Might make sense to join them with local services as well, though.

cschramm:
Closes #877

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -59,44 +59,50 @@ __all__ = ["check_bluetooth_status", "launch", "setup_icon_path", "get_icon",
            "have", "set_proc_title", "create_logger", "create_parser", "open_rfcomm"]
 
 
-def check_bluetooth_status(message, exitfunc, *args, **kwargs):
+def check_bluetooth_status(message, exitfunc):
     try:
         applet = AppletService()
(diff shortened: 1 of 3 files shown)


blueman/Functions.py
```diff
@@ -84,6 +86,18 @@ def check_bluetooth_status(message, exitfunc, *args, **kwargs):
                     print('Failed to enable bluetooth')
                     exitfunc()
 
+        config = Config("org.blueman.plugins.powermanager", None)
+        if config["auto-power-on"] is None:
+            d = Gtk.MessageDialog(None, type=Gtk.MessageType.ERROR)
```
infirit: You can pass properties like icon_name and text as keyword arguments.

PR #1030, opened 2019-05-20
Sendto fixes

infirit:
I still need to  pass the parent to a dialog but I need some help with an obex error I am seeing.

Steps to reproduce (after these fixes).

1. Start a largish transfer (so we get the opportunity to stop it on the receiver).
2. Accept the file on the receiver
3. Stop the transfer on the receiver
4. Choose cancel in sendto dialog

This will result in an error removing the session because of invalid argument. But I don't understand why as we pass a proper objectpath to ``RemoveSession``.. :confused: 

```
blueman-sendto 23.57.51 ERROR    Client:39 on_session_remove_failed: /org/bluez/obex/client/session8 org.bluez.obex.Error.InvalidArguments
```

### DIFF
--- a/apps/blueman-sendto.in
+++ b/apps/blueman-sendto.in
@@ -114,6 +114,7 @@ class SendTo:
         adapter_name = os.path.split(self.adapter_path)[-1]
         d = DeviceSelectorDialog(discover=False, adapter_name=adapter_name)
         resp = d.run()
+        d.close()
         if resp == Gtk.ResponseType.ACCEPT:
             if d.selection:
                 self.adapter_path, self.device = d.selection
--- a/blueman/bluez/AnyBase.py
+++ b/blueman/bluez/AnyBase.py
@@ -16,7 +16,7 @@ class AnyBase(GObject.GObject):
     __bus_interface_name = 'org.freedesktop.DBus.Properties'
 
     def __init__(self, interface_name):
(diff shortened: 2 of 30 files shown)


cschramm:
The cause for the error seems to be that while `/org/bluez/obex/client/session8` looks like a valid object path, that object does not exist anymore. https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/obexd/client/manager.c#n230 We must either check if the session object still exists before calling RemoveSession or just live with the resulting error for vanished ones.

infirit:
That makes sense (not the error) as the reciever disconnects right after I
reject. It would have been nice to error saying the object no longer exists
instead of invalid arguments.

I think as it now is fine then except that sendto keeps the dialog open
instead of closing. I'll add something for this as well.

Thx for the feedback.

infirit:
I done all I set out to do in this PR but.. retry and skip is broken when an error occurs and I get

``
org.freedesktop.DBus.Error.UnknownObject Method "SendFile" with signature "s" on interface "org.bluez.obex.ObjectPush1" doesn't exist
``

It looks like we can not reuse the session, did bluez4 allow this?

infirit:
Fixed the missing session issue by using obex.Manager to keep track of session creation and removal. Now skip and retry work for me but it will be good if you can do some testing as well.

PR #1033, opened 2019-05-23
Fix trust icon surface source

cschramm:
(Found this while playing with SonarCloud.) The calculation for x is bogus as it is always just a negated 1 / scale. I did not notice any difference (does the negative value fall back to 0?) but I think the intended value is a positive 1 / scale to align with the paired icon.


infirit:
If I remember correctly I only added the scale to the existing calculation
but I could be wrong. I'll check it out when I'm at home later today.

The idea is that with a scale of 2 there are twice as many pixels in the
source surface so we should adjust the position accordingly. In practice it
most likely will only be a pixel or two off so not noticeable.

PR #1047, opened 2019-06-10
update circleci

infirit:


cschramm:
Would probably make sense for 2-1-stable as well and Python 3.8 also for 2-0-stable.

infirit:
For 2.1 yes and openend PR. 2.0 is different and does not run flake8, just follows debian unstable.

cschramm:
Ah, sure, the old days... :smile:

PR #1054, opened 2019-06-23
Several fixes for adapter changes

infirit:


blueman/main/Manager.py
```diff
@@ -111,11 +111,16 @@ def on_dbus_name_appeared(_connection, name, owner):
...
-                bluez.Manager().get_adapter(self.Config['last-adapter'])
+                manager.get_adapter(self.Config['last-adapter'])
             except bluez.errors.DBusNoSuchAdapterError:
-                logging.error('No bluetooth adapter(s) found.')
-                exit(1)
+                logging.error('Default adapter not found, trying first available.')
+                try:
+                    manager.get_adapter(None)
+                except bluez.errors.DBusNoSuchAdapterError:
+                    logging.error('No adapyer(s) found, exiting')
```
cschramm: Typo

infirit:
06073878feffa928af299d38decab86681d89c9d there are two bugs. One is that set_adapter always emit signal with None. The other is that we should stop if we succeed setting the adapter with pattern. it doesn't and happily continue executing the rest of the function. The refactor now deal with this properly.

42deb53777b1554ca900a219a68247773f1cae84 on clear the monitor is still active and we may end up not adding a new monitor as its address may still be in the monitored_devices list. I just thought of a better way of dealing with this but using the object path instead of the address.

infirit:
Using the object path makes no difference so probably my conclusion is not correct. I'll dig deeper and see what the hell is going on.

ps: Have not forgotten about the typo, will fix :smile:

infirit:
I dropped the last commit about for the monitor as its not critical. unless you need more changes we can move this to master.

PR #1068, opened 2019-07-05
Several fixes

infirit:


infirit:
> What's the reason for [4751ab7](https://github.com/blueman-project/blueman/commit/4751ab7cab6971cc9b02112195190e51d0150d18)?

To be consistent as we use self.liststore already and it is the same model that self.get_model() returns. I'll update the commit message to clarify this.

PR #1078, opened 2019-07-22
Add minimum version for PyGObject

### LINKED ISSUE #1077: issue: Shall bluetooth get enabled automatically
Hi,

Brand new Linux Mint 19.1 install on Dell Precision M4800 laptop.
Internal Bluetooth mini-card disabled. External USB Bluetooth dongle used.

Default Blueberry is installed and that works perfectly. Connection to Bluetooth devices, file transfer, ... all works OK.
Needed for proprietary reasons more control than the simple Blueberry allows.
Removed Blueberry (sudo apt remove --purgeblueberry) and after reboot installed Blueman (sudo apt install blueman). Both operations happen without errors or warnings.
After reboot tried Blueman by clicking icon in taskbar. This brings a lot more options than available in Blueberry.
Top of pop-up menu says: Turn Bluetooth off --> Thus I expect it's running fine.
When clicking any other menu selection a small screen pop-up appears with the question:
 "Shall Bluetooth get enabled automatically?" [Yes] [No].
Clicking either button, yes or no, does not do anything else than close the pop-up.

Any idea what can be wrong?

I've checked if there were missing dependencies, but everything required is installed.
Blueberry works and Blueman doesn't. In most cases it's the way around.

Many thanks in advance for any hint or solution,

Regards,

Marc

cschramm:
Fixes #1077

### DIFF
--- a/Dependencies.md
+++ b/Dependencies.md
@@ -14,7 +14,7 @@
 * [Pango](http://www.gtk.org/) (GI bindings)
 * [pulseaudio](http://www.freedesktop.org/wiki/Software/PulseAudio/) (optional; its bluetooth module is required to actually use audio devices)
 * [pycairo](http://cairographics.org/pycairo/)
-* [PyGObject 3](https://wiki.gnome.org/PyGObject)
+* [PyGObject 3](https://wiki.gnome.org/PyGObject) (>= 3.27.2)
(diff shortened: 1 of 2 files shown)


cschramm:
#1079

PR #1084, opened 2019-08-04
Make codebase work with basic mypy usage

cschramm:
`mypy -p blueman --ignore-missing-imports` reports no findings with these changes (mypy 0.670 and 0.720).

### DIFF
--- a/blueman/Service.py
+++ b/blueman/Service.py
@@ -3,11 +3,11 @@ from blueman.Sdp import ServiceUUID
 
 
 class Service(object):
-    __group__ = None
-    __svclass_id__ = None
+    __group__: str
+    __svclass_id__: int
     __description__ = None
-    __icon__ = None
-    __priority__ = None
+    __icon__: str
+    __priority__: int
 
     def __init__(self, device, uuid):
         self.__device = device
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-from typing import Dict, Any, Tuple
+from typing import Dict, Tuple
 
 from gi.repository import Gio, GLib, GObject
 from gi.types import GObjectMeta
@@ -49,7 +49,7 @@ class Base(Gio.DBusProxy, metaclass=BaseMeta):
     __name = 'org.bluez'
     __bus_type = Gio.BusType.SYSTEM
 
-    __gsignals__: Dict[str, Tuple[Any, GObject.SignalFlags, Tuple]] = {
+    __gsignals__: Dict[str, Tuple[GObject.SignalFlags, None, Tuple]] = {
         'property-changed': (GObject.SignalFlags.NO_HOOKS, None,
                              (GObject.TYPE_PYOBJECT, GObject.TYPE_PYOBJECT, GObject.TYPE_PYOBJECT))
     }
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -1,4 +1,6 @@
 # coding=utf-8
+from typing import Dict, Any, Tuple
+
 from gi.repository import Gio, GLib, GObject
 from gi.types import GObjectMeta
 from blueman.bluez.errors import parse_dbus_error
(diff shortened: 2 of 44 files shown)


blueman/Service.py
```diff
@@ -3,11 +3,11 @@
...
 class Service(object):
-    __group__ = None
-    __svclass_id__ = None
+    __group__: str
+    __svclass_id__: int
     __description__ = None
-    __icon__ = None
-    __priority__ = None
+    __icon__: str
+    __priority__: int
```
infirit: ```Optional[str] = None```

blueman/main/NetworkManager.py
```diff
@@ -16,7 +16,7 @@ class NMConnectionError(Exception):
 
 
 class NMConnectionBase(object):
-    conntype = None
+    conntype: str
```
infirit: Same as before ```conntype: Optional[str] = None```

blueman/plugins/BasePlugin.py
```diff
@@ -1,29 +1,48 @@
...
-    __depends__ = []
-    __conflicts__ = []
+    __depends__: List[str] = []
+    __conflicts__: List[str] = []
     __priority__ = 0
 
-    __description__ = None
-    __author__ = None
+    __description__: str
+    __author__: str
```
infirit: ```Optional[str]```

blueman/plugins/BasePlugin.py
```diff
@@ -1,29 +1,48 @@
...
 
     __unloadable__ = True
     __autoload__ = True
 
     __instance__ = None
 
-    __gsettings__ = None
+    __gsettings__: "GSettings"
```
infirit: ```__gsettings__: Optional["Gsettings"] = None```

blueman/plugins/ServicePlugin.py
```diff
@@ -1,9 +1,10 @@
 # coding=utf-8
+from typing import List, Tuple
 
 
 class ServicePlugin(object):
-    instances = []
-    __plugin_info__ = None
+    instances: List["ServicePlugin"] = []
+    __plugin_info__: Tuple[str, str]
```
infirit: ```__plugin_info__: Optional[Tuple[str, str]] = None```

blueman/plugins/BasePlugin.py
```diff
@@ -1,29 +1,48 @@
...
+        default: Any
+
+    class Option(OptionBase, total=False):
+        name: str
+        desc: str
+        range: Tuple[int, int]
+
+    class GSettings(TypedDict):
+        schema: str
+        path: None
```
infirit: ```path: Optional[str]```

infirit:
Also do you have typehints for PyGobject? I tried [1] but they are incomplete so mypy is reporting lots of problems for me. [2] looks promising but is not merged.

[1] https://github.com/pygobject/pygobject-stubs
[2] https://github.com/pygobject/pgi-docgen/pull/176

blueman/bluez/Base.py
```diff
@@ -47,7 +49,7 @@ class Base(Gio.DBusProxy, metaclass=BaseMeta):
     __name = 'org.bluez'
     __bus_type = Gio.BusType.SYSTEM
 
-    __gsignals__ = {
+    __gsignals__: Dict[str, Tuple[Any, GObject.SignalFlags, Tuple]] = {
```
infirit: The first Tuple starts with GObject.SignalFlags not Any, next is None and then a tuple. For the second tuple we know the first is a str and we know what types to expect from dbus properties which we can represent as a Union. So I think it should be the below which works for Base.
```Dict[str, Tuple[GObject.SignalFlags, None, Tuple[str, Union[str, int, bool, List[str], Dict[str, GLib.Variant], Dict[int, GLib.Variant]], str]]]]```.

But not for subclasses that add their own signals so they will need their own type hints.

cschramm:
> variables not having a value assigned where before it would be None. Not assigning a value causes an AttributeError for me

Where do you get such an exception? The idea is that those properties are *not* optional, so they cannot be `None` and are just undefined in the abstract base class but defined in all concrete child classes. The latter might not hold everywhere but I did not run into any issues so far.

> Also do you have typehints for PyGobject?

Nope. Not much more than what you know. `PyGObject` is the main reason for relying on `--ignore-missing-imports` at the moment. IntelliJ generates some decent stubs but also recommends installing `PyGObject-stubs` and both aren't perfect.

infirit:
> Where do you get such an exception? The idea is that those properties are _not_ optional, so they cannot be `None` and are just undefined in the abstract base class but defined in all concrete child classes. The latter might not hold everywhere but I did not run into any issues so far.

Not in blueman as at the moment all subclasses define them before using. If these are not optional how are you going to enforce the plugin defining them without crashing? Or is the plugin crashing the way :smile:.

The Optional for the path in gsettings is really optional so it needs fixing.

PR #1087, opened 2019-08-10
Use Python types for GSignals and annotate them

cschramm:
Note that the type is too generic as "type" is actually only int, bool, str, float, or object and GObject.GType currently evaluates to Any


infirit:
typing.py was not added to Makefile.am/meson.build.

cschramm:
Oops. #1116 It does not belong anywhere in meson.build, does it? :thinking:

Issue #1092, opened 2019-08-20
blueman should take initial bluetooth mode from blutooth config file


h0tw4t3r:
blueman: 2.0.8-1
BlueZ: 5.50-1+b1
Distribution: ParrotOS/Debian
Desktop environment: KDE

Hi. Spent couple hours today on trying to get why blueman start's bluetooth on boot, though I have AutoEnable=false in `/etc/bluetooth/bluetooth.conf`. So far I understood that blueman has default:true for PowerManager plugin. I think it should get the AutoEnable property via dBus because it basically overrides this option, even if you have AutoEnable=false blueman with turned on PowerManager will register the device, which is actually not supposed to be.

I guess every other plugin also has the "default" values which actually have to be sent with bluez via dbus.

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

<!--

ℹ Please use some service like [Pastebin](http://pastebin.com/) or [GitHub Gist](https://gist.github.com/) to post your logs and keep the thread clean and readable. Make sure to describe the exact steps you took when producing it.

->

cschramm:
Hi, our PowerManager's auto-power-on feature is actually older than BlueZ' AutoEnable option (added in BlueZ 5.35) and was regularly requested by users whose systems did not set up auto-enable mechanism e.g. with udev rules. As our default behavior is a constant source for confusion, blueman 2.1 asks the user for his preference, so I think your actual issue should be fixed in 2.1.

If we'd respect BlueZ' AutoEnable option, our feature would not spring into action at all as it's a supplement for cases where no auto-enable mechanism like a udev rule or AutoEnable is in place.

What D-Bus interfaces do you refer to? I'm not aware of one that exposes such BlueZ options.

h0tw4t3r:
Weii even though it is a supplement, can't we respect the AutoEnable option if so exists?

Yes, I am reffering to non-existent D-Bus interface, or well at least I would like blueman to read /etc/bluetooth/bluetooth.conf file at the startup and get the AutoEnable feature (because it's not possible to change plugin configuration on fly).

Don't get me wrong but I find it a little dumb when there is an AutoEnable feature disabled, blueman doesn't respect it and simply turns on power devices because of PowerManager applet. I've seen a lot of topics opened over the internet which ask "Why AutoEnable does not work", and I personally spent around 3 hours to get that blueman doesn't care if you have this feature or not, it puts it's own.

[...]

PR #1094, opened 2019-08-21
PowerManager scan bluetooth config file

h0tw4t3r:
PowerManager applet has to scan for bluetooth main.conf file otherwise main.conf "AutoEnable" parameter simply gets overwritten by PowerManager auto-power default: True.

I am sorry I am not a python dev and they may be errors, I am a ParrotOS/Debian user and I get many errors when I launch blueman or reboot, so I am unable to run it from source.

### DIFF
--- a/blueman/plugins/applet/PowerManager.py
+++ b/blueman/plugins/applet/PowerManager.py
@@ -30,14 +30,12 @@ class PowerManager(AppletPlugin):
         }
     }
 
-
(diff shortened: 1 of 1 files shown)


cschramm:
See discussion in #1092. If we disable our auto-power feature when BlueZ' AutoEnable is not active, we could just remove it completely.

PR #1102, opened 2019-09-15
Update adapter state on powered property change

### LINKED ISSUE #1098: blueman-applet did not change state if bluetoothctl power off
blueman: blueman-git 2.1.alpha3.r5.g088da0c7-1
BlueZ: bluez-5.50-7
Distribution: Arch Linux
Desktop environment: Awesome WM

**Commands:**
_`$ bluetoothctl power on`
`$ blueman-applet &`
`$ bluetoothctl power off`_

**Expected:**
  blueman-applet will change its state to inactive

**Presently:**
  blueman-applet did not change its state to inactive

cschramm:
Fixes #1098

### DIFF
--- a/blueman/plugins/applet/PowerManager.py
+++ b/blueman/plugins/applet/PowerManager.py
@@ -190,6 +190,7 @@ class PowerManager(AppletPlugin):
                 logging.warning("adapter powered on while in off state, turning bluetooth on")
                 self.request_power_state(True)
 
+            self.adapter_state = self.get_adapter_state()
             self.update_power_state()
 
     def on_bluetooth_toggled(self):


infirit:
I hate the spaghetti code in PowerManager :sob: . The function ``set_adapter_state`` is responsible for this and it is called from ``Callback`` which in turn is run in ``request_power_state`` and we end up in the same function you manually set ``self.adapter_state``, ``on_adapter_property_changed``. Somewhere in this chain there is a bug :roll_eyes:

cschramm:
`on_adapter_property_changed` only calls `request_power_state` if an adapter gets powered on but not if it gets powered off.

infirit:
Right, hmm. I did the same as you 3 years ago and reverted it in 0fe55b6976575edb475e9890f6880b163b2332bd. But I don't remember why I reverted it but I do remember it had to do with some "state" going out of sync. 

Anyway, I'll approve and report back if it breaks something :smile:

PR #1103, opened 2019-09-16
Move D-Bus conf file to share/dbus-1/system.d

worldofpeace:
Since D-Bus 1.9.18 configuration files installed by third-party should
go in share/dbus-1/system.d. The old location is for sysadmin overrides.

### DIFF
--- a/Dependencies.md
+++ b/Dependencies.md
@@ -3,7 +3,7 @@
 ## Runtime dependencies
 
 * [BlueZ 5](http://www.bluez.org/)
-* [dbus](http://www.freedesktop.org/wiki/Software/dbus/)
+* [dbus](http://www.freedesktop.org/wiki/Software/dbus/) (>= 1.9.18)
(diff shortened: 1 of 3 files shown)


cschramm:
This technically raises the minimum version of dbus that we support to 1.9.18, correct? Can you please reflect that in Dependencies.md? Thanks!

worldofpeace:
> This technically raises the minimum version of dbus that we support to 1.9.18, correct? Can you please reflect that in Dependencies.md? Thanks!

Yes I can do this. Though it's possible that your minimum required dbus version is much higher than this.

worldofpeace:
> There is some unrelated white space changes. As its just the Makefile I don't care but make sure to avoid this in the future wink

Ah, I have my editor configured to strip trailing whitespace. So that is probably why it was there.
Thanks for merging @infirit :sparkles:

cschramm:
> Though it's possible that your minimum required dbus version is much higher than this.

:man_shrugging: :grin:

PR #1109, opened 2019-10-10
Update Popen annotation

cschramm:


cschramm:
Popen became generic in https://github.com/python/typeshed/commit/3ad3ed82c75b4fb82f71f5500b50f18a98f12f49#diff-ce63351cfc3a0791e309115212da484f and `--disallow-any-generics` requires that generics are used with their parameters.

PR #1121, opened 2019-10-17
PowerManager.py: provides a new dbus method

parr0tr1ver:
Provides a new dbus method for PowerManager that whether dbus method
SetBluetoothStatus() has finished. Otherwise GetBluetoothStatus() may
get a wrong value if called right after SetBluetoothStatus().

Signed-off-by: Kai Kang <kai.kang@windriver.com>

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -17,7 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-from time import sleep
+from time import sleep, time
 from typing import Optional, Dict, Tuple, List
 import re
 import os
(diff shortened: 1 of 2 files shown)


cschramm:
mypy errors are related to new mypy 740. I will look into that.

infirit:
I don't like blocking by using sleep from the time module. Also if you
don't want to use something like Glib.timeout_add you should make use of
the BluetoothStatusChanged signal.

You can connect this signal on the AppletService dbus proxy.

cschramm:
> mypy errors are related to new mypy 740. I will look into that.

#1122

parr0tr1ver:
@infirit I once tried with Glib.timeout_add in SetBluetoothStatus function request_power_state() but it doesn't block the function. So when call GetBluetoothStatus it still may get wrong status value.
I'll try with BluetoothStatusChanged signal.
Thanks.

cschramm:
@parr0tr1ver: Still interested in this?

PR #1122, opened 2019-10-17
Improve type annotations

cschramm:
mypy 0.740 improved --check-untyped-defs by giving self the correct type instead of Any and thus requires us to improve some annotations / usages

### DIFF
--- a/blueman/gui/DeviceList.py
+++ b/blueman/gui/DeviceList.py
@@ -134,7 +134,7 @@ class DeviceList(GenericList):
             self.emit("device-selected", dev, tree_iter)
 
     def _on_property_changed(self, _adapter, key, value, path):
-        if self.Adapter.get_object_path() != path:
+        if not self.Adapter or self.Adapter.get_object_path() != path:
             return
 
         if key == "Discovering":
@@ -190,6 +190,7 @@ class DeviceList(GenericList):
             logging.info("starting monitor")
             tree_iter = self.find_device(device)
 
+            assert self.Adapter is not None
             hci = os.path.basename(self.Adapter.get_object_path())
             cinfo = conn_info(bt_address, hci)
             try:
@@ -279,7 +280,7 @@ class DeviceList(GenericList):
 
     def add_device(self, device):
         # device belongs to another adapter
-        if not device['Adapter'] == self.Adapter.get_object_path():
+        if not self.Adapter or not device['Adapter'] == self.Adapter.get_object_path():
             return
 
         logging.info("adding new device")
@@ -297,9 +298,10 @@ class DeviceList(GenericList):
 
     def display_known_devices(self, autoselect=False):
         self.clear()
-        devices = self.manager.get_devices(self.Adapter.get_object_path())
-        for device in devices:
-            self.device_add_event(device)
+        if self.Adapter:
+            devices = self.manager.get_devices(self.Adapter.get_object_path())
+            for device in devices:
+                self.device_add_event(device)
 
         if autoselect:
             self.selection.select_path(0)
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -2,7 +2,7 @@
 import logging
 from gettext import gettext as _, bind_textdomain_codeset
 from operator import itemgetter
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Optional
 
 from blueman.Constants import UI_PATH
 from blueman.Functions import create_menuitem, e_
@@ -52,7 +52,7 @@ class ManagerDeviceMenu(Gtk.Menu):
         self._any_device.connect_signal('property-changed', self._on_service_property_changed)
 
         try:
-            self._appl = AppletService()
+            self._appl: Optional[AppletService] = AppletService()
         except DBusProxyFailed:
             logging.error("** Failed to connect to applet", exc_info=True)
             self._appl = None
@@ -172,6 +172,8 @@ class ManagerDeviceMenu(Gtk.Menu):
             prog.message(_("Success!"))
             MessageArea.close()
             self.unset_op(device)
+
+        assert self._appl
 
         if connect:
             self.set_op(self.SelectedDevice, _("Connecting..."))
--- a/blueman/main/DhcpClient.py
+++ b/blueman/main/DhcpClient.py
@@ -37,7 +37,6 @@ class DhcpClient(GObject.GObject):
             if path:
                 self._command = [path] + command[1:] + [self._interface]
                 break
-        self._client = None
 
     def run(self):
         if not self._command:
--- a/blueman/main/Manager.py
+++ b/blueman/main/Manager.py
@@ -210,8 +210,11 @@ class Blueman(Gtk.Window):
     def send(self, device, f=None):
         adapter = self.List.Adapter
 
+        assert adapter
+
         command = "blueman-sendto --source=%s --device=%s" % (adapter["Address"], device['Address'])
         launch(command, None, False, "blueman", _("File Sender"))
 
     def remove(self, device):
+        assert self.List.Adapter
         self.List.Adapter.remove_device(device)
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -294,16 +294,14 @@ class NetConf(object):
         self.dhcp_handler = None
         self.ipt_rules: List[Tuple[str, str, str]] = []
 
-        self.ip4_address = None
-        self.ip4_mask = None
         self.ip4_changed = False
 
     def set_ipv4(self, ip, netmask):
         if self.ip4_address != ip or self.ip4_mask != netmask:
             self.ip4_changed = True
 
-        self.ip4_address = ip
-        self.ip4_mask = netmask
+        self.ip4_address: str = ip
+        self.ip4_mask: str = netmask
 
     def get_ipv4(self):
         return self.ip4_address, self.ip4_mask
--- a/blueman/main/Sendto.py
+++ b/blueman/main/Sendto.py
@@ -77,7 +77,7 @@ class Sender(Gtk.Dialog):
         self.total_transferred = 0
 
         self._last_bytes = 0
-        self._last_update = 0
+        self._last_update = 0.0
 
         self.error_dialog = None
         self.cancelling = False
@@ -234,7 +234,8 @@ class Sender(Gtk.Dialog):
             d.add_button(_("Retry"), Gtk.ResponseType.YES)
             d.add_button("_Cancel", Gtk.ResponseType.CANCEL)
 
-            self.client.remove_session(self.object_push.get_object_path())
+            if self.object_push:
+                self.client.remove_session(self.object_push.get_object_path())
 
             def on_response(dialog, resp):
                 dialog.destroy()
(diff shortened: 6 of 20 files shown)


cschramm:
Needs fixes. :grimacing:

infirit:
I'll look closer this weekend but I think in DeviceList self.Adapter being
None is something we rely on in places when adapter is not powered for
example.

blueman/main/DhcpClient.py
```diff
@@ -37,7 +37,6 @@ def __init__(self, interface, timeout=30):
             if path:
                 self._command = [path] + command[1:] + [self._interface]
                 break
-        self._client = None
```
infirit: Pycharm does not like but flake8 does. But we do create an instance attribute outside of ``__init__``. If flake8 is happy then fine :weary:

blueman/main/Manager.py
```diff
@@ -209,8 +209,11 @@ def toggle_trust(self, device):
     def send(self, device, f=None):
         adapter = self.List.Adapter
 
+        assert adapter
```
infirit: We should just use self.List.adapter, having the ``adapter`` variable is a bit pointless. But why does the assert need to be here? We exited long before if we could not find an adapter.

blueman/main/Manager.py
```diff
@@ -209,8 +209,11 @@ def toggle_trust(self, device):
...
+
         command = "blueman-sendto --source=%s --device=%s" % (adapter["Address"], device['Address'])
         launch(command, None, False, "blueman", _("File Sender"))
 
     def remove(self, device):
+        assert self.List.Adapter
```
infirit: Same as above.

blueman/main/NetConf.py
```diff
@@ -294,16 +294,14 @@ def __init__(self):
         self.dhcp_handler = None
         self.ipt_rules: List[Tuple[str, str, str]] = []
 
-        self.ip4_address = None
-        self.ip4_mask = None
```
infirit: Same as before, Pycharm is not happy but flake8 is fine with it. Ok I guess.. :weary:

blueman/main/Sendto.py
```diff
@@ -232,7 +232,8 @@ def on_transfer_error(self, _transfer, msg=""):
             d.add_button(_("Retry"), Gtk.ResponseType.YES)
             d.add_button("_Cancel", Gtk.ResponseType.CANCEL)
 
-            self.client.remove_session(self.object_push.get_object_path())
+            if self.object_push:
+                self.client.remove_session(self.object_push.get_object_path())
```
infirit: ``self.object_push`` is created when a session is added. And the transfer can only error after the session is created. So why is there a need to check?

blueman/main/Sendto.py
```diff
@@ -271,9 +272,10 @@ def on_session_added(self, _manager, session_path):
 
     def on_session_removed(self, _manager, session_path):
         logging.debug('Session removed: %s' % session_path)
-        self.object_push.disconnect_by_func(self.on_transfer_started)
-        self.object_push.disconnect_by_func(self.on_transfer_failed)
-        self.object_push = None
+        if self.object_push:
```
infirit: We never have a session removed without a session being added, why is there a need to check?

blueman/plugins/applet/KillSwitch.py
```diff
@@ -91,6 +91,7 @@ def io_event(self, _, condition):
         if condition & GLib.IO_ERR or condition & GLib.IO_HUP:
             return False
 
+        assert self._fd is not None
```
infirit: Why, if os.open in on_load fails we crash?

blueman/plugins/applet/NetUsage.py
```diff
@@ -207,8 +205,8 @@ def update_time(self):
             delta = datetime.datetime.now() - self.datetime
 
             d = ngettext("day", "days", delta.days)
-            h = ngettext("hour", "hours", delta.seconds / 3600)
-            m = ngettext("minute", "minutes", delta.seconds % 3600 / 60)
+            h = ngettext("hour", "hours", delta.seconds // 3600)
+            m = ngettext("minute", "minutes", delta.seconds % 3600 // 60)
```
infirit: You are changing to floor division here, this needs imo its own commit with explanation.

blueman/plugins/applet/TransferService.py
```diff
@@ -66,6 +66,7 @@ def on_action(action):
             logging.info("Action %s" % action)
 
             if action == "accept":
+                assert self._pending_transfer
```
infirit: Why, self._pending_transfer is created in the parent function before creating the notification.

blueman/plugins/applet/TransferService.py
```diff
@@ -131,7 +132,8 @@ def on_action(action):
             on_action("accept")
 
     def _cancel(self):
-        self._notification.close()
+        if self._notification:
+            self._notification.close()
```
infirit: I don't care about this one but I am wondering if it is even possible to have ``Cancel`` method being called before ``AuthorizePush``..

blueman/plugins/applet/TransferService.py
```diff
@@ -140,8 +142,6 @@ class TransferService(AppletPlugin):
     __description__ = _("Provides OBEX file transfer capabilities")
     __icon__ = "blueman-send-file"
 
-    _config = None
-
```
infirit: Similar as before Pycharm warn but flake8 is ok with it.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, blueman):
         self._any_device.connect_signal('property-changed', self._on_service_property_changed)
 
         try:
-            self._appl = AppletService()
+            self._appl: Optional[AppletService] = AppletService()
```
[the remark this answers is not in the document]
cschramm: In case of an exception, we set it to `None` and actually check that before all usages.

blueman/gui/DeviceList.py
```diff
@@ -188,6 +188,7 @@ def update(row_ref, cinfo, address):
             logging.info("starting monitor")
             tree_iter = self.find_device(device)
 
+            assert self.Adapter is not None
```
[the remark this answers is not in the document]
cschramm: Depends on which level you look at the logic. You can only state that if you assume certain (reasonable) properties of third-party mechanics. From a quick view, `monitor_power_levels` could e.g. get called while `self.Adapter is None` if we receive a `device-created` event before we receive the respective `adapter-added` event. So if BlueZ, D-Bus or something misbehaves we actually crash here due to unexpected state.

blueman/main/Manager.py
cschramm: `self.List.Adapter` might be `None` so that the attribute getters used below would not be available and crash. `assert` just moves that crash here but also tells mypy that the value is not treated optional after this point.

blueman/plugins/applet/KillSwitch.py
cschramm: Hm, right, we should probably just remove `_fd = None` from this class so that it's not optional. :thinking:

blueman/gui/DeviceList.py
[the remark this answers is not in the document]
infirit: It won't hurt to check so ok with me.

blueman/gui/DeviceList.py
```diff
@@ -295,9 +296,10 @@ def add_device(self, device):
 
     def display_known_devices(self, autoselect=False):
         self.clear()
-        devices = self.manager.get_devices(self.Adapter.get_object_path())
-        for device in devices:
-            self.device_add_event(device)
+        if self.Adapter:
```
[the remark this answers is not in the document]
infirit: Similar as before, checking will not hurt.

blueman/gui/manager/ManagerDeviceMenu.py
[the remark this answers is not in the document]
infirit: I do wonder if it makes any sense to set this to None as we pretty much can't do anything in blueman-manager. But I get why this is Optional. Still think we should reuse ``blueman.Applet``, but I don't care enough.

blueman/main/Sendto.py
infirit: Like before, the check won't hurt but I still wonder why mypy can't figure this out.

cschramm: You cannot derive the expected state from our code as this is triggered by a D-Bus event which could theoretically happen at any time, even if `self.object_push` is `None`, i.e. before a `session-added` signal got processed or after a `session-removed` signal.

blueman/plugins/applet/KillSwitch.py
infirit:
```
iochan = GLib.IOChannel.new_file("/dev/rfkill", "r")
self._iom = GLib.io_add_watch(iochan, GLib.IO_IN | GLib.IO_ERR | GLib.IO_HUP, self.io_event)
```

The IOChannel documentation states that the file will be closed when all references to it are dropped. And then in on_unload when the watch is removed the last reference should dropped.

Weird GLib.io_watch_add accepted just a file descriptor, it must convert it to a IOChannel in the background.

infirit: Fully working patch below, tested with my usb dongles.

```patch
diff --git a/blueman/plugins/applet/KillSwitch.py b/blueman/plugins/applet/KillSwitch.py
index c8b3e2c2..1052247a 100644
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -51,7 +51,6 @@ class KillSwitch(AppletPlugin):
 
     _switches: Dict[int, Switch] = {}
     _iom = None
-    _fd = None
     _enabled = True
     _hardblocked = False
 
@@ -60,17 +59,17 @@ class KillSwitch(AppletPlugin):
         self._connman_watch_id = Gio.bus_watch_name(Gio.BusType.SYSTEM, "net.connman", Gio.BusNameWatcherFlags.NONE,
                                                     self._on_connman_appeared, self._on_connman_vanished)
 
-        self._fd = os.open('/dev/rfkill', os.O_RDONLY | os.O_NONBLOCK)
+        channel = GLib.IOChannel.new_file("/dev/rfkill", "r")
+        if channel is None:
+            raise ImportError('Failed to open rfkill device')
 
-        self._iom = GLib.io_add_watch(self._fd, GLib.IO_IN | GLib.IO_ERR | GLib.IO_HUP, self.io_event)
+        self._iom = GLib.io_add_watch(channel, GLib.IO_IN | GLib.IO_ERR | GLib.IO_HUP, self.io_event)
 
     def on_unload(self):
         Gio.bus_unwatch_name(self._connman_watch_id)
         self._connman_proxy = None
         if self._iom:
             GLib.source_remove(self._iom)
-        if self._fd:
-            os.close(self._fd)
 
     def _on_connman_appeared(self, connection, name, owner):
         logging.info("%s appeared" % name)
@@ -87,11 +86,11 @@ class KillSwitch(AppletPlugin):
         logging.info("%s vanished" % name)
         self._connman_proxy = None
 
-    def io_event(self, _, condition):
+    def io_event(self, channel, condition):
         if condition & GLib.IO_ERR or condition & GLib.IO_HUP:
             return False
 
-        data = os.read(self._fd, RFKILL_EVENT_SIZE_V1)
+        data = channel.read(RFKILL_EVENT_SIZE_V1)
         if len(data) != RFKILL_EVENT_SIZE_V1:
             logging.warning("Bad rfkill event size")
             return True
```

cschramm:
Rebased, applied your IOChannel patch (first commit), extracted float division fix (second commit), resolved conflicts.

blueman/plugins/applet/TransferService.py
cschramm: I give you that one. That's rather obvious from the code as `on_action` does not get referenced before `_pending_transfer` gets set but mypy does not know. There's even one small exception though: `__init__` might get called on the already initialized object so that `_pending_transfer` would get reset to `None`.

PR #1123, opened 2019-10-18
Add types to blueman.services

cschramm:
Depends on #1122

### DIFF
--- a/blueman/Service.py
+++ b/blueman/Service.py
@@ -3,7 +3,6 @@ from blueman.Sdp import ServiceUUID
 
 
 class Service(object):
-    __group__: str
     __svclass_id__: int
     __description__ = None
     __icon__: str
@@ -40,7 +39,3 @@ class Service(object):
     @property
     def priority(self):
         return self.__priority__
-
-    @property
-    def group(self):
-        return self.__group__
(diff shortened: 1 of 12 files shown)


infirit:
There is more than just improving type information in these commits. There is added logic, cast (why ??) and removal of what I assume are unused/unneeded code? Can this be split as it is not clear why a lot of these changes are needed and it needs an explanation imo.

cschramm:
I will split it next week.

The cast is just a type hint "converting" `Any` (mostly due to imports without typing) to the expected types. The actual runtime implementation does nothing.
I can do it without `cast` and use a type variable instead. `return cast(T, expr)` -> `v: T = expr; return v`. That way the type does not get enforced and would still get checked. We could probably use Python 3.8's assignment expressions to make that a one-liner again, but I don't think it's a good idea to require 3.8.

Removed code was unused, yes.

infirit:
The cast is wrong then as we should instead annotate ``__getitem__`` in bluez.Base and account for all the types it can return (which I think you defined somewhere already).

cschramm:
Rebased, split into three commits, replaced the cast usages for `Any` types with typed variables. :slightly_smiling_face:

infirit:
Cool, thanks. Assuming the asserts are for thing that got pointed out by mypy so lgtm.

PR #1125, opened 2019-10-18
Add types to python files in blueman root

cschramm:


blueman/Functions.py
```diff
@@ -131,20 +141,20 @@ def launch(cmd, paths=None, system=False, icon_name=None, sn=True, name="blueman
...
 
 
-def setup_icon_path():
+def setup_icon_path() -> None:
     ic = Gtk.IconTheme.get_default()
     ic.prepend_search_path(ICON_PATH)
 
 
-def get_icon(name, size=24, fallback="image-missing"):
+def get_icon(name: str, size: int = 24, fallback: str = "image-missing") -> GdkPixbuf.Pixbuf:
```
infirit: Unrelated but I immediately started wondering if we actually use it. We do but only in get_notification_icon which is also unused. I'll open a PR to remove them later.

blueman/Functions.py
```diff
@@ -170,32 +180,33 @@ def get_icon(name, size=24, fallback="image-missing"):
...
+def e_(msg: Union[str, Exception]) -> Tuple[str, Optional[str]]:
     if isinstance(msg, Exception):
         return str(msg), traceback.format_exc()
     else:
         s = msg.strip().split(": ")[-1]
         return s, None
 
 
-def opacify_pixbuf(pixbuf, alpha):
+def opacify_pixbuf(pixbuf: GdkPixbuf.Pixbuf, alpha: int) -> GdkPixbuf.Pixbuf:
```
infirit: Some as earlier, unused.

blueman/Functions.py
```diff
@@ -205,7 +216,10 @@ def opacify_pixbuf(pixbuf, alpha):
 
 # pixbuf, [(pixbuf, x, y, alpha), (pixbuf, x, y, alpha)]
 
-def composite_icon(target, sources):
+def composite_icon(
```
infirit: Whee, another one unused.

blueman/Sdp.py
```diff
@@ -365,17 +366,19 @@
...
         if self.reserved:
             return self.int >> 96 & 0xFFFF
+        else:
+            return None
 
     @property
-    def name(self):
-        if self.reserved:
+    def name(self) -> str:
+        if self.short_uuid:
```
infirit: reserved -> short_uuid? If this is a bug please do a separate commit if typo please fix :smile:

blueman/main/NetConf.py
```diff
@@ -52,7 +52,9 @@ def do_apply(self):
                 self.do_remove()
 
             ipiface = ipaddress.ip_interface('/'.join((self.netconf.ip4_address, '255.255.255.0')))
-            cmd = [have("dnsmasq"), "--port=0", "--pid-file=/var/run/dnsmasq.pan1.pid", "--except-interface=lo",
+            dnsmasq = have("dnsmasq")
+            assert dnsmasq is not None
```
infirit: Hmm, I agree this could be a problem but I prefer to raise an exception and have it bubble up to the UI. Would you object to instead explicitly check for None and ``RuntimeError('Could not find dnsmasq')`` if not found? Same applies to dhcpcd of course.

blueman/main/NetConf.py
```diff
@@ -153,7 +155,9 @@ def do_apply(self):
                 f.write(dhcp_config)
                 f.write(subnet)
 
-            cmd = [have("dhcpd3") or have("dhcpd"), "-pf", "/var/run/dhcpd.pan1.pid", "pan1"]
+            dhcpd = have("dhcpd3") or have("dhcpd")
+            assert dhcpd is not None
```
infirit: See above for dnsmasq

blueman/main/NetConf.py
```diff
@@ -221,7 +225,9 @@ def do_apply(self):
             os.close(config_file)
 
             logging.info("Running udhcpd with config file %s" % config_path)
-            cmd = [have("udhcpd"), "-S", config_path]
+            udhcpd = have("udhcpd")
+            assert udhcpd is not None
```
infirit: See above for dnsmasq

blueman/main/NetConf.py
cschramm: Fine for me. We could raise that exception from `have` and make it return non-optional `str`.

blueman/Sdp.py
cschramm: This is a trick to avoid a type error two lines below.

If `reserved` is `True` then `short_uuid` is an `int`, if it's `False` then it's `None`. So both basically express the same thing (neglecting the "impossible" edge case `uuid == 0`).

However, for the type checker it makes a difference: If we use `reserved` here, `short_uuid` two line above is of type `Optional[int]`. If we use `short_uuid` here, we know that it's not `None`, so it is of type `int`.

cschramm:
Rebased and improved the error handling in `NetConf` in a separate commit. Removal of unused functions is in #1139.

PR #1128, opened 2019-10-22
Stub for _blueman binary module

infirit:
I created it some time ago when we first started the discussion. Fixed everything up and made mypy take notice. It immediately pointed out that I forgot to add errno to the BridgeException :balloon:


infirit:
One question though, do we want to install this stub and any future ones we may add?

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
+from typing import Any, List, Dict, Union, Callable
+from mypy_extensions import TypedDict
```
cschramm: Should be `typing_extensions` to stay consistent.

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+    errno: int
+    def __init__(self, errno: int) -> None: ...
+
+class ConnInfoReadError(Exception): ...
+
+class RFCOMMError(Exception): ...
+
+class conn_info:
+    failed: bool
+    def __init__(self, addr: str, hci_name: str) -> None: ...
```
cschramm: Regarding the implementation, `hci_name` is optional as it has a default value.

infirit:
Last comment for today, promise :smiling_imp:. I did not look at the places _blueman is used because you have PR open that touch the same code. When those are merged I'll have a look and add/update annotations.

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+    def get_rssi(self) -> int: ...
+    def get_tpl(self) -> int: ...
+    def init(self) -> None: ...
+
+def create_bridge(name: str) -> None: ...
+def create_rfcomm_device(local_address: str, remote_address: str, channel: int) -> int: ...
+def destroy_bridge(name: str) -> None: ...
+def device_info(hci_name: str) -> HciInfo: ...
+def page_timeout(hci_name: str) -> int: ...
+def get_rfcomm_channel(uuid: str, bdaddr: str) -> int: ...
```
cschramm: Might return `None`, so it's `Optional[int]`.

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+class conn_info:
+    failed: bool
+    def __init__(self, addr: str, hci_name: str) -> None: ...
+    def deinit(self) -> None: ...
+    def get_lq(self) -> int: ...
+    def get_rssi(self) -> int: ...
+    def get_tpl(self) -> int: ...
+    def init(self) -> None: ...
+
+def create_bridge(name: str) -> None: ...
```
cschramm: optional

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+def create_rfcomm_device(local_address: str, remote_address: str, channel: int) -> int: ...
+def destroy_bridge(name: str) -> None: ...
```
cschramm: optional

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+def device_info(hci_name: str) -> HciInfo: ...
```
cschramm: optional

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+def page_timeout(hci_name: str) -> int: ...
```
cschramm: Argument is optional. Return type is `float`, no?

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+    id: int
+    channel: int
+    flags: int
+    state: str
+    src: str
+    dst: str
+
+class HciInfo(TypedDict):
+    acl_mtu: int
+    acl_plts: int
```
cschramm:
```suggestion
    acl_pkts: int
```

stubs/_blueman.pyi
```diff
@@ -0,0 +1,58 @@
...
+from mypy_extensions import TypedDict
+
+ERR: Dict[int, str]
+RFCOMM_HANGUP_NOW: int
+RFCOMM_RELEASE_ONHUP: int
+RFCOMM_REUSE_DLC: int
+RFCOMM_STATES: List[str]
+RFCOMM_TTY_ATTACHED: int
+
+class RfcommDev(TypedDict):
```
cschramm: Just a side note: If we define this here, mypy thinks it exists in the module and would accept code like `from _blueman import RfcommDev` although it will not actually work. Not sure if there's a way to avoid that, though.

stubs/_blueman.pyi
infirit: Will fix

stubs/_blueman.pyi
infirit: Will fix.

stubs/_blueman.pyi
infirit: Hmm, ``Optional[X]`` is the same as ``Union[X, None]`` but we do not allow None. But annotation with ``Union[str, str]`` does not look right to me.

stubs/_blueman.pyi
infirit: Not sure, the c function will always return ``port_num`` which defaults to zero. And if for some reason we try to pass ``None`` as bdaddr we are sure to raise an exception encoding it to UTF8. Am I missing something?

stubs/_blueman.pyi
infirit: See comment earlier, Union[str, str] does not make sense to me.

stubs/_blueman.pyi
infirit: Correct on float will fix that. Optional see earlier comment.

stubs/_blueman.pyi
infirit: I'll check if I can hide it somehow, until then I can prefix it with and underscore.

stubs/_blueman.pyi
cschramm: I think you're right, the `if bdaddr is not None:` in that function just does not make sense as it's always the case...

stubs/_blueman.pyi
infirit: Ok, I see I did not account for the default value in the annotation which I'll add for these functions.

infirit:
Forced pushed the commits with the requested changes.

Installing makes sense if other people are using _blueman but I don't know of anyone that does. I don't see a need now so let's wait for someone to ask for them and take it from there :smiley:

stubs/_blueman.pyi
infirit: Yeah, I think I am responsible for the not making sense part when I added python3 support (bytes vs codepoints). I opened #1129 as a reminder for myself to fix this.

PR #1139, opened 2019-11-02
Drop unused functions in Functions.py

cschramm:
Mentioned by @infirit in #1125 and double checked with [vulture](https://github.com/jendrikseipp/vulture)

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -53,8 +53,7 @@ from gi.repository import GLib
 from gi.repository import Gio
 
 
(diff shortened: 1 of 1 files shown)


blueman/Functions.py
```diff
@@ -45,13 +45,11 @@
 gi.require_version("Gdk", "3.0")
 from gi.repository import Gtk
 from gi.repository import Gdk
-from gi.repository import GdkPixbuf
```
infirit: I just noticed that this is still used in ``create_menuitem`` so please back this out before merging.

cschramm:
Rebased (`GdkPixbuf` is back :tada:)

PR #1141, opened 2019-11-02
Refactor singleton GObjects

cschramm:
We define multiple GObjectMeta-based classes to implement the same argument-independent singleton pattern. This is now combined into a single SingletonGObjectMeta class.

### DIFF
--- a/blueman/Makefile.am
+++ b/blueman/Makefile.am
@@ -12,6 +12,7 @@ blueman_PYTHON = 	\
 	Functions.py	\
 	Sdp.py		\
 	Service.py	\
+	gobject.py	\
 	typing.py	\
 	__init__.py
 
--- a/blueman/bluez/Manager.py
+++ b/blueman/bluez/Manager.py
@@ -3,26 +3,15 @@ import logging
 from typing import List, Optional, Callable
 
 from gi.repository import GObject, Gio
-from gi.types import GObjectMeta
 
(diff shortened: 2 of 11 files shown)


cschramm:
> If a new file is created why stop at just these. We should then put BaseMeta in gobject.py as well.

`BaseMeta` is rather specific to our classes and only used once while `SingletonGObjectMeta` combines the common `GObjectMeta` with a singleton pattern that would ignore any given `__init__` arguments and is used all over.

We could actually combine both into a common meta class that disallows positional arguments and stores and reuses instances based on all given keyword arguments (not as straightforward as it sounds as `dict`s are not hashable, so we'd have to build some key from `kwargs`) but I thought it'd be better to implement the simple common case without arguments here and keep the specific case for our `bluez.Base` there. No strong opinion though.

> More importantly, Makefile.am feels left out smiley

:grimacing: :see_no_evil:

voronaivan93:
g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.331 was not provided by any .service files (2)

voronaivan93:
g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.331 was not provided by any .service files (2)
[Uploading md5sum.txt…]()

PR #1142, opened 2019-11-02
Fix issue from #1124

cschramm:

PR #1143, opened 2019-11-03
Simplify BaseMeta

infirit:
Needs some proper testing but this looks to me like the best solution to have properly type-checked code. Supersedes  #1142

* Only allow keyword arguments
* Only use obj_path keyword
* Still fallback on the class _interface_name
* Fixes Optional mypy error for obj_path and interface_name

### DIFF
--- a/blueman/bluez/Adapter.py
+++ b/blueman/bluez/Adapter.py
@@ -10,7 +10,7 @@ class Adapter(Base):
     _interface_name = 'org.bluez.Adapter1'
 
     def __init__(self, obj_path: str):
-        super().__init__(self._interface_name, obj_path=obj_path)
+        super().__init__(interface_name=self._interface_name, obj_path=obj_path)
 
     def start_discovery(self) -> None:
         self._call('StartDiscovery')
--- a/blueman/bluez/Adapter.py
+++ b/blueman/bluez/Adapter.py
@@ -10,7 +10,7 @@ class Adapter(Base):
     _interface_name = 'org.bluez.Adapter1'
 
     def __init__(self, obj_path: str):
-        super().__init__(interface_name=self._interface_name, obj_path=obj_path)
+        super().__init__(obj_path=obj_path)
 
     def start_discovery(self) -> None:
         self._call('StartDiscovery')
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -12,18 +12,17 @@ from blueman.typing import GSignals
 class BaseMeta(GObjectMeta):
     def __call__(cls, **kwargs: str) -> "Base":
         if not hasattr(cls, "__instances__"):
-            cls.__instances__: Dict[str, Dict[str, "Base"]] = {}
+            cls.__instances__: Dict[str, "Base"] = {}
 
         path = kwargs.get('obj_path')
         if path is None:
             path = cls._obj_path
 
-        if cls._interface_name in cls.__instances__:
-            if path in cls.__instances__[cls._interface_name]:
-                return cls.__instances__[cls._interface_name][path]
+        if path in cls.__instances__:
+            return cls.__instances__[path]
 
         instance: "Base" = super().__call__(**kwargs)
-        cls.__instances__[cls._interface_name] = {path: instance}
+        cls.__instances__[path] = instance
 
         return instance
 
@@ -38,7 +37,7 @@ class Base(Gio.DBusProxy, metaclass=BaseMeta):
     __gsignals__: GSignals = {
         'property-changed': (GObject.SignalFlags.NO_HOOKS, None, (str, object, str))
     }
-    __instances__: Dict[str, Dict[str, "Base"]]
+    __instances__: Dict[str, "Base"]
 
     def __init__(self, obj_path: str):
         super().__init__(
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -10,33 +10,24 @@ from blueman.typing import GSignals
 
 
 class BaseMeta(GObjectMeta):
-    def __call__(cls, *args: str, **kwargs: str) -> "Base":
-        instances: Dict[str, Dict[str, "Base"]] = cls.__dict__.setdefault("__instances__", {})
+    def __call__(cls, **kwargs: str) -> "Base":
+        if not hasattr(cls, "__instances__"):
+            cls.__instances__: Dict[str, Dict[str, "Base"]] = {}
 
-        path = None
-        interface_name = None
-
-        if kwargs:
-            interface_name = kwargs.get('interface_name')
-            path = kwargs.get('obj_path')
-
-        if args:
-            for arg in args:
-                if args is None:
-                    continue
-                elif '/' in arg:
-                    path = arg
-                elif '.' in arg:
-                    interface_name = arg
-
-        if not interface_name:
+        # cls._interface_name or cls._obj_path need to exist to act as fallback, they may not
+        interface_name = kwargs.get('interface_name')
+        if interface_name is None:
             interface_name = cls._interface_name
 
-        if interface_name in instances:
-            if path in instances[interface_name]:
-                return instances[interface_name][path]
+        path = kwargs.get('obj_path')
+        if path is None:
+            path = cls._obj_path
 
-        instance: "Base" = super().__call__(*args, **kwargs)
+        if interface_name in cls.__instances__:
+            if path in cls.__instances__[interface_name]:
+                return cls.__instances__[interface_name][path]
+
+        instance: "Base" = super().__call__(**kwargs)
         cls.__instances__[interface_name] = {path: instance}
 
         return instance
@@ -52,6 +43,7 @@ class Base(Gio.DBusProxy, metaclass=BaseMeta):
     __gsignals__: GSignals = {
         'property-changed': (GObject.SignalFlags.NO_HOOKS, None, (str, object, str))
     }
+    __instances__: Dict[str, Dict[str, "Base"]]
 
     def __init__(self, interface_name: str, obj_path: str):
         super().__init__(
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -14,21 +14,16 @@ class BaseMeta(GObjectMeta):
         if not hasattr(cls, "__instances__"):
             cls.__instances__: Dict[str, Dict[str, "Base"]] = {}
 
-        # cls._interface_name or cls._obj_path need to exist to act as fallback, they may not
-        interface_name = kwargs.get('interface_name')
-        if interface_name is None:
-            interface_name = cls._interface_name
-
         path = kwargs.get('obj_path')
         if path is None:
             path = cls._obj_path
 
-        if interface_name in cls.__instances__:
-            if path in cls.__instances__[interface_name]:
-                return cls.__instances__[interface_name][path]
+        if cls._interface_name in cls.__instances__:
+            if path in cls.__instances__[cls._interface_name]:
+                return cls.__instances__[cls._interface_name][path]
 
         instance: "Base" = super().__call__(**kwargs)
-        cls.__instances__[interface_name] = {path: instance}
+        cls.__instances__[cls._interface_name] = {path: instance}
 
         return instance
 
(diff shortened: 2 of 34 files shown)


blueman/bluez/Adapter.py
```diff
@@ -10,7 +10,7 @@ class Adapter(Base):
     _interface_name = 'org.bluez.Adapter1'
 
     def __init__(self, obj_path: str):
-        super().__init__(self._interface_name, obj_path=obj_path)
+        super().__init__(interface_name=self._interface_name, obj_path=obj_path)
```
cschramm: There's no real point in this as `blueman.bluez.Base.BaseMeta.__call__` will receive the arguments of the call to `blueman.bluez.Adapter.Adapter.__init__` not to `blueman.bluez.Base.Base.__init__`.

If we want to enforce keyword arguments for the base class, `blueman.bluez.Base.Base.__init__` could be `__init__(self, *, interface_name: str, obj_path: str)` ([PEP 3102](https://www.python.org/dev/peps/pep-3102/))

blueman/bluez/Base.py
```diff
@@ -11,30 +11,24 @@
...
-        if not interface_name:
+        # cls._interface_name or cls._obj_path need to exist to act as fallback, they may not
+        interface_name = kwargs.get('interface_name', cls._interface_name)
+        if interface_name is None:
             interface_name = cls._interface_name
 
-        if interface_name in instances:
-            if path in instances[interface_name]:
-                return instances[interface_name][path]
+        path = kwargs.get('obj_path')
```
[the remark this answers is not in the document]
infirit: I added a default but this was not before you commented :smile:

blueman/bluez/Base.py
```diff
@@ -11,30 +11,24 @@
...
-                    continue
-                elif '/' in arg:
-                    path = arg
-                elif '.' in arg:
-                    interface_name = arg
-
-        if not interface_name:
+        # cls._interface_name or cls._obj_path need to exist to act as fallback, they may not
+        interface_name = kwargs.get('interface_name', cls._interface_name)
+        if interface_name is None:
```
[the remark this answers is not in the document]
infirit: > Why? This can only be true if `interface_name=None` got provided explicitly which is not expected.

interface_name is pretty much guaranteed not to be in the kwargs passed to ``__call__``.

> Actually no concrete class accepts an `interface_name` keyword at all, so we could just use `cls._interface_name` in any case

Currently it is always provided to ``__init__`` with ``self._interface_name`` so in theory we could just use ``self._interface_name`` and remove it from ``__init__``. I can do that if this is what you  want.

The issues is not so much with interface_name being in kwargs but that cls._interface_name __must__ exist to be used as second argument to ``dict.get()`` otherwise we end up with a ``NameError``. You are correct that this currently is not the case for interface_name, it may never, but it might be true in the future so I made sure to handle this.

blueman/bluez/Base.py
```diff
@@ -52,6 +46,7 @@ class Base(Gio.DBusProxy, metaclass=BaseMeta):
     __gsignals__: GSignals = {
         'property-changed': (GObject.SignalFlags.NO_HOOKS, None, (str, object, str))
     }
+    __instances__: Dict[str, Dict[str, "Base"]]
```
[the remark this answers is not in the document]
infirit: mypy does not handle metaclasses very well and both are needed. On line 18 for mypy to actually understand BaseMeta and here for it to actually show up in ``__annotation__``,

blueman/bluez/Adapter.py
infirit: I'll check out PEP3102 and see if this does what I want. If so I'll make the changes you suggest.

blueman/bluez/Base.py
```diff
@@ -11,30 +11,24 @@
 
 class BaseMeta(GObjectMeta):
     def __call__(cls, *args: str, **kwargs: str) -> "Base":
```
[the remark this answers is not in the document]
infirit: ``__call__`` does not allow pep3102 notation so we indeed have to remove ``*args``. But then we get an ugly error, see below. Which looks very confusing to me.

```python
d1 = Device('/org/bluez/hci0/dev_3C_F7_2A_07_BF_E3')
  File "<pyshell#1>", line 1, in <module>
    d1 = Device('/org/bluez/hci0/dev_3C_F7_2A_07_BF_E3')
TypeError: __call__() takes 1 positional argument but 2 were given
```

If I leave ``*args`` in ``__call__`` and add the pep3102 notation to ``__init__`` we die horrible death like below.

```python
d1 = Device('/org/bluez/hci0/dev_3C_F7_2A_07_BF_E3')
  File "<pyshell#4>", line 1, in <module>
    d1 = Device('/org/bluez/hci0/dev_3C_F7_2A_07_BF_E3')
  File "/home/sander/repos/blueman/blueman/bluez/Base.py", line 24, in __call__
    path = cls._obj_path
AttributeError: type object 'Device' has no attribute '_obj_path'
```

I like raising a RuntimeError much better than this...

blueman/bluez/Base.py
```diff
@@ -11,30 +11,24 @@
...
-            for arg in args:
-                if args is None:
-                    continue
-                elif '/' in arg:
-                    path = arg
-                elif '.' in arg:
-                    interface_name = arg
-
-        if not interface_name:
+        # cls._interface_name or cls._obj_path need to exist to act as fallback, they may not
```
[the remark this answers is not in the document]
infirit: Adding a default should suffice, will fix.

blueman/bluez/Base.py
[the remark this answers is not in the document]
cschramm: `__call__() takes 1 positional argument but 2 were given` makes sense to me. If you don't like it, I'm fine with the explicit `*args` check, but in any case we should make clear what calls are expected by changing all `__init__(self, obj_path: str)` to `__init__(self, *, obj_path: str)`. (If we would not check `*args` before, this would give `__init__() takes 1 positional argument but 2 were given` on `super().__call__(*args, **kwargs)`.)

Side note: PEP 3102 works fine for `__call__` but is not suitable in our case as we do not have any named keyword arguments.

blueman/bluez/Base.py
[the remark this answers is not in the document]
cschramm: As `interface_name` is not an argument of any of the implementations we could just use `cls.__instances__` instead of `cls.__instances__[interface_name]` but I'm fine with keeping this unused mechanism if you like.

cschramm:
> disconnect_by_func has only one argument

? It should accept the same data argument as `connect_signal`.

infirit:
> ? It should accept the same data argument as `connect_signal`.

You would think that but no, probably because of the pygobject overrides the documented c behaviour.

```
  File "/home/sander/repos/blueman/blueman/bluez/obex/Manager.py", line 80, in _on_object_removed
    transfer.disconnect_by_func(self._on_transfer_completed, True)
TypeError: GObject.disconnect_by_func() takes exactly 1 argument (2 given)
```

cschramm:
Hmpf, I was pretty sure I tested that for #1091. :grimacing: Will have to review all usages...

blueman/bluez/Base.py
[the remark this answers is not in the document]
infirit: I'll drop interface_name from ``__init__`` in a follow up commit.

blueman/bluez/Base.py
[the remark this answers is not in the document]
infirit: OK, I'll drop it in ``__call__``.

infirit:
> review all

I'll hunt them down and add it to the commit here.

infirit:
> > review all
> 
> I'll hunt them down and add it to the commit here.

Ugh, https://gitlab.gnome.org/GNOME/pygobject/issues/106. I did not know about this when I asked you to change this. I'll fix all of this with ``GObject.signal_handlers_disconnect_matched``

blueman/bluez/Base.py
```diff
@@ -10,34 +10,20 @@
 
 
 class BaseMeta(GObjectMeta):
-    def __call__(cls, *args: str, **kwargs: str) -> "Base":
-        instances: Dict[str, Dict[str, "Base"]] = cls.__dict__.setdefault("__instances__", {})
+    def __call__(cls, **kwargs: str) -> "Base":
+        if not hasattr(cls, "__instances__"):
+            cls.__instances__: Dict[str, Dict[str, "Base"]] = {}
```
cschramm: Just `Dict[str, "Base"]`. No need to use `cls._interface_name` here at all. :slightly_smiling_face:

infirit: I am confused, cls.__instances__ still contains the interface name and we really need it.

infirit:
>  I'll fix all of this with `GObject.signal_handlers_disconnect_matched`

So I can't use it as it expects things we can't provide. For now we can call disconnect_by_func as many times as we connect. A bit ugly and going back to storing the handler id's may be preferable right now. Commit 2bca79dafe3d4c6a4b6a69f50e02f4e27fa8f372 now calls disconnect_by_func as many times as needed.

blueman/bluez/Base.py
cschramm: Why? It's always the same for a given class, so `__instances__` will always have a single entry. Am I missing something?

infirit:
How I see it is that a single objectpath can and pretty much always exposes
different interfaces. A device may have org.bluez.Device1 and
org.bluez.Network1 and we have a class for each. Although technically
possible (but this is not how we implemented it) we can not call methods
for one interface on a class created for another.

So even though interface_name was removed as keyword argument we still have
to deal with them.

Or I am missing something obvious 😲😁

blueman/bluez/Base.py
infirit: How I see it is that a single objectpath can and pretty much always exposes
different interfaces. A device may have org.bluez.Device1 and
org.bluez.Network1 and we have a class for each. Although technically
possible (but this is not how we implemented it) we can not call methods
for one interface on a class created for another.

So even though interface_name was removed as keyword argument we still have
to deal with them.

Or I am missing something obvious 😲😁

infirit: See https://github.com/blueman-project/blueman/commit/aa90fdbb39d92a54491f61e15b27c7a27cb29ed2 for what is technically possible and what would be needed to drop ``interface_name`` from ``Base.__instances__``. I am happy to do this as it cuts down on the classes significantly :smile_cat:

cschramm: > A device may have org.bluez.Device1 and
> org.bluez.Network1 and we have a class for each.

That's my point. Each of our classes only handles a single interface, so there's no need to store instances by interface in the class property. The change to `Base` in https://github.com/blueman-project/blueman/commit/aa90fdbb39d92a54491f61e15b27c7a27cb29ed2 are exactly what I mean.

cschramm:
> Commit [2bca79d](https://github.com/blueman-project/blueman/commit/2bca79dafe3d4c6a4b6a69f50e02f4e27fa8f372) now calls disconnect_by_func as many times as needed.

Looks fine for `obex.Manager` an `PluginDialog` but not for `ManagerMenu` and `SerialService` as we actually want to disconnect the the signal for a specific argument (adapter path or serial port) there.

blueman/bluez/Base.py
infirit: Ah, the obvious part I was missing is that each class will get it's own ``cls.__instances__``. I thought that regardless of the class BaseMeta would just return the any instance based on the object path. I'll take care of this.

infirit:
> Looks fine for `obex.Manager` an `PluginDialog` but not for `ManagerMenu` and `SerialService` as we actually want to disconnect the the signal for a specific argument (adapter path or serial port) there.

I'll just go back to tracking id's for now :disappointed: .

blueman/main/Manager.py
```diff
@@ -168,8 +173,12 @@ def prop_changed(lst, adapter, key_value):
             key, value = key_value
             if key == "Discovering" and not value:
                 prog.finalize()
-                self.List.disconnect_by_func(on_progress)
-                self.List.disconnect_by_func(prop_changed)
+                # FIXME for some reason the signal handler is None
```
cschramm, line 176: Just stumbled across this while adding loads of type hints. The reason is that `ManagerProgressbar` overrides `connect` with a version that does not return anything. We could make it return the obtained handle, however it stores it internally to disconnect it in the `finalize` method, so I think those two `disconnect` calls would compete and I'm not sure how that'd work out. :thinking:

PR #1170, opened 2020-01-28
ShowConnected: Update connections in tooltip on bluetooth disable

infirit:


cschramm:
Relevant for master as well?

PR #1183, opened 2020-02-16
DhcpClient: increase timeout from 120 second

infirit:
Proxy timeouts are in milliseconds


infirit:
Hmm, changing commit message now requires re-review :crying_cat_face:

Issue #1192, opened 2020-03-16
AttributeError: 'PulseAudioUtils' object has no attribute 'connect_rfcomm'


nico202:
Hi, when pulse audio crashes, the error in the title appears.

It seems that connect_rfcomm
https://github.com/blueman-project/blueman/blob/a0ab19f5bb6a9c4f290604960856b3f3949bcf1f/blueman/main/PulseAudioUtils.py#L213

is defined in the PPPConnection and not in PulseAudioUtils.
https://github.com/blueman-project/blueman/blob/a0ab19f5bb6a9c4f290604960856b3f3949bcf1f/blueman/main/PPPConnection.py#L123

The only other usage of the method I could find is here:
https://github.com/blueman-project/blueman/blob/a0ab19f5bb6a9c4f290604960856b3f3949bcf1f/blueman/plugins/mechanism/Ppp.py#L27

where it's called on the right object.

Maybe the right function is connect_pulseadio

https://github.com/blueman-project/blueman/blob/a0ab19f5bb6a9c4f290604960856b3f3949bcf1f/blueman/main/PulseAudioUtils.py#L362

But it's my first look at the code, so I'm not sure

Thanks, Nicolò

nico202:
Created a PR in case my guess was right

infirit:
Yeah that looks like an obvious mistake. Do you have a way to reproduce
this, just kill pulseaudio?

nico202:
Basically, yes.  My pulseaudio was dying as soon as bluetooth was opened because of a broken plugin.

infirit:
Fix was merged, closing

PR #1193, opened 2020-03-16
Call the right method when pulseaudio crashes

nico202:
Fix for #1192

PulseAudioUtils does not implement connect_rfcomm but connect_pulseaudio


cschramm:
Thanks for the catch! I'm a bit confused about the build errors. Will have a look when I find some time...

infirit:
Merged, thx

PR #1195, opened 2020-03-16
Set proper cython language level to 3

infirit:
FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release! File: /home/sander/repos/blueman/module/_blueman.pyx

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,12 +7,12 @@ version: 2
 jobs:
   releasebuild:
     docker:
-      - image: debian:unstable
+      - image: circleci/buildpack-deps:testing
 
     steps:
       - checkout
-      - run: apt-get update
-      - run: apt-get install -y -qq --no-install-recommends sudo automake autoconf libglib2.0-dev libtool intltool python3-dev python-gi python-gi-dev cython dh-autoreconf libbluetooth-dev gtk-update-icon-cache python-pip
+      - run: sudo apt-get update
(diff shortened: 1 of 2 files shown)


infirit:
@cschramm For some reason on the debian image the linker fails. The difference comparing to a local build it is missing libpython3.8.so on the linker line. Why it is missing I don't know, do you see the same on your machine or is this maybe python3.8?

infirit:
It appears it is something on debian unstable, testing works fine.

infirit:
Ok, so circleci has its own testing images so I went with it. Do we really need unstable? Maybe do non required unstable builds and make testing required for now?

cschramm:
Still trying to replicate (Internet and thus package downloads are slooooowww...) but unstable does not seem to be necessary anyway.

cschramm:
Yes, I can confirm the issue with the meson build when using the current `python3` from Debian unstable.

cschramm:
The reason is in https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build. The `--embed` flag needs to be used to make pkgconfig include the library and is only meant for embedding the Python runtime, not for extensions. To be honest I do not see how not linking the library to the extension is supposed to work...

cschramm:
Got it. The only problem is that meson by default does not allow undefined symbols when linking. After adding `-Db_lundef=false` to the meson call, the ninja build works fine. :slightly_smiling_face:

PR #1196, opened 2020-03-17
Call the right method when pulseaudio crashes

infirit:
This is @nico202 patch rebased against master so that CI passes. I'll open a PR for stable as well that  contains this and the cython langversion commit.


infirit:
Meson did something for this in version 0.53, see https://github.com/mesonbuild/meson/pull/6341 however this is for the ``python`` meson module not ``dependency()``. Other projects do something like below.

``diff
-	PYTHON3 = dependency('python3')
+	PYTHON3 = dependency('python3-embed', required: false)
+	if not PYTHON3.found()
+		PYTHON3 = dependency('python3')
+	endif
``

I'll deal with meson, thanks for finding the source :+1:

PR #1197, opened 2020-03-17
Two fixes and move to debian testing for now

infirit:


cschramm:
I would mention that the fixes deserve changelog entries but the changelog is not up to data anyway... :see_no_evil:

infirit:
I am used to only updating CHANGELOG just before a release. I will try to think about this in the future.

I updated CHNAGELOG.md and bumped the version in configure.ac while I am at it.

cschramm:
I think keeping it in sync with the changes is easier than collecting it on release. Also one can see at first glance what happend in a branch.

PR #1199, opened 2020-03-18
Fix build on python 3.8

infirit:
This tries to get the embedded python lib and fallback on the normal one. Build is ok again on unstable so switched back.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - checkout
       - run: apt-get update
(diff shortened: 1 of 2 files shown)


cschramm:
Using the embed variant seems like a hack to me. I think the proper solution as thought of by the Python project (https://bugs.python.org/issue21536) would be to just allow the undefined symbols instead of linking the library, i.e. disable meson's `b_lundef` for our extension. This makes our extension more portable.

infirit:
Updated. Both builds pass again.

Issue #1204, opened 2020-03-29
Gtk.Menu.popup() is deprecated and breaks on Wayland


Xenopathic:
The method `popup()` of `Gtk.Menu` is deprecated since GTK 3.22 (https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Menu.html#Gtk.Menu.popup) (https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-popup), and does not work on Wayland compositors such as Sway. It should be replaced by `popup_at_widget()`, `popup_at_pointer()` or `popup_at_rect()`.

Other GTK projects have also encountered this, e.g. [Nautilus](https://bugzilla.gnome.org/show_bug.cgi?id=772096).

Also see https://github.com/swaywm/sway/issues/3135

infirit:
Merged  #1205 but would be nice if you could confirm it's working now.

Xenopathic:
Confirmed, it now works \o/

PR #1205, opened 2020-03-29
ManagerDeviceMenu: Create menu at pointer

infirit:
See #1204

@Xenopathic, assuming you use wayland so could you try this out? Under X it's fine.


infirit:
Added commit to change GtkStatusIcon.

PR #1206, opened 2020-04-03
Simplify autotools check for systemd paths

ukleinek:
PKG_CHECK_VAR already provides a way to overwrite the checked result using
an environment variable (preferably passed as an argument to ./configure).

Dropping the --with-systemd*unitdir parameters simplifies configure.ac,
makes the output of ./configure --help less confusing -- it had

	Optional Packages:
	  --with-systemdsystemunitdir=PATH
	                          Directory for systemd system unit files
	                          [default=/usr/lib/systemd/system]
	  --with-systemduserunitdir=PATH
	                          Directory for systemd user unit files
	                          [default=/usr/lib/systemd/user]

	Some influential environment variables:
	...
	  systemd_system_unit_dir
	              value of systemdsystemunitdir for systemd, overriding pkg-config
	  systemd_user_unit_dir
	              value of systemduserunitdir for systemd, overriding pkg-config

before -- and stops providing two ways to overwrite systemd's pkg-config
information.

For consistency with other variables use uppercase variable names.

The user visible change is that --with-systemd*unitdir doesn't exist any
more and instead of

	./configure --with-systemdsystemunitdir=/some/path

you have to use

	./configure SYSTEMD_SYSTEM_UNIT_DIR=/some/path

and to disable use:

	./configure SYSTEMD_SYSTEM_UNIT_DIR=''

.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -215,25 +215,10 @@ AC_ARG_ENABLE([appindicator],
 
 AM_CONDITIONAL([HAVE_APPINDICATOR], [test "x$use_appindicator" = "xyes"])
 
-AC_ARG_WITH([systemdsystemunitdir],
-  AS_HELP_STRING([--with-systemdsystemunitdir=PATH],
-    [Directory for systemd system unit files]
-    [[default=/usr/lib/systemd/system]]),
-  [systemd_system_unit_dir="$withval"],
-  [PKG_CHECK_VAR([systemd_system_unit_dir], [systemd], [systemdsystemunitdir],
-   [], [systemd_system_unit_dir='/usr/lib/systemd/system'])])
-AC_SUBST([systemd_system_unit_dir])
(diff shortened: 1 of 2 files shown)


infirit:
Appreciate your effort to simplify :+1: 

However I would have to say no to this, @cschramm may think otherwise but here are my reasons.

* --with-systemd{user,system}unitdir is the recommended way from systemd and it is what most distributions look for. https://www.freedesktop.org/software/systemd/man/daemon.html
* I think configuring packages through environment variables is bad practice.

Personally I would much rather drop autotools completely in favour of meson.

infirit:
If #1207 is working also for @cschramm then I will most likely remove autotools completely in master.

infirit:
I pretty much have meson handling everything we need, closing.

PR #1207, opened 2020-04-05
Attempt making run uninstalled with meson

infirit:
I had an epiphany this morning and I think I got running uninstalled with meson working now. This does break autotools but I don't want to keep it around anyway now that this is working.

@cschramm can you test if you can run uninstalled? Two things you need to know.

* You run the apps from withing the meson build directory (we have to as there is no other way to figure out the build dir)
* The meson build directory has to be a sub-directory in the source directory.


infirit:
fyi, blueman is completely broken for me atm. Not sure what is going on but I did see a recent glib update. So I tested that there are no errors importing things but never got to a ui or something.

infirit:
Now that I downgraded glib confirmed this works for me. I'll also add a commit to drop autotools completely :eyes:

cschramm:
Seems to work pretty good (I reverted 1cbd175f9d28a8af257d8e55a6f882ba1c9e8442 to test on glib 2.64). Icons are not working for me for some reason... :thinking:

infirit:
Thanks, I'll look into the icons and the mypy ci errors.

cschramm:
@infirit: Shall we revive this? :slightly_smiling_face:

infirit:
Yeah, devenv look interesting thx for the suggestion Eli.

eli-schwartz:
It's not great to have to copy every single source file into the build directory either.

I can think of three possible solutions off the top of my head:

- somehow inject the constants file into the right location without assuming that it's importable as `blueman.Constants`, likely involves e.g.
  ```
  if UNINSTALLED:
      exec(compile(open('real/path/to/Constants.py').read(), ...), globals(), locals())
  else:
      # Constants.py contents
  ```
- delete `__init__.py` from the source tree, create it at install time only, turning the in-source version into a [namespace package](https://peps.python.org/pep-0420/) that merges both the source tree and the build tree. ***Only*** Constants.py needs to be mirrored into the build tree, and it's importable as `blueman.Constants` after adding both `<sourcedir>/` and `<builddir>/` to `sys.path` to make `blueman` importable from both.

- Meson implements support for https://peps.python.org/pep-0660/ editable installs so that this automagically works. There's already interest in this from e.g. SciPy, so this will likely happen in the long run, although that's not super useful today.

eli-schwartz:
Probably number 2 is the best idea at the moment.

PR #1211, opened 2020-04-06
Fix IOChannel encoding

### LINKED ISSUE #1208: MASTER: GLib.IOChannel.read never returns on glib 2.64
It register and owns the name fine but we then block and timeout. It may be that we don't get a proxy for org.bluez. I downgraded back to 2.62 and things started working again.

Debian testing and unstable also have glib 2.64 so it would be nice to know if this is just my gentoo install or also show up in debian. @cschramm if you have the means could you check on your machine or the vm you created recently?

cschramm:
Closes #1208

### DIFF
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -63,6 +63,8 @@ class KillSwitch(AppletPlugin):
         if channel is None:
             raise ImportError('Failed to open rfkill device')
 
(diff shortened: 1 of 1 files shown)


infirit:
Does not work for me, it still blocks

infirit:
> Does not work for me, it still blocks

not to self, do not edit python file in /usr/. Do the sensible thing and create a damn patch en build it again... :disappointed:

PR #1218, opened 2020-04-15
Kill switch strings changed

comradekingu:
As per https://www.kernel.org/doc/html/latest/driver-api/rfkill.html

### DIFF
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -35,9 +35,8 @@ class Switch:
 
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
-    __description__ = _("Toggles a platform Bluetooth killswitch when Bluetooth power state changes "
-                        "(Useless with USB dongles) and makes sure a status icon is shown if there is a bluetooth "
-                        "killswitch but no adapter.")
+    __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
+                        "If the device is physically removed, a staus icon is shown.")
     __depends__ = ["PowerManager", "StatusIcon"]
     __icon__ = "system-shutdown"
 
@@ -61,7 +60,7 @@ class KillSwitch(AppletPlugin):
 
         channel = GLib.IOChannel.new_file("/dev/rfkill", "r")
         if channel is None:
-            raise ImportError('Failed to open rfkill device')
+            raise ImportError('Could not access RF kill switch')
 
         channel.set_encoding(None)
 
@@ -153,7 +152,7 @@ class KillSwitch(AppletPlugin):
             Mechanism().SetRfkillState('(b)', state, result_handler=reply, error_handler=error)
 
     def on_query_status_icon_visibility(self):
-        # Force status icon to show if bluetooth is soft-blocked
+        # Force status icon to show if Bluetooth is soft-blocked
         if not self._hardblocked and not self._enabled:
             return StatusIcon.FORCE_SHOW
 
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -36,7 +36,7 @@ class Switch:
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
     __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
-                        "So long as the device isn't physically removed, an icon is shown to change its power state.")
+                        "Allows turning Bluetooth back on from an icon that shows its status; provided it isn't unplugged by the system, or physically.")
     __depends__ = ["PowerManager", "StatusIcon"]
     __icon__ = "system-shutdown"
 
--- a/blueman/plugins/applet/KillSwitch.py
+++ b/blueman/plugins/applet/KillSwitch.py
@@ -36,7 +36,7 @@ class Switch:
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
     __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
-                        "If the device is physically removed, a staus icon is shown.")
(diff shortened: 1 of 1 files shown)


blueman/plugins/applet/KillSwitch.py
```diff
@@ -35,9 +35,8 @@ def __init__(self, idx, switch_type, soft, hard):
 
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
-    __description__ = _("Toggles a platform Bluetooth killswitch when Bluetooth power state changes "
-                        "(Useless with USB dongles) and makes sure a status icon is shown if there is a bluetooth "
-                        "killswitch but no adapter.")
+    __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
```
comradekingu, line 38: Not 100% sure here.

blueman/plugins/applet/KillSwitch.py
```diff
@@ -35,9 +35,8 @@ def __init__(self, idx, switch_type, soft, hard):
...
+                        "If the device is physically removed, a staus icon is shown.")
```
comradekingu, line 39: Not 100% sure here.

cschramm: It's the other way around. The plugin makes sure that the icon is displayed if Bluetooth is soft-blocked by a killswitch so that the user can use it to re-enable it. Otherwise, the StatusIcon plugin would hide it as it cannot see any adapters and feels useless.

I guess "there is a bluetooth killswitch but no adapter" was meant to mean exactly that: Bluetooth soft-blocked by a killswitch an no adapter *visible by the system* (otherwise the icon would be shown anyway).

comradekingu, line 39:
```suggestion
                        "So long as the device isn't physically removed, an icon is shown to change power state.")
```

comradekingu, line 39:
```suggestion
                        "If the device is physically removed, a status icon is shown.")
```

blueman/plugins/applet/KillSwitch.py
```diff
@@ -35,9 +35,8 @@ def __init__(self, idx, switch_type, soft, hard):
 
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
-    __description__ = _("Toggles a platform Bluetooth killswitch when Bluetooth power state changes "
-                        "(Useless with USB dongles) and makes sure a status icon is shown if there is a bluetooth "
-                        "killswitch but no adapter.")
+    __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
+                        "If the device is physically removed, a status icon is shown.")
```
comradekingu, line 39:
```suggestion
                        "So long as the device isn't physically removed, an icon is shown to change power state.")
```

comradekingu:
@cschramm I feel like I know how a soft and hard killswitch works, but "Bluetooth soft-blocked by a killswitch an no adapter visible by the system (otherwise the icon would be shown anyway)." has me confused. I had a go nonetheless.
I think explaining when the icon isn't there, and why isn't needed (?)

blueman/plugins/applet/KillSwitch.py
```diff
@@ -35,9 +35,8 @@ def __init__(self, idx, switch_type, soft, hard):
 
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
-    __description__ = _("Toggles a platform Bluetooth killswitch when Bluetooth power state changes "
-                        "(Useless with USB dongles) and makes sure a status icon is shown if there is a bluetooth "
-                        "killswitch but no adapter.")
+    __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
+                        "So long as the device isn't physically removed, an icon is shown to change power state.")
```
comradekingu, line 39:
```suggestion
                        "So long as the device isn't physically removed, an icon is shown to change its power state.")
```

cschramm:
Still not correct. When a killswitch is active, typically no adapters are visible (e.g. `bluetoothctl list` does not show any), even if there is one physically. A soft killswitch can be unlocked by software, a hard killswitch can't.

Standard behavior of the StatusIcon plugin (without the Killswitch plugin):
adapters available -> icon
no adapters -> no icon

Behavior with the Killswitch plugin:
adapters available -> icon
no adapters and no block -> no icon (there is no working adapter)
no adapters and a hard block -> no icon (blocked by a hard killswitch; nothing we could do about that)
no adapters and only soft blocks -> icon (blueman can enable Bluetooth by unblocking the soft killswitch(es))

So "So long as the device isn't physically removed, an icon is shown to change its power state." isn't correct in case of hard blocks.

infirit:
@comradekingu I can imagine that you wish you never started this ;-). I doubt Chris and I ever look at these plugin descriptions so thanks for reviewing this :+1:

blueman/plugins/applet/KillSwitch.py
```diff
@@ -35,9 +35,8 @@ def __init__(self, idx, switch_type, soft, hard):
 
 class KillSwitch(AppletPlugin):
     __author__ = "Walmis"
-    __description__ = _("Toggles a platform Bluetooth killswitch when Bluetooth power state changes "
-                        "(Useless with USB dongles) and makes sure a status icon is shown if there is a bluetooth "
-                        "killswitch but no adapter.")
+    __description__ = _("Switches Bluetooth killswitch status to match Bluetooth power state."
+                        "So long as the device isn't physically removed, an icon is shown to change its power state.")
```
comradekingu, line 39:
```suggestion
                        "Allows turning Bluetooth back on from an icon that shows its status; provided it isn't unplugged by the system, or physically.")
```

comradekingu:
@infirit My issue is foisting a translation I don't understand fully myself onto others. My confusion in, and attempted clarification of the matter hopefully means there is improvement to be had.

PR #1223, opened 2020-04-19
NetworkManager: Fix disconnecting NMDevice

infirit:
NMDevice.disconnect shadows the signal handler disconnect

  File "/home/sander/repos/blueman/blueman/main/NetworkManager.py", line 83, in _on_device_state_changed
    device.disconnect(self._statehandler)
TypeError: argument cancellable: Expected Gio.Cancellable, but got int


infirit:
In 2-1-stable I used disconnect_by_func which is fine.

PR #1230, opened 2020-04-26
Add workflow for weblate template update

cschramm:


.github/workflows/weblate.yml
```diff
@@ -0,0 +1,20 @@
...
+      - weblate
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo apt-get install intltool curl
+      - run: intltool-update --pot
```
infirit, line 16: I just dropped intltool so I'll merge this one first and then update the workflow.

cschramm:
Hrm, it looks like this constantly results in updated timestamps in the po files from Weblate, see https://github.com/blueman-project/blueman/pull/1233/commits/148b44556441a388e906047859bcc5701a77d89a... :confused:

infirit:
weblate appears hosted in germany so maybe a timezone thing?

cschramm:
I think the point is that when the job generates the pot file, it adds the current time as `POT-Creation-Date` and Weblate seems to spread that into all po files even if nothing changed. To avoid that, we'd have to check if something changed before uploading. A straightforward solution would be to leave that to gettext by first downloading Weblate's pot file. I think gettext will only touch the timestamp if something actually changed, so that otherwise we'll upload the exact same file that we just downloaded.

PR #1232, opened 2020-04-26
Drop intltool

infirit:
I added the polkit gettext its/loc files as debian stable does not ship them yet. We should revisit for 2.3. Also feel free to suggest changes in po/Makevars.

In meson do translation maintenance from the build dir and then run,
``ninja blueman-pot``
``ninja blueman-update-po``
``ninja blueman-gmo``

Autotools,
``make -C po blueman.pot-update``
``make -C po update-po``


infirit:
I updated the worlflow with the command that works from the ``po`` directory. I am not sure how you test this before but I do not expect this to fail. 

Also I dropped the commit that updates all the *.po files as I am sure weblate will take care of this.

PR #1234, opened 2020-04-26
Fix device removal handling

### LINKED ISSUE #1231: blueman-manager crashes when bluetoothd is restarted
**Environment**

blueman: `1:2.1.2-1.fc31`
BlueZ: `5.54-1.fc31`
Distribution: Fedora 31
Desktop environment: gnome

**The issue**

When restarting bluetoothd (`systemctl restart bluetooth.service`), `blueman-manager` crashes:
```
...
blueman-manager 12.42.06 DEBUG    Base:74 do_g_properties_changed: /org/bluez/hci0 {'Powered': False, 'Discovering': False, 'Class': 0}
blueman-manager 12.42.06 DEBUG    Manager:66 _on_object_removed: /org/bluez/hci0/dev_0C_A6_94_7D_B3_03
blueman-manager 12.42.06 DEBUG    Manager:66 _on_object_removed: /org/bluez/hci0/dev_38_18_4C_D3_19_E8
blueman-manager 12.42.06 DEBUG    Manager:66 _on_object_removed: /org/bluez/hci0/dev_00_22_37_46_50_2E
blueman-manager 12.42.06 DEBUG    Manager:62 _on_object_removed: /org/bluez/hci0
blueman-manager 12.42.06 DEBUG    DeviceList:238 set_adapter: Setting adapter to: None
blueman-manager 12.42.06 DEBUG    ManagerToolbar:62 on_adapter_changed: toolbar adapter None
blueman-manager 12.42.06 INFO     Manager:82 on_dbus_name_vanished: org.bluez
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf7 in position 1: invalid start byte
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd4 in position 3: invalid continuation byte
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd8 in position 3: invalid continuation byte
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb0 in position 0: invalid start byte
blueman-manager 12.42.06 INFO     Manager:102 on_dbus_name_appeared: org.bluez :1.2393
blueman-manager 12.42.06 ERROR    Manager:118 on_dbus_name_appeared: Default adapter not found, trying first available.
blueman-manager 12.42.06 ERROR    Manager:122 on_dbus_name_appeared: No adapter(s) found, exiting
blueman-manager 12.42.06 INFO     PulseAudioUtils:669 _on_delete: Destroying PulseAudioUtils instance
...
```

This is especially annoying as my initramfs is locking my encrypted luks `/` volume and chrooting into the initramfs to ensure data security. To do that, it need to `systemctl stop bluetooth` before suspending, and `systemctl start bluetooth` while resuming, because bluetoothd can otherwise inhibit the suspend process as it generates i/o on the locked encrypted root volume.

This means that blueman-manager crashes after every suspend.

**Suggested solution** 

Simply show "no bluetooth adapter found" in blueman-manager.
I found [this](https://github.com/blueman-project/blueman/blob/master/blueman/main/Manager.py#L100) FIXME, which seems related to this:  
> FIXME ui can handle BlueZ start/stop but we should inform user

Yannik:
This fixes multiple issues:
- if two devices are removed within the time the fader animation takes,
  _self._faderhandler is overwritten with the faderhandler id of the latter
  fader, which leads to non-removal of the first fader handler
- if two devices are removed within the time the fader animation takes,
  _self._faderhandler is set to None by the first __on_fader_finished
  callback, leading to a crash in the second callback because
  disconnect is called as `fader.disconnect(None)`
- If /org/bluez/hci0/dev_XX is removed, and then /org/bluez/hci0 is
  removed within the time the fader animation takes, the removal of
  /org/bluez/hci0 will lead to a second duplicate device_remove_event for
  /org/bluez/hci0/dev_XX, as it's not been removed from the liststore

This is one of the required patches to  fix #1231.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 from gettext import gettext as _
-from typing import Optional
+from typing import Dict
 import html
 import logging
 import cairo
@@ -85,7 +85,7 @@ class ManagerDeviceList(DeviceList):
 
         self.set_search_equal_func(self.search_func, None)
 
-        self._faderhandler: Optional[int] = None
+        self._faderhandlers: Dict[str, int] = {}
 
     def _on_settings_changed(self, settings, key):
         if key in ('sort-by', 'sort-order'):
@@ -221,16 +221,18 @@ class ManagerDeviceList(DeviceList):
         tree_iter = self.find_device(device)
 
         row_fader = self.get(tree_iter, "row_fader")["row_fader"]
-        self._faderhandler = row_fader.connect("animation-finished", self.__on_fader_finished, device, tree_iter)
+        super().device_remove_event(device)
+        self._faderhandlers.update({
+            device.get_object_path(): row_fader.connect("animation-finished", self.__on_fader_finished, device)
+        })
+
         row_fader.thaw()
         self.emit("device-selected", None, None)
         row_fader.animate(start=row_fader.get_state(), end=0.0, duration=400)
 
-    def __on_fader_finished(self, fader, device, tree_iter):
-        fader.disconnect(self._faderhandler)
-        self._faderhandler = None
+    def __on_fader_finished(self, fader, device):
+        fader.disconnect(self._faderhandlers.pop(device.get_object_path()))
         fader.freeze()
-        super().device_remove_event(device)
 
     def device_add_event(self, device):
         self.add_device(device)


blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, adapter=None, inst=None):
 
         self.set_search_equal_func(self.search_func, None)
 
-        self._faderhandler: Optional[int] = None
+        self._faderhandlers: List[int] = []
```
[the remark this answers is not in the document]
Yannik: Done

Yannik:
> Thanks for finding the bug. I rather not have the fade and the device_remove_event run at same (ish) time like this. We may end up removing the row before the fade has completed.
What happens when stopping bluetoothd is the following:
```
blueman-manager 12.42.06 DEBUG    Manager:66 _on_object_removed: /org/bluez/hci0/dev_0C_A6_94_7D_B3_03
blueman-manager 12.42.06 DEBUG    Manager:66 _on_object_removed: /org/bluez/hci0/dev_38_18_4C_D3_19_E8
blueman-manager 12.42.06 DEBUG    Manager:66 _on_object_removed: /org/bluez/hci0/dev_00_22_37_46_50_2E
blueman-manager 12.42.06 DEBUG    Manager:62 _on_object_removed: /org/bluez/hci0
```

At the time `_on_object_removed` is called for `/org/bluez/hci0`, all it's devices need to be deleted from the liststore already, as otherwise [DeviceList.py](https://github.com/blueman-project/blueman/blob/master/blueman/gui/DeviceList.py#L128) will call `device_remove_event` a second time for all the devices. 
 
> See my comment about using a dict inline with the code
Can you check the mypy error? I think it complains because it doesn't know the return type of `device.get_object_path`, right? How to fix this?

blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, adapter=None, inst=None):
 
         self.set_search_equal_func(self.search_func, None)
 
-        self._faderhandler: Optional[int] = None
+        self._faderhandlers = Dict[str, int]
```
cschramm, line 88: The syntax is wrong here. `self._faderhandlers: Dict[str, int] = {}` should be what you want. Otherwise you're assigning the type.

Yannik: @cschramm Thanks. Fixed.

blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -221,16 +218,16 @@ def device_remove_event(self, device):
...
+
         row_fader.thaw()
         self.emit("device-selected", None, None)
         row_fader.animate(start=row_fader.get_state(), end=0.0, duration=400)
 
-    def __on_fader_finished(self, fader, device, tree_iter):
-        fader.disconnect(self._faderhandler)
-        self._faderhandler = None
+    def __on_fader_finished(self, fader):
+        fader.disconnect_by_func(self.__on_fader_finished)
```
infirit, line 229: disconnect_by_func is broken (kindof see https://gitlab.gnome.org/GNOME/pygobject/-/issues/106) so don't use it.

I don't have time to look at the PR right now maybe in the weekend I can actually have a proper look.

Yannik: I know about that bug, but it does not effect us as we only ever attach one fader anyway. I did check if the handler is correctly disconnected by disconnect_by_func: it is.

infirit: Today it may cause no problems but tomorrow we change things and wonder why  it is broken. I have been bitten by this before so no disconnect_by_func until the bug is resolved.

Yannik: Ok, I will undo this commit

PR #1235, opened 2020-04-26
Fix bluetoothd start/stop blueman crash

### LINKED ISSUE #1231: blueman-manager crashes when bluetoothd is restarted

Yannik:
Fixes #1231

### DIFF
--- a/blueman/main/Manager.py
+++ b/blueman/main/Manager.py
@@ -14,7 +14,6 @@ from blueman.gui.manager.ManagerStats import ManagerStats
 from blueman.gui.manager.ManagerProgressbar import ManagerProgressbar
 from blueman.main.Config import Config
 from blueman.main.DBusProxies import AppletService, DBusProxyFailed
-from blueman.gui.CommonUi import ErrorDialog
 from blueman.gui.MessageArea import MessageArea
 from blueman.gui.Notification import Notification
 from blueman.main.PluginManager import PluginManager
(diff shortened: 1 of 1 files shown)


infirit:
With the signal handler problem fixed I see no crash. It properly closes as intended. If you do get a crash provide the log for ``blueman-manager --loglevel debug``.

As per the FIXME the UI can handle this but we need to inform the user. I see nothing in this PR that informs the user.

Yannik:
@infirit I think it's logged as crash by fedora because of the `exit(1)`.

Anyway, I don't think that the current behavior of blueman-manager suddenly closing after/during a suspend is desirable. 

>As per the FIXME the UI can handle this but we need to inform the user. I see nothing in this PR that informs the user. 

Quoting [this](https://github.com/blueman-project/blueman/issues/1231#issuecomment-619978737):

> With my patch, when no adapter is available, no devices are shown and the adapter menu is greyed out. I think this is a pretty good indication that the bluetooth adapter is not available (plus there's always the stdout/stderr output). Adding an extra frame like you suggested seems like a nice suggestion though. What about opening another issue to track this?
> 
> I'd like to keep this separate because this PR is targeted at fixing the crash on suspend, which seems like the higher priority.

infirit:
Our usage of exit(1) is probably not right here. We shouldn't be using it in production code anyway. I'll send a pr to replace all of them with SystemExit

infirit:
As mentioned, I really like to see the user notified properly. The crash report should be gone now by not using exitcode 1.

Yannik:
@cschramm Has this been fixed? I.e. no force-close on standby (with bluetooth getting disabled during the standby procedure) anymore?

cschramm:
I got confused by all the interleaved topics, so I'm not sure, but this PR says it fixes #1231 and #1231 is already closed (by #1234).

PR #1258, opened 2020-05-16
ManagerDeviceList: Hide devices with no name

infirit:
I don't like the menu entry name but at 1 in the morning I am out of ideas :sleepy:. If you have a suggestion feel free to comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Add Ctrl+Q and Ctrl+W accelerators for closing blueman-manager
 * Allow cancelling device connection attempts
 * Improved passkey handling (fixed padding, highlighting, single notifitication)
+* Hide devices with no name
 
 ### Bugs fixed
 
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -562,3 +562,10 @@ class ManagerDeviceList(DeviceList):
                 cell.set_property("surface", surface)
             else:
                 cell.set_property("surface", None)
+
+    def add_device(self, device: Device) -> None:
+        if "Name" not in device and self.Config["hide-unnamed"]:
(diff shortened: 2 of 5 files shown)


cschramm:
I'm not really sure if this is a good idea. The hypothesis is that once you set devices into pairing mode, they do provide a name? I'll test some devices to see if that holds.

If that's correct, don't we have to add devices on property changes (i.e. when a name appears) as well then? Currently we only add them when they show up initially.

Regarding the terms: unnamed, unknown, anonymous? A no name device is typically Chinese with no name on it. :wink:

infirit:
> The hypothesis is that once you set devices into pairing mode, they do provide a name?

Not really but if they do they will show up when searching. Android literally hides these when searching and you have to go into developer options to change it. I think windows does the same as I never see them there as well. I don't own one of these devices but have plenty showing up from my neighbours. If you do have one It would be nice if you can test this :smile: .

> If that's correct, don't we have to add devices on property changes (i.e. when a name appears) as well then? Currently we only add them when they show up initially.

I can move this into a filter function on the liststore but this is a pita to do.

> Regarding the terms: unnamed, unknown, anonymous?

Unnamed sound much better. I'll reverse the logic and make the menu entry "Show unnamed". Maybe add a little tooltip with explanation.

cschramm:
I tried to the hypothesis a little bit but did not find good examples. I guess it's valid, though. We can try.

cschramm:
After playing with Bluetooth a little bit in the last time I'm rather confident that this *is* a good idea now. :smile:

infirit:
Cool. I'll fix them conflicts later this week.

cschramm (dismissed):

cschramm:
Definitely deserves a changelog entry.

infirit:
Doh!! 😫. I'll do it tonight.

cschramm (approved):

infirit:
Rebased and updated changelog.

PR #1273, opened 2020-06-05
Do not use LaunchContext when event time is zero

infirit:
Same commits from #1272


infirit:
Forgot changelog, will do on next PR :disappointed:

PR #1279, opened 2020-06-17
Make more strings visible in the UI translatable

cwendling:
Possibly more strings lack translations, and likely most device and UUID names should be translated, but this takes care of what I encountered through the UI.


cschramm:
Great catches! There's one more at https://github.com/blueman-project/blueman/blob/1f48c31380af1c2c9a3782da7e03a07da863f7dd/blueman/gui/manager/ManagerMenu.py#L142.

Regarding device classes and service names the selection of translatable dates back to the original author or blueman and seems actually rather random to me. We should probably think about making all service names and all i18n strings for device classes translatable.

cwendling:
> Great catches! There's one more at
> 
> https://github.com/blueman-project/blueman/blob/1f48c31380af1c2c9a3782da7e03a07da863f7dd/blueman/gui/manager/ManagerMenu.py#L142
> .

Indeed, I updated the PR to include it, thanks!  I guess I didn't see it in the UI because I'm using French where it's mostly the same string (plus a couple accents, but that's not as obvious) ^^

I tried to grep a couple patterns that would have caught what I fixed, and they come up empty of relevant matches now, so there's a good chance at least those kind are fixed:
```console
$ git grep -Pi '(menuitem|label)(.new_[a-z_0-9]+)? *[(] *"'
$ git grep 'text="' 
data/configs/org.blueman.Mechanism.conf:        <policy context="default">
```

> Regarding device classes and service names the selection of translatable dates back to the original author or blueman and seems actually rather random to me. We should probably think about making all service names and all i18n strings for device classes translatable.

I would think so, yes.  Especially the i18n device class strings, as, well, being named i18n suggests they are supposed to be internationalized :slightly_smiling_face:

infirit:
Lgtm. I noticed those as well but forgot as I primarily use it in English
(Dutch is my primary  language).

For the deviceclass and sdp strings my preference would be to either
translate them all or none. I can make those translatable this weekend.

cwendling:
> For the deviceclass and sdp strings my preference would be to either translate them all or none.

OK, I marked them all in a second commit.

infirit:
Sorry for missing the approval request. I am a bit busy lately 🙄

PR #1283, opened 2020-06-19
Gettext translation maintanance

infirit:
This is my attempt to stop the loop we are in with weblate. 

Running ``make -C po/ blueman.pot-update`` does not update the ``POT-Creation-Date`` which hopefully fixes weblate trying to update it on every commit.

This does mean that we need to update ``blueman.pot`` when we do string changes with ``make -C po/ blueman.pot-update`` followed by ``make -C po/ update-po``.

### DIFF
--- a/.github/workflows/weblate.yml
+++ b/.github/workflows/weblate.yml
@@ -12,8 +12,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - run: sudo apt-get install gettext curl
-      - run: /usr/bin/xgettext --default-domain=blueman --directory=.. --from-code=UTF-8 --add-comments --files-from=./POTFILES.in --output=blueman.pot
         working-directory: po
       - env:
           TOKEN: ${{ secrets.WEBLATE_KEY }}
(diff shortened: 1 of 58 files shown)


cschramm:
I've commited some automation attempts with caching the blueman.pot as an artificat on GitHub. Unfortunately there's no native way to download artifacts from other workflow runs (yet) and https://github.com/blueman-project/blueman/runs/813963030 simply uses the artifact of a specific one but I can easily get the latest URL with `curl (curl 'https://api.github.com/repos/blueman-project/blueman/actions/workflows/weblate.yml/runs?status=success&branch=pot-test&event=push' | jq -r .workflow_runs[0].artifacts_url) | jq -r .artifacts[0].archive_download_url`.

What do you think about that approach?

We could also think about using `--no-location` to stop even more noise. Weblate does not seem to show those locations anywhere anyway.

cschramm:
https://github.com/blueman-project/blueman/commit/8829b9a6dde4c213d4cc6e1e6aab49e72f21bff7 is a complete version (modulo branches and Weblate upload).

infirit:
Hi, sorry so busy right now 😥. Let's try your approach first 😁.

Also we don't have complicated strings so stripping some of the comments
with --no-location should be fine.

cschramm:
#1286

infirit:
Seems things are much improved so closing.

PR #1298, opened 2020-07-20
Drop bind_textdomain_codeset

cschramm:
This method is deprecated in Python 3.8 and 3.9, will be removed in 3.10, and appears to be unnecessary in 3.6 and 3.7. The calls got added in https://github.com/blueman-project/blueman/pull/332 to fix an issue where GtkBuilder otherwise uses gettext with the environments' encoding and passes the results to g_markup_escape_text which always expects UTF-8.

The original report reffered to Python 2.7 and a blueman version that used PyGTK, so PyGTK must have called the *Python* gettext() family of functions for the setting to have an effect. When we fixed it, we already used PyGObject and currently we depend on Python 3 which changed the effect to only the lgettext() family of functions, so to still see this PyGObject would have to call the Python lgettext() family of functions. I could not verify that this is not the case as I did not spot *any* Python gettext calls in PyGTK and PyGObject at all, so I'm a little puzzled how it even happend in the first place with Python 2.7 and PyGTK.

### DIFF
--- a/apps/blueman-assistant.in
+++ b/apps/blueman-assistant.in
@@ -4,8 +4,6 @@ import os
 import sys
 import signal
 import logging
-from gettext import bind_textdomain_codeset
-
 import gi
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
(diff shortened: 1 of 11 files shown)


mgorny:
If hell breaks loose, I'll surely let you know ;-).

Issue #1269, opened 2020-05-28
blueman-applet 2.1.3 open popup window in background or minimized


joakim-tjernlund:
blueman: 2.1.3
BlueZ: git master
Distribution: Gentoo 
Desktop environment: MATE 1.24

* [ x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Whenever I click on applet to see my devices, the window either opens behind other windows or
even minimized.

I ran a blueman-applet --loglevel debug and ther was no output when a clicked the icon
<!--

ℹ Please use some service like [Pastebin](http://pastebin.com/) or [GitHub Gist](https://gist.github.com/) to post your logs and keep the thread clean and readable. Make sure to describe the exact steps you took when producing it.

->

joakim-tjernlund:
a blueman-manager --loglevel debug will open window on top though.

cschramm:
I've seen that as well recently. I'm pretty sure it's related to MATE 1.24 somehow. :thinking:

infirit:
This is because of GAppinfo but I am not sure how to fix this. It is definitely the focus stealing prevention most metacity based windowmanager do. I tried to drop the event time and a few other things, no joy 😞. 

I can workaround this by using a GDesktopAppinfo but then we need to provide desktop files for all blueman launch scripts. Adding desktop files is the easiest. The question is which do we want to show in the desktop menu.

joakim-tjernlund:
you mean:
```
cat /etc/xdg/autostart/blueman.desktop
[Desktop Entry]
Name=Blueman Applet
Comment=Blueman Bluetooth Manager
Icon=blueman
Exec=blueman-applet
Terminal=false
Type=Application
Categories=
```
What should be in there ?

joakim-tjernlund:
Why is blueman special here? its the only app I have with this problem?

cschramm:
@infirit: Doesn't seem to be GAppInfo alone. If I just run

```
from gi.repository import Gio
Gio.AppInfo.create_from_commandline("blueman-manager", "blueman", Gio.AppInfoCreateFlags.SUPPORTS_STARTUP_NOTIFICATION).launch()
```

which is basically the same just not from within blueman-applet, the window opens on top.

I do have similar issues with other applications in MATE, though, e.g. Thunderbird windows opening in background when clicking `mailto:` links.

infirit:
[...]

PR #1299, opened 2020-07-20
Fix #1296

### LINKED ISSUE #1296: blueman-applet occasionally launches 2 instances of blueman-tray
blueman: 2.1.3
BlueZ: 5.54
Distribution: Manjaro 20.0.3
Desktop environment: i3

I think this is a race condition involving the lock file.
If the `StatusIcon` and `AppIndicator` plugins are both enabled, then during startup the `StatusIcon` plugin will launch `blueman-tray` twice in `_on_plugins_changed()`: once when `StatusIcon` is loaded, once when `AppIndicator` is loaded. The function checks if there is an instance already running by reading the tray's lock file, but that's racy, since the tray can be running without having created the lock file. There's yet another check in `blueman-tray` using `check_single_instance()`, but that's racy too, here's how both trays can pass the check, assuming a stale lock file exists:
1. Tray 1 sees that a lock file exists, then reads a stale PID.
2. Tray 2 sees that a lock file exists, reads the same stale PID, removes the lock file, and writes its own lock file.
3. Tray 1 still thinks that the lock file is stale, so it removes tray 2's lock file and writes its own lock file.

Overall, there seem to be two problems here, and both need to be addressed for this to work correctly:
1. The check for a running tray in `_on_plugins_changed()` inside `StatusIcon`: even if `check_single_instance()` worked correctly, we could end up with the wrong instance of the tray running (i.e. the one launched first). This could be fixed by storing the PID of the tray process in a field of the `StatusIcon` class instead of reading it from the lock file, and reading the lock file when initializing that field in `StatusIcons`'s `__init__()` to handle the case where `blueman-tray` is launched first.
2. `check_single_instance()`: I think it can be fixed by serializing calls that check the same lock file. We can use [fcntl.flock](https://docs.python.org/3/library/fcntl.html#fcntl.flock) for that.

If my explanation is correct, then a workaround would be to disable the `AppIndicator` plugin on systems that can display GTK status icons.

j-piecuch:
This is my attempt at fixing #1296. It does two things:
- Serializes accesses to the same lock file from different processes using `fcntl.flock`
- Stores the PID of the tray application in the `StatusIcon` class

I'm open to feedback, especially since I don't have a lot of experience writing GTK/Python code.

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -250,15 +250,9 @@ class LockFileHandle(object):
 
     def get_pid(self) -> Optional[int]:
         with self as f:
-            # `return f.get_data[0]` is enough,
-            # but mypy can't infer the correct return type.
-            pid = f.get_data()[0]
-            if pid:
-                return int(pid)
-            else:
-                return None
+            return f.get_data()[0]
 
-    def __enter__(self):
+    def __enter__(self) -> LockFile:
         logging.debug("locking file %s" % self._filename)
         fcntl.flock(self._file.fileno(), fcntl.LOCK_EX)
         return self.LockFile(self._file)
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -300,9 +300,13 @@ def check_single_instance(name: str, unhide_func: Optional[Callable[[int], Any]]
                 print("Stale PID, overwriting")
             else:
                 print("There is an instance already running")
-                time = os.getenv("BLUEMAN_EVENT_TIME") or 0
+                env_time = os.getenv("BLUEMAN_EVENT_TIME")
+                if env_time:
+                    timestamp = int(env_time)
+                else:
+                    timestamp = int(time.clock_gettime(time.CLOCK_MONOTONIC_RAW) * 1000)
 
-                f.set_data(pid, time)
+                f.set_data(pid, timestamp)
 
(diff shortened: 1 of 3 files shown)


blueman/Functions.py
```diff
@@ -112,18 +110,14 @@ def launch(
...
-    else:
-        timestamp = gtktimestamp
-        display = Gdk.Display.get_default()
-        context = display.get_app_launch_context()
-        context.set_timestamp(timestamp)
+
+    timestamp = Gtk.get_current_event_time()
+    display = Gdk.Display.get_default()
+    context = display.get_app_launch_context()
+    context.set_timestamp(timestamp)
```
infirit, line 120: Don't undo this, see #1269 for details

j-piecuch: I see, I have adjusted the code to make up a timestamp when GTK can't provide one. I have tested the behavior from #1269 on MATE 1.24 and it seems to work fine.

infirit:
Ok, apologies for you doing all this work but I already had something I was working on from 2 years ago. See #1304

With Gtk/Gio.Application we can handle uniqueness properly and there is no need anymore for check_single_instance.

j-piecuch:
No problem, `Gtk.Application` looks like a much cleaner solution. Closing.

PR #1305, opened 2020-07-28
A few fixes (some from #1304)

infirit:
A couple of things I ran into while testing #1304

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ Makefile
 
 # meson build files
 /build*/
+
+# mypy
+/.mypy_cache/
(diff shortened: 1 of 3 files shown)


infirit:
> If blueman-tray gets launched it should just show nothing. So the only change is that we "save" the process, right?

My thought process here is that when manager state is not True the applet is most likely not ready and we should not be starting or listening to anything. In this situation there are no consequences so in essence it is to save the process.

Let me know if you want me to drop that commit, the primary fix is registerting the dbus service sooner.

cschramm:
Fine for me. Was just wondering if it fixes any actual issue that I missed. 👍

PR #1306, opened 2020-07-29
Use mapping keys for messages with plural forms

cschramm:
msgfmt --check-format does not require format specifications with message keys to be used in the translation, so this gives more flexibility like using "one" instead of an argument containing 1.

### DIFF
--- a/blueman/main/Adapter.py
+++ b/blueman/main/Adapter.py
@@ -143,7 +143,7 @@ class BluemanAdapters(Gtk.Window):
                 else:
                     return _("Hidden")
             else:
-                return gettext.ngettext("%d Minute", "%d Minutes", value) % value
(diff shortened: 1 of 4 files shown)


infirit:
If we are changing translatable strings may as well use fstrings, right?

cschramm:
Translating f-strings does and can not work as far as I know. While xgettext just extracts the raw f-string and translations can be added for it, the gettext function will not know the raw string, look up the interpolated one and of course find nothing. It's like `_("%s" % var)` instead of `_("%s") % var`.

cschramm:
...or do you mean `str.format()`? That I don't see the advantage over %-formatting to be honest.

infirit:
Didn't know fstrings are not supported by gettext 😞 .

PR #1308, opened 2020-07-30
Drop Generic Connect service name

cschramm:
This is unused as ServiceUUID.name has a special handling for that case.


infirit:
We should also drop ``BLUEMAN_GENERIC_CONNECT_ID`` at the top (also unused).

PR #1311, opened 2020-07-30
Fix Weblate workflow

cschramm:
Without xgettext blueman.pot-update does nothing and as it is not checked explicitly in configure.ac, this is not detected.


cschramm:
Not sure how this got closed and why I cannot re-open... :man_shrugging: :arrow_right: #1315

PR #1318, opened 2020-07-31
Migrate from CircleCI to GitHub actions

cschramm:
Also adds Python 3.9


infirit:
The only reason for circleci was that travis was lagging on their os image. It looks like actions uses docker images right? If so let's go :+1:

PR #1319, opened 2020-08-01
Set textdomain in blueman's binscripts

infirit:
I am confused translations worked at all :confused:. Here is what I see without this commit.
afbeelding

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -3,6 +3,7 @@ import os
 import sys
 import signal
 import logging
+import gettext
 
 
 # support running uninstalled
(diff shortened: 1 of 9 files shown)


cschramm:
eb6492f1 broke it, specifically the explicit `from gettext import gettext as _` (in `ManagerMenu.py` in your case, also e. g. `ManagerDeviceMenu.py`).

There are gettext `install()` calls for the correct domain in `Constants.py` and `Sdp.py`. It seems that the explicit import beforehand somehow breaks them. :thinking:

cschramm:
Ah, sure, `GNUTranslations.install` adds a `_` to `builtins` that points to its own `gettext` method while `gettext.gettext` depends on the currently set textdomain to find the translations and if it gets imported as `_` it shadows the builtin. If we always use an import and not the builtin (and mypy should complain if we don't), I think there's no point in any `install()` call at all. 

While we have to handle `BLUEMAN_SOURCE` before any (possibly) local imports, I think we should move other boilerplate stuff like translations, logging, signal handling, proc title etc. to a common place.

infirit:
The minimal required is setting the textdomain in the binscripts so no need to bind or install, updated the PR.

Also dropped the install in Sdp and Constants as it does not make sense to me, if we always imported these in the binscripts maybe.

PR #1322, opened 2020-08-01
Add a test job with Python 3.9

cschramm:
It needs at least cython 0.29.10, so Ubuntu 18.04 is insufficient.


cschramm:
Nope, wrong way. The relevant image is of course the Python image and those are based on Debian buster, so there is currently no sufficient cython version available in the official images.

PR #1325, opened 2020-08-02
Add and fix type hints in RecentConns

cschramm:
* Actually convert between stored and runtime items
* Add type hints
* Add a cast to resolve the circular type dependency issue
* Rename some internal methods and members

### DIFF
--- a/blueman/plugins/applet/RecentConns.py
+++ b/blueman/plugins/applet/RecentConns.py
@@ -2,13 +2,14 @@ from gettext import gettext as _
 from operator import itemgetter
 import time
 import logging
-from typing import Dict, List, TYPE_CHECKING, Optional, Callable
+from typing import Dict, List, TYPE_CHECKING, Optional, Callable, cast
 
 from blueman.bluez.Device import Device
-from blueman.bluez.errors import DBusNoSuchAdapterError
+from blueman.bluez.errors import DBusNoSuchAdapterError, BluezDBusException
 from blueman.gui.Notification import Notification
 from blueman.Sdp import ServiceUUID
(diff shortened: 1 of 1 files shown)


blueman/plugins/applet/RecentConns.py
```diff
@@ -150,15 +145,13 @@ def on_manager_state_changed(self, state):
...
-                self.Adapters[str(adapter.get_object_path())] = str(adapter["Address"])
+            self._adapters = {str(adapter.get_object_path()): str(adapter["Address"])
+                              for adapter in adapters}
 
-            for i in reversed(self.items):
-                try:
-                    i["device"] = self.get_device_path(i)
-                except (AdapterNotFound, DeviceNotFound):
-                    pass
+            for i in self.items:
```
infirit, line 151: You dropped ``reversed()`` here but left it in ``on_device_removed`` is that intentional? Also ``self.initialize()`` already sorts so I am confused why the reversed was there in the first place?

cschramm: The reverse iteration is necessary there as the loop removes the current item from the list. It does not have any use here. It is a pre ac5eb6d8a3c09b116a07258d1a8914f667e197bc leftover where this loop actually used to remove as well.

cschramm:
Quick notice: Removing the handler arguments in `on_item_activated` breaks some paths in `DBusService.connect_service`. I'm currently figuring out handler hell regarding service connections with type hints.

infirit:
callback/handler hell 😒🥺

The removal of ``*args`` in reply is causing problems?

cschramm:
Should break the handling for serial and network services.

cschramm:
See #1329

PR #1326, opened 2020-08-02
Update mypy

cschramm:
Depends on #1325

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -29,9 +29,9 @@ schemadir = join_paths([datadir, 'glib-2.0', 'schemas'])
 pythondir = join_paths([prefix, python.sysconfig_path('purelib')])
 
 if get_option('policykit')
-    have_polkit = 'yes'
+    have_polkit = 'True'
 else
-    have_polkit = 'no'
+    have_polkit = 'False'
 endif
 
 # Setup congfiguration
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -29,9 +29,9 @@ schemadir = join_paths([datadir, 'glib-2.0', 'schemas'])
 pythondir = join_paths([prefix, python.sysconfig_path('purelib')])
 
 if get_option('policykit')
-    have_polkit = 'yes'
+    have_polkit = 'True'
 else
-    have_polkit = 'no'
+    have_polkit = 'False'
 endif
```
infirit, line 35: These are strings not booleans. Meson's booleans are lowercase without quotes.

infirit: better even, change the option to store a boolean.

cschramm: That's actually what I intended. The `POLKIT` variable is meant to hold a Python boolean literal so that `POLKIT = @POLKIT@` in Constants.py.in initializes `POLKIT` to a boolean. The previous version, `POLKIT = "@have_polkit@" == "yes"`, leads to problems with `mypy --strict-equality` if `have_polkit` is `no`.

Using a boolean would be nice, but how could we process that in in `Constants.py.in` to make it available Python-wise (with both autoconf and meson)?

PR #1329, opened 2020-08-05
Add type hints to service connection handling

cschramm:
...and fix some issues.

The ok callback in connect_service and disconnect_service is used in places where it is expected to take no argument (Device.connect, Device.disconnect, SerialService.disconnect, NetworkService.disconnect) or a str argument (service_connect_handler, service_disconnect_handler, explicit call after rfcomm_connect_handler, SerialService.connect, NetworkService.connect). None of the usages of those methods - including the the DBus call due to its empty output signature - actually cares about the argument, so we can unify the signature to a noargs reply handler. ab84869 already dropped the unused reply argument in RecentConns and thus broke it for the serial and network service cases.

The reply value passed on by service_connect_handler, service_disconnect_handler, and NMBaseConnection is never actually used, so we can drop it and use a noargs callback. So is the NMBaseConnection's ability to not pass a reply_handler.

The reply handler for serial port services in ManagerDeviceMenu.on_connect seems broken as the result of the ConnectService DBus method will always be None due to its signature. (I think the code works, the message is just useless as it contains "None")

The err callback in connect_service typically receives an Exception but one explicit call in connect_service passes a str instead. The same holds for disconnect_service where a str gets passed in SerialService.disconnect. We can leave this
as it is for now, but require the callback to handle both.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -101,14 +101,11 @@ class ManagerDeviceMenu(Gtk.Menu):
     def on_connect(self, _item, service):
         device = service.device
 
-        def success(obj, result, _user_data):
+        def success(_obj, _result, _user_data):
             logging.info("success")
             prog.message(_("Success!"))
 
-            if isinstance(service, SerialPort) and SERIAL_PORT_SVCLASS_ID == service.short_uuid:
-                MessageArea.show_message(_("Serial port connected to %s") % result, None, "dialog-information")
-            else:
-                MessageArea.close()
+            MessageArea.close()
 
             self.unset_op(device)
 
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -5,12 +5,12 @@ from typing import Dict, List, Tuple, Optional
 
 from blueman.Constants import UI_PATH
 from blueman.Functions import create_menuitem, e_
+from blueman.Service import Service
 from blueman.bluez.Network import AnyNetwork
 from blueman.bluez.Device import AnyDevice, Device
 from blueman.gui.manager.ManagerProgressbar import ManagerProgressbar
 from blueman.main.DBusProxies import AppletService, DBusProxyFailed
 from blueman.gui.MessageArea import MessageArea
-from blueman.services import SerialPort
 from blueman.Sdp import (
     ServiceUUID,
     AUDIO_SOURCE_SVCLASS_ID,
@@ -18,8 +18,7 @@ from blueman.Sdp import (
     HANDSFREE_AGW_SVCLASS_ID,
     HANDSFREE_SVCLASS_ID,
     HEADSET_SVCLASS_ID,
-    HID_SVCLASS_ID,
-    SERIAL_PORT_SVCLASS_ID)
+    HID_SVCLASS_ID)
 
 import gi
 gi.require_version("Gtk", "3.0")
@@ -98,10 +97,10 @@ class ManagerDeviceMenu(Gtk.Menu):
         if key == "Connected":
             self.generate()
 
-    def on_connect(self, _item, service):
+    def on_connect(self, _item: Gtk.MenuItem, service: Service) -> None:
         device = service.device
 
-        def success(_obj, _result, _user_data):
+        def success(_obj: AppletService, _result: None, _user_data: None) -> None:
             logging.info("success")
             prog.message(_("Success!"))
 
@@ -109,7 +108,7 @@ class ManagerDeviceMenu(Gtk.Menu):
 
             self.unset_op(device)
 
-        def fail(obj, result, _user_data):
+        def fail(_obj: Optional[AppletService], result: GLib.Error, _user_data: None) -> None:
             prog.message(_("Failed"))
 
             self.unset_op(device)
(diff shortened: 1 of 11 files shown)


blueman/main/NetworkManager.py
```diff
@@ -36,27 +39,19 @@ def __init__(self, service, reply_handler=None, error_handler=None):
 
         self.find_or_create_connection()
 
-    def _return_or_reply_handler(self, msg):
+    def _handle_reply(self, msg):
```
infirit, line 42: This now became pointless log function, just use the callback directly.

blueman/main/NetworkManager.py
```diff
@@ -36,27 +39,19 @@ def __init__(self, service, reply_handler=None, error_handler=None):
...
         logging.debug(msg)
-        if not self.reply_handler:
-            return msg
-        else:
-            self.reply_handler(msg)
-            return
+        self.reply_handler()
```
infirit, line 44: ``msg`` got lost here.

blueman/main/NetworkManager.py
```diff
@@ -36,27 +39,19 @@ def __init__(self, service, reply_handler=None, error_handler=None):
...
 
-    def _raise_or_error_handler(self, error):
+    def _handle_error(self, error: Union[NMConnectionError, GLib.Error]) -> None:
```
infirit, line 46: This now became pointless log function, just use the callback directly.

blueman/main/NetworkManager.py
```diff
@@ -19,9 +21,10 @@ class NMConnectionError(Exception):
 class NMConnectionBase:
     conntype: str
 
-    def __init__(self, service, reply_handler=None, error_handler=None):
+    def __init__(self, service: Service, reply_handler: Callable[[], None],
+                 error_handler: Callable[[Union[NMConnectionError, GLib.Error]], None]):
```
infirit, line 25: This is not just adding typing, a callback is now mandatory. It should be a separate commit explaining why.

blueman/plugins/applet/PPPSupport.py
```diff
@@ -1,19 +1,28 @@
...
 from gi.repository import GLib
 
-from blueman.Sdp import DIALUP_NET_SVCLASS_ID
 import subprocess
 import logging
 
+from blueman.services import DialupNetwork
+
+if TYPE_CHECKING:
+    from blueman.main.Applet import BluemanApplet
```
infirit, line 20: What is the problem with always importing it?

blueman/plugins/applet/DBusService.py
```diff
@@ -69,9 +78,11 @@ def reply(rfcomm):
             else:
                 if not self.parent.Plugins.run_ex("service_connect_handler", cb, service, ok, err) \
                         and isinstance(service, (SerialService, NetworkService)):
-                    service.connect(reply_handler=ok, error_handler=err)
+                    service.connect(reply_handler=lambda *args: ok(), error_handler=err)
```
infirit, line 81: What is the point of ``*args``?

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -98,21 +97,18 @@ def _on_service_property_changed(self, _service, key, _value, _path):
...
+    def on_connect(self, _item: Gtk.MenuItem, service: Service) -> None:
         device = service.device
 
-        def success(obj, result, _user_data):
+        def success(_obj: AppletService, _result: None, _user_data: None) -> None:
             logging.info("success")
             prog.message(_("Success!"))
 
-            if isinstance(service, SerialPort) and SERIAL_PORT_SVCLASS_ID == service.short_uuid:
-                MessageArea.show_message(_("Serial port connected to %s") % result, None, "dialog-information")
```
infirit, line 109: Does the user get feedback that the serial port got connected?

blueman/plugins/applet/PPPSupport.py
cschramm: Circular imports.

blueman/main/NetworkManager.py
cschramm: Split out. Sorry for the mess. I was too happy when I finally figured things out. :sweat_smile:

blueman/plugins/applet/DBusService.py
cschramm: `SerialService.connect` and `NetworkService.connect` expect the callback to take a `str` argument (see their already present signatures). We do not care here but we at least care about it for `SerialService` in `rfcomm_connect_handler` methods. Doesn't look like we actually use it anywhere for `NetworkService` (it contains the network interface name which we always grab with the `property-changed` signal of `AnyNetwork`), but I think it makes sense that it returns it, so I'd leave it as it is.

blueman/gui/manager/ManagerDeviceMenu.py
cschramm: `SerialManager.on_rfcomm_connected` triggers a notification for that (That's actually a little problem if `PPPSupport` handles a dialup connection and `SerialManager` is not loaded, however, a message saying that a serial port got connected to None won't be helpful either. :laughing:). Also see first commit now.

blueman/main/NetworkManager.py
```diff
@@ -19,9 +19,9 @@ class NMConnectionError(Exception):
 class NMConnectionBase:
     conntype: str
 
-    def __init__(self, service, reply_handler=None, error_handler=None):
+    def __init__(self, service, reply_handler, error_handler):
         if self.conntype not in ('dun', 'panu'):
-            self._raise_or_error_handler(
```
cschramm, line 24: This was actually wrong as it'd raise an exception in the async call instead of calling the error handler.

blueman/main/NetworkManager.py
infirit: No worries :+1:

PR #1340, opened 2020-08-18
Add type hints to blueman.gui

cschramm:


blueman/gui/GenericList.py
```diff
@@ -127,11 +141,11 @@ def get(self, iterid, *items):
...
                         ret[items[i]] = self.liststore.get(tree_iter, self.ids[items[i]])[0]
         else:
-            return False
+            return {}
 
         return ret
 
-    def get_iter(self, path):
+    def get_iter(self, path: Union[Gtk.TreePath, int, str]) -> Optional[Gtk.TreeIter]:
         if path is None:
```
infirit, line 149: Path may be None so the Union is optional, right? If not then the check for None can be removed.

blueman/gui/GtkAnimation.py
```diff
@@ -7,6 +7,8 @@
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
+gi.require_version("Gdk", "3.0")
```
infirit, line 10: Is there a warning on the Gdk version?

blueman/gui/GtkAnimation.py
```diff
@@ -138,12 +142,14 @@ def on_draw(self, widget, cr):
...
-    def state_changed(self, state):
+        return False
+
+    def state_changed(self, state: float) -> None:
         self.tw.queue_draw()
 
 
 class CellFade(AnimBase):
-    def __init__(self, tw, path, columns=None):
+    def __init__(self, tw: Gtk.TreeView, path: Gtk.TreePath, columns: Iterable[int]) -> None:
```
infirit, line 152: tw is not a plain TreeView but ManagerDeviceList.

blueman/gui/GtkAnimation.py
cschramm: Yup, 
> PyGIWarning: Gdk was imported without specifying a version first. Use gi.require_version('Gdk', '3.0') before import to ensure that the right version gets loaded.

blueman/gui/GtkAnimation.py
cschramm: Does it have to be? From the attributes `CellFade` expects I think any `Gtk.TreeView` should work.

infirit: I think when the pygobject stubs are installed mypy will complain. But we can cross that bridge when that happens.

blueman/gui/GenericList.py
cschramm: You're right. At least `ManagerDeviceList.level_setup_event` might theoretically pass `None`, however `mypy` does not know that due to no GObject stubs as it sees just `Any`. Also `mypy --strict-equality` would catch this but it sees `path` as `Any` as well, as `Gtk.TreePath` is `Any`.

infirit (approved):

PR #1341, opened 2020-08-18
Add type hints to blueman.main 

cschramm:
Depends on #1332 (probably for no good reason :see_no_evil:)

### DIFF
--- a/blueman/main/DBusProxies.py
+++ b/blueman/main/DBusProxies.py
@@ -8,7 +8,8 @@ class DBusProxyFailed(Exception):
 
 
 class ProxyBase(Gio.DBusProxy, metaclass=SingletonGObjectMeta):
-    def __init__(self, name, interface_name, object_path='/', systembus=False, flags=0, *args, **kwargs):
+    def __init__(self, name: str, interface_name: str, object_path: str = "/", systembus: bool = False,
+                 flags: Gio.DBusProxyFlags = 0) -> None:
         if systembus:
             bustype = Gio.BusType.SYSTEM
         else:
@@ -19,8 +20,7 @@ class ProxyBase(Gio.DBusProxy, metaclass=SingletonGObjectMeta):
             g_interface_name=interface_name,
             g_object_path=object_path,
             g_bus_type=bustype,
-            g_flags=flags,
-            *args, **kwargs
+            g_flags=flags
         )
 
         try:
@@ -30,31 +30,31 @@ class ProxyBase(Gio.DBusProxy, metaclass=SingletonGObjectMeta):
 
 
 class Mechanism(ProxyBase):
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(name='org.blueman.Mechanism', interface_name='org.blueman.Mechanism',
                          object_path="/org/blueman/mechanism", systembus=True)
 
 
 class AppletService(ProxyBase):
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(name='org.blueman.Applet', interface_name='org.blueman.Applet',
                          object_path="/org/blueman/Applet")
 
 
 class ManagerService(ProxyBase):
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(name="org.blueman.Manager", interface_name="org.freedesktop.Application",
                          object_path="/org/blueman/Manager",
                          flags=Gio.DBusProxyFlags.DO_NOT_AUTO_START_AT_CONSTRUCTION)
 
     def _call_action(self, name: str) -> None:
-        def call_finish(proxy, resp):
+        def call_finish(proxy: "ManagerService", resp: Gio.AsyncResult) -> None:
             proxy.call_finish(resp)
 
         param = GLib.Variant('(sava{sv})', (name, [], {}))
         self.call('ActivateAction', param, Gio.DBusProxyFlags.NONE, -1, None, call_finish)
 
-    def startstop(self):
+    def startstop(self) -> None:
         if self.get_name_owner() is None:
             self._call_action("Activate")
         else:
--- a/blueman/main/DbusService.py
+++ b/blueman/main/DbusService.py
@@ -1,7 +1,7 @@
 import logging
 import sys
 import traceback
-from typing import Dict, Tuple, Callable, Set
+from typing import Dict, Tuple, Callable, Set, Optional, Any, Collection
 
 from gi.repository import Gio, GLib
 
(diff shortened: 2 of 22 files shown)


cschramm:
And there we are again, Python 3.6. Didn't last too long since our last encounter... :cowboy_hat_face:

cschramm:
I managed to get away without using `Generic` (which is an issue in Python 3.6 as `GenericMeta` and `GObjectMeta` are kind of incompatible, see #1349) although `PluginManager` feels a little strange now regarding its type variable which gets kind of erased. The given `plugin_class` defines the types of `__plugins` and `__classes` on initialization but when we know we have a `PluginManager` in our hands we have to specify the exact types of what we expect from `get_classes` or `get_plugins` and the argument to `__load_plugin` is not really checked (it just has to be an arbitrary type).

blueman/main/DBusProxies.py
```diff
@@ -8,7 +10,8 @@ class DBusProxyFailed(Exception):
 
 
 class ProxyBase(Gio.DBusProxy, metaclass=SingletonGObjectMeta):
-    def __init__(self, name, interface_name, object_path='/', systembus=False, flags=0, *args, **kwargs):
+    def __init__(self, name: str, interface_name: str, object_path: str = "/", systembus: bool = False,
+                 flags: Gio.DBusProxyFlags = 0, *args: Any, **kwargs: Any) -> None:
```
infirit, line 14: Feel free to drop the *args **kwargs as I doubt we actually make use of them. If we do add a new keyword and annotate it properly :-)

blueman/main/Manager.py
```diff
@@ -166,26 +169,26 @@ def on_dbus_name_appeared(_connection, name, owner):
 
         self.window.present_with_time(Gtk.get_current_event_time())
 
-    def _on_configure(self, window, event):
+    def _on_configure(self, _window: Gtk.Widget, event: Gdk.Event) -> bool:
```
infirit, line 172: Should be ``Gtk.ApplicationWindow``  If we don't get one of them we have a big problem :-).

blueman/main/NetConf.py
```diff
@@ -325,47 +331,47 @@ def get_default(cls):
...
         logging.info("init")
         self.version = class_id
-        self.dhcp_handler = None
+        self.dhcp_handler: Optional[DHCPHandler] = None
         self.ipt_rules: List[Tuple[str, str, str]] = []
 
         self.ip4_changed = False
 
-    def set_ipv4(self, ip, netmask):
+    def set_ipv4(self, ip: Optional[str], netmask: Optional[str]) -> None:
```
infirit, line 342: How are these optional, they shouldn't ever be afaik.

blueman/main/Sendto.py
```diff
@@ -137,10 +139,10 @@ def __init__(self, device, adapter_path, files):
 
         self.show()
 
-    def create_session(self):
+    def create_session(self) -> None:
         self.client.create_session(self.device['Address'], self.adapter["Address"])
 
-    def on_cancel(self, button):
+    def on_cancel(self, button: Optional[Gtk.Button]) -> None:
```
infirit, line 145: button should never be None', https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Button.html#Gtk.Button.signals.clicked. I see we check for the button but I don't understand why.

blueman/main/Sendto.py
```diff
@@ -204,24 +206,24 @@ def on_transfer_progress(self, _transfer, progress):
...
             self.emit("result", True)
 
-    def send_file(self, file_path):
+    def send_file(self, file_path: str) -> None:
         logging.info(file_path)
         if self.object_push:
             self.object_push.send_file(file_path)
 
-    def on_transfer_error(self, _transfer, msg=""):
+    def on_transfer_error(self, _transfer: Optional[Transfer], msg: str = "") -> None:
```
infirit, line 226: Similar to the button case earlier, the signal prohibits the first argument from being None.

blueman/main/SpeedCalc.py
```diff
@@ -3,13 +3,13 @@
 
 
 class SpeedCalc:
-    def __init__(self, moving_avg=3):
+    def __init__(self, moving_avg: float = 3) -> None:
```
infirit, line 6: Shouldn't this cause an error, 3 being an int while the annotation says float? If this is indeed a float let's make it clear with the default of 3.0.

blueman/main/DbusService.py
```diff
@@ -1,28 +1,28 @@
...
 
     @property
-    def message(self):
+    def message(self) -> Optional[str]:
         return self._message
 
 
 class DbusService:
-    def __init__(self, bus_name, interface_name, path, bus_type):
+    def __init__(self, bus_name: Optional[str], interface_name: str, path: str, bus_type: Gio.BusType) -> None:
```
infirit, line 25: bus_name should never be optional as there is no point in having a dbus service but not own a name.

blueman/main/NetConf.py
cschramm: See `Network._disabled_network` :confused:

blueman/main/Sendto.py
cschramm: `on_response` in `on_transfer_error` calls this directly with button set to `None`.

blueman/main/Sendto.py
cschramm: `on_transfer_failed`

blueman/main/NetConf.py
infirit: Hmm, ok.

blueman/main/Sendto.py
infirit: Did not think we did that, ok then.

blueman/main/SpeedCalc.py
cschramm: mypy is fine with using `int` values where `float` is expected. I can change the default to `3.0`.

infirit: I think this would make more sense and consistant. Even if python and mypy are fine with it.

blueman/main/DbusService.py
cschramm: There is. We use it for our agents which we then just register (by object path). They do not need to own a specific bus name.

infirit (approved):

infirit:
I shouldn't do review on windows any-more before playing games. Next time when I make little sense remind me 😉

PR #1342, opened 2020-08-18
Fixes from type hints

cschramm:
Some actual fixes for issues found while adding type hints, cherry-picked from #1338 #1340 #1341.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Handle os.remove failing
 * Fix disconnecting NMDevice
 * Untranslated string in DiscvManager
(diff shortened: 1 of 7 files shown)


cschramm:
The issues I added to the changelog are actually in the current releases. I wonder why the default PIN thing didn't bite us yet. :thinking:

infirit:
> 
> 
> The issues I added to the changelog are actually in the current releases. I wonder why the default PIN thing didn't bite us yet. 🤔

BlueZ handles most keyboard, mice and other devices with default pins. The defaults still left are for more rare devices which I don't own and very likely neither do you.

infirit (approved):

PR #1347, opened 2020-08-18
Proposed french translation

ymollard:
Here are some proposals for new french messages, phrasing corrections and syntax correction for "...".
I also made some modifications to the commented ones, in case.

Do you confirm that:
* "Next DST Change Service" is related to the Daylight saving time?
* the following message "Connect to" [there](https://github.com/ymollard/blueman/commit/4f6d6c86196c529ac079948155f2159298476078#diff-42793ab00dcd448b33776fc86e5eaaf4L269) is actually supposed to be "Connect to" and not "Connected to"?

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -123,7 +123,7 @@ msgstr "Exécuter l'assistant de configuration pour ce périphérique"
 #. translators: toolbar item: keep it as short as possible
 #: data/ui/manager-main.ui:151
 msgid "Setup..."
-msgstr "Installation..."
+msgstr "Installation …"
 
 #: data/ui/manager-main.ui:164 blueman/gui/manager/ManagerDeviceMenu.py:385
 msgid "Remove this device from the known devices list"
@@ -274,11 +274,11 @@ msgstr "Connecter"
 
 #: data/ui/assistant.ui:161
 msgid "<b>Please wait...</b>"
-msgstr "<b>Veuillez patienter...</b>"
+msgstr "<b>Veuillez patienter …</b>"
 
 #: data/ui/assistant.ui:166
 msgid "Connecting..."
-msgstr "Connexion en cours..."
+msgstr "Connexion en cours …"
 
 #: data/ui/assistant.ui:173
 msgid "<b>Congratulations, device successfully added</b>"
@@ -302,7 +302,7 @@ msgstr "<b>Description du plugin :</b>"
 
 #: data/ui/applet-plugins-widget.ui:130
 msgid "Not specified"
-msgstr "Pas spécifié"
+msgstr "Non spécifié"
 
 #: data/ui/applet-plugins-widget.ui:146
 msgid "<b>Author:</b>"
@@ -373,7 +373,7 @@ msgstr "Envoyer le pense-bête"
 #: blueman/main/Adapter.py:49
 msgid "Bluetooth needs to be turned on for the adapter manager to work"
 msgstr ""
-"Bluetooth doit être activé pour que ce gestionnaire de périphériques puisse "
+"Le Bluetooth doit être activé pour que le gestionnaire de périphériques puisse "
 "fonctionner"
 
 #: blueman/main/Adapter.py:86 data/blueman-adapters.desktop.in:3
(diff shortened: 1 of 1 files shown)


cschramm:
Translations are managed on Weblate: https://hosted.weblate.org/projects/blueman/blueman/fr/
If we change it here, changes will get overridden.

>  "Next DST Change Service" is related to the Daylight saving time?
That would be my guess as well, but I don't know, to be honest. It's GATT service 0x1807: https://www.bluetooth.com/specifications/gatt/services/

> the following message "Connect to" there is actually supposed to be "Connect to" and not "Connected to"?

Yes. It's used in blueman-assistant's "Connect" step.

ymollard:
> Ellipsis has a GUI meaning, 

Oh thanks I didn't know! As a user I'm not convinced by their guidelines though, I've just discovered ellipsis is used in all my apps and I think it's never good when your users don't pay attention to UI details that are supposed to help them >:)

> > How can we also internationalize messages such as "Host is down", "Software caused connection abort" ...?
> 
> Short answer: We can't. Those are errors reported by the kernel (EHOSTDOWN and ENOPROTOOPT in those cases). Honest answer: We could replace them, but that feels wrong to me regarding effort and maintenance.

Aren't these EHOSTDOWN and ENOPROTOOPT just integers? Shouldn't all apps translate them on the contrary? It looks wrong to me to display an error message in a language the user cannot understand. No "popular" OS such as Android or Windows behave like this and I think this kind of details is the ones that make Linux distros hard to use for many users. How does it require more effort than translating other messages? 

> Adding a new device to the list happens automatically by discovering it. The setup dialog is used for much more, especially pairing devices and connecting services. To be honest, I do not really like it and consider dropping it from time to time in favor of the clear "atomic" actions that we have anyway.

OK let's keep it like this, then.

cschramm:
> Aren't these EHOSTDOWN and ENOPROTOOPT just integers?

Yes and no. Basically it's errno.h error codes, yes. We do not get them as integers, no. bluetoothd already calls `strerror` on them and sends us the obtained description via DBus. Even if we would have the codes in hand and call `strerror` on them, the output depends on the libc implementation and although `strerror` respects `LC_MESSAGES` is usually not translated.

What we could do, is:
* Specify libc implementations that we want to support.
* Collect all errors that bluetoothd might send via DBus, immediately or forwarded from the kernel (as most of the ~133 total errors are probably not relevant at all).
* Collect the respective messages from the targeted libc implementations.
* Unify them where necessary.
* Make the unified messages translatable.
* Keep watching the kernel development for additional error codes that might be returned to BlueZ.
* Keep watching BlueZ for additional error codes that it might return or forward from the kernel.
* Keep watching the targeted libc implementations' development for changed error descriptions.

That feels wrong to me regarding effort and maintenance.

We could think about going through (most of) this for regular errors like the ones mentioned and describe what happend and what to do about it to the user. Feel free to collect some as a starting point. You should find the relevant ones in our GitHub issues with keywords like `DBusFailedError`. Besides the messages it makes sense to also look up the error codes e.g. in [errlist.h](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/gnu/errlist.h;h=5d11ed723d1b0b8e056855906d1aa0d5e0e9cb0e;hb=HEAD) and check other libc implementations (I'd say at least musl), and we'll also need the possible contexts of the errors (usually device connection, most of the times).

infirit:
We have to accept that in the computing world English is the dominant language. I not ok with localizing error messages for the simple reason I only speak 2.5 out of hundreds and I am not willing to help people posting error messages in Hebrew or French.

ymollard:
It looks pretty clear to me that the kernel won't speak French indeed. I agree that the logs/console outputs or anything that is supposed to be hidden at first sight must speak English and provide useful debugging info, e.g. *Host is down*. But when connection through the GUI fails, the GUI should not say it using a different language that the language used to initiate the connection, this is what I'm talking about here. Additionally, these messages could be re-contextualized, even in English, e.g. *The Bluetooth device is off*.

I would have expected that a basic switch/case with errnos would do the job, but I'm unaware of libc implementations and how DBus works. Watching new errnos or new implementations is always good but in the worst case we can always present the English string as it works so far anyway. What would be a more serious problem is that if, depending of the libc implementation, some errno may have different meanings. 

> Feel free to collect some as a starting point.

OK, done as a new Issue

cschramm:
> depending of the libc implementation, some errno may have different meanings.

You mean the integer values? No, they are defined by Linux, depending on the platform, though, i.e. they are different for at least Alpha, PA-RISC, MIPS, and SPARC. We do not have them in hand, so we do not have to worry about that, though. :smile:

PR #1353, opened 2020-08-23
Add stubs/gi

cschramm:
Opening the GI repository stubs discussion.

First off: This will be a long way.

We need a `gi.repository` structure to be able to use gir stubs. Unfortunately that means that we need a complete `gi` structure but it's rather simple for us as all we need is `gi.require_version` and `gi.types.GObjectMeta`. (Adding it already finds two additional typing issues :see_no_evil:)

Regarding stub generation I'm using https://github.com/cschramm/pgi-docgen/tree/stub-generation which is currently just https://github.com/pygobject/pgi-docgen/commit/12c9dbbe1112abf77aa03a1e5cf2d1d9013489d4 merged with https://github.com/kaiw/pgi-docgen/commit/992976e34c5b3846ed487cbeaac916ab55dfba25 and a Pipfile for my convenience (the dependencies are otherwise just mangled into an (incomplete) setup script with pip calls or a Dockerfile).

One major problem with the currently existing stub generation is that it only considers GIR files and completely neglects both `gi.overrides` and `_gi`, leading to a lot of typing errors regarding basic stuff like `GLib.timeout_add` and `GObject.connect` and https://gitlab.gnome.org/GNOME/pygobject/-/tree/master/gi/overrides is *huge*.

When I now generate GLib stubs with `pipenv run ./pgi-docgen stubs $BLUMEAN_SRC/stubs/gi/repository/ GLib-2.0` running `MYPYPATH=stubs mypy -p blueman --ignore-missing-imports --warn-unused-configs --disallow-any-generics --disallow-incomplete-defs --check-untyped-defs --disallow-untyped-decorators --no-implicit-optional --warn-redundant-casts --warn-unused-ignores --warn-return-any --no-implicit-reexport --strict-equality` gives me ~200 errors (including a couple of errors in the pyi files themselves, mainly unused ignores). With Gtk and GdkX11 in place, it's ~1650 (side note: that's some more than what I get with pygobject-stubs).

I found the following paths if we want to iterate one module after another (and that makes a lot of sense to me):
girtree
(mainly GLib -> GObject -> Gio -> Gdk -> Gtk for us)

pgi-docgen stubs do not even seem ready to add overrides e.g. for `GLib.timeout_add` as there is `pgidocgen/stuboverrides.py` to *add* stuff to modules and classes, but it cannot be used to actually override stuff. E.g. if I just add (overloading) definitions for `timeout_add` to the `GLib` namespace, the actual implementation that still gets generated below them will not match / replace them again and to be correct there, we would probably even have to remove it anyway as that signature is not valid from Python due to `gi.overrides`. I'm not sure if it makes sense to work on pgi-docgen stubs to make it capable of actually overriding stuff so that it generates useful stubs or - much easier but feels kinda selfish -  generate useless stubs once and then adapt them to reality.


cschramm:
(This is base on #1341 for more complete type hints now)

The path through GLib, GObject and Gio was surprisingly quick with a "pgi-docgen stubs, then edit" approach. https://github.com/blueman-project/blueman/pull/1304#discussion_r476778561 is currently blocking the only error that is left for them. :smiley:

cschramm:
:scream: :scream: :scream:

cschramm:
I'm a little sad that I found just one little bug that could actually show up (https://github.com/blueman-project/blueman/pull/1353/commits/55792c52220aa3de5bb287c9cb6b9c1aee85c2d7). On the other side... pretty cool! :smiley:

infirit:
What, only one, you must have made a mistake 😆. This is going to take about of time to go through 😱.

cschramm:
I tried to keep a clean commit history, in case you did not realize. There's obviously no need to review the generation commits. 🙂

infirit:
Yea, my plan is to skim through the generated stubs but look closely at the modification that were done.

Gtk3 is not going to change much but this will have to be redone when we move to Gtk4.

cschramm:
I actually thought about removing all stuff that we do not use from the generated stubs as there are probably a lot of falsehoods in there that might lead to confusion if somebody makes the mistake to rely on the them (e.g. callback types are generally defined to take an optional object as user data while they usually take a variable amount of arguments; if one implements that signature as it is defined it would be wrong to pass zero or more than one user data argument but would be perfectly fine from the stubs perspective). Not sure how much work that would be, though, and one would have to restore the generated stuff as a basis when using additional things.

blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -200,14 +204,20 @@ def make_device_icon(self, icon_info: Gtk.IconInfo, is_paired: bool = False, is_
         ctx = cairo.Context(target)
 
         if is_paired:
-            icon_info = self.get_icon_info("dialog-password", 16, False)
+            _icon_info = self.get_icon_info("dialog-password", 16, False)
```
infirit, line 207: What is the point of the new variable? Below does the same, right?

```python
icon_info = self.get_icon_info("dialog-password", 16, False)
assert icon_info is not None
```

blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -200,14 +204,20 @@ def make_device_icon(self, icon_info: Gtk.IconInfo, is_paired: bool = False, is_
...
             ctx.paint_with_alpha(0.8)
 
         if is_trusted:
-            icon_info = self.get_icon_info("blueman-trust", 16, False)
+            _icon_info = self.get_icon_info("blueman-trust", 16, False)
+            assert _icon_info is not None
+            icon_info = _icon_info
             trusted_surface = icon_info.load_surface(window)
+            assert isinstance(target, cairo.ImageSurface)
+            assert isinstance(trusted_surface, cairo.ImageSurface)
```
infirit, line 220: This should be fixed in the stub that load_surface return an instance of cairo.Surface. Or just force it in the stub to be an cairo.ImageSurface as I doubt we'll ever see anything else.

blueman/gui/manager/ManagerProgressbar.py
```diff
@@ -67,15 +69,20 @@ def __init__(self, blueman: "Blueman", cancellable: bool = True, text: str = _("
...
         c = Gdk.Cursor.new(Gdk.CursorType.HAND2)
-        assert self.Blueman.window is not None
-        self.Blueman.window.get_window().set_cursor(c)
+        self._get_window().set_cursor(c)
         return False
 
     def _on_leave(self, _evbox: Gtk.EventBox, _event: Gdk.Event) -> bool:
         assert self.Blueman.window is not None
-        self.Blueman.window.get_window().set_cursor(None)
+        self._get_window().set_cursor(None)
```
infirit, line 85: If we are going to add a helper function let's move it to Blueman and use it everywhere we use Blueman.window probably getting rid of a few asserts in the process.

blueman/plugins/manager/Info.py
```diff
@@ -117,5 +119,8 @@ class Info(ManagerPlugin, MenuItemsProvider):
     def on_request_menu_items(self, manager_menu: ManagerDeviceMenu, device: Device) -> List[Tuple[Gtk.MenuItem, int]]:
         item = create_menuitem(_("_Info"), "dialog-information")
         item.props.tooltip_text = _("Show device information")
-        item.connect('activate', lambda x: show_info(device, manager_menu.get_toplevel()))
+        _window = manager_menu.get_toplevel()
```
infirit, line 122: Similar to earlier, why is the new variable needed?

```python
window = manager_menu.get_toplevel()
assert isinstance(window, Gtk.Window)
```

infirit (commented):
I see _get_widget copy/pasted all over the place. My suggestion would be to subclass Gtk.Builder, override get_object and move the instance checking in there. Alternatively which may be cleaner, add a new method ``get_widget``.

As for the obviously wrong annotation in the stubs. Not sure, but we can take the approach that we fix them when it causes problems.

I still hate it with a passion how verbose and unreadable some of the annotations make function and classes. Oh well :shrug:

blueman/gui/manager/ManagerProgressbar.py
```diff
@@ -166,7 +166,7 @@ def pulse() -> bool:
 
         if not self.pulsing:
             self.pulsing = True
-            GLib.timeout_add(1000 / 24, pulse)
+            GLib.timeout_add(41, pulse)
```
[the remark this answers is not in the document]
cschramm: Maybe it was meant to be 24 Hz but I have no idea why.

blueman/main/PPPConnection.py
```diff
@@ -176,7 +177,7 @@ def send_command(self, command: str) -> None:
         os.write(self.file, out.encode("UTF-8"))
         termios.tcdrain(self.file)
 
-    def on_data_ready(self, _source: GLib.IOChannel, condition: GLib.IOCondition, command_id: int) -> bool:
+    def on_data_ready(self, _source: int, condition: GLib.IOCondition, command_id: int) -> bool:
```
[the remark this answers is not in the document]
cschramm: This is an override, see https://gitlab.gnome.org/GNOME/pygobject/-/blob/7d5f4cd41ee0221c13842ab6323b7a016df02c7e/gi/overrides/GLib.py#L638.
If `io_add_watch` gets called with a file descriptor (`int`) the callback will receive a file descriptor as well (kept in the `func_fdtransform` closure).

blueman/bluez/AnyBase.py
```diff
@@ -49,6 +49,7 @@ def _on_properties_changed(
 
     def close(self) -> None:
         if self.__signal:
-            self.__bus.signal_unsubscribe(self.__signal)
+            if self.__bus is not None:
+                self.__bus.signal_unsubscribe(self.__signal)
```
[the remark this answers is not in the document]
cschramm: From the interface there's nothing holding a caller from calling `close` twice. Instead of the condition we could also use an `assert`, causing a crash instead of silently doing nothing if it actually happens. mypy just wants to know that we are aware of the fact that `self.__bus` might be `None`.

blueman/gui/GenericList.py
```diff
@@ -61,12 +61,12 @@ def _load(self, data: Iterable[ListDataDict]) -> None:
...
         (model, tree_iter) = self.selection.get_selected()
-
+        assert tree_iter is not None
         return tree_iter
 
     def delete(self, iterid: Union[Gtk.TreeIter, Gtk.TreePath, int, str]) -> bool:
-        if type(iterid) == Gtk.TreeIter:
-            tree_iter = iterid
+        if isinstance(iterid, Gtk.TreeIter):
+            tree_iter: Optional[Gtk.TreeIter] = iterid
```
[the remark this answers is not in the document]
cschramm: `get_iter` in the other branch may return `None` and the variable needs a common type.

blueman/gui/Notification.py
```diff
@@ -30,8 +30,8 @@ def __init__(self, summary: str, message: str, _timeout: int = -1,
                  actions: Optional[Iterable[Tuple[str, str]]] = None,
                  actions_cb: Optional[Callable[[str], None]] = None, icon_name: Optional[str] = None,
                  image_data: Optional[GdkPixbuf.Pixbuf] = None) -> None:
-        super().__init__(parent=None, flags=0, type=Gtk.MessageType.QUESTION,
-                         buttons=Gtk.ButtonsType.NONE, message_format=None)
+        super().__init__(parent=None, type=Gtk.MessageType.QUESTION,
+                         buttons=Gtk.ButtonsType.NONE, text=None)
```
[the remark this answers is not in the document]
cschramm: I was a little bit confused here. `text` defaults to `""`, not `NULL`, see https://developer.gnome.org/gtk3/stable/GtkMessageDialog.html#GtkMessageDialog--text, so I think `text=None` should actually be something else than the default but I did not really look into it.

blueman/gui/manager/ManagerDeviceList.py
cschramm: `icon_info` is already defined as non-optional in the method arguments, so we cannot assign the optional value from `get_icon_info` to it, but I could actually save `icon_info = _icon_info` as I could just use `_icon_info.load_surface` after the assertion directly.

blueman/gui/manager/ManagerDeviceList.py
cschramm: > This should be fixed in the stub that load_surface return an instance of cairo.Surface.
What do you mean? I think that's the case there, but we expect that `Surface` to be an `ImageSurface` which is not guaranteed.

blueman/plugins/manager/Info.py
cschramm: `_window` first gets implicitly defined as optional, the assertion narrows it down to non-optional, but if we use it in the lambda it will be considered optional again (https://github.com/python/mypy/issues/2608). Opposed to that, `window` gets defined as the narrowed type of `_window` so it is non-optional right away.

blueman/main/PPPConnection.py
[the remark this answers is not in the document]
infirit: @infirit sad now.. Ok then :+1:

infirit (approved):

infirit:
Apologies it took so long. I have been busy and my motivation was very low to work on blueman.

Issue #1278, opened 2020-06-15
blueman-applet tooltip shows "Bluetooth Enabled" even with bluetooth disabled


fdmarcin:
blueman: 2.1.3-1
BlueZ: 5.54-2
Distribution: Manjaro
Desktop environment: XFCE 4.14.2

`blueman-applet` tooltip always says "Bluetooth Enabled".

The string `"Bluetooth Enabled"` appears in [ShowConnected.py](https://github.com/blueman-project/blueman/blob/master/blueman/plugins/applet/ShowConnected.py#L60) and [StatusIcon.py](https://github.com/blueman-project/blueman/blob/master/blueman/plugins/applet/StatusIcon.py#L32) and all the `.po` files.

![image](https://user-images.githubusercontent.com/25668969/84664257-901e3580-af1e-11ea-9de6-2523ab47dfe9.png)

When running `blueman-applet` from terminal (log below), the error complains about missing `AppIndicator3`, so this might be at fault.

I have `libappindicator-gtk2` installed, and all my status indicators have worked correctly.

<details><summary>Log: starting <code>blueman-applet</code> from terminal</summary>

```shell
blueman-applet version 2.1.3 starting
Stale PID, overwriting
blueman-applet 15.46.57 ERROR    PluginManager:90 load_plugin: Unable to load plugin module AppIndicator
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/blueman/plugins/applet/AppIndicator.py", line 7, in <module>
    require_version('AppIndicator3', '0.1')
  File "/usr/lib/python3.8/site-packages/gi/__init__.py", line 129, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace AppIndicator3 not available

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/blueman/main/PluginManager.py", line 88, in load_plugin
    importlib.import_module(self.module_path.__name__ + ".%s" % plugin)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
[...]

PR #1358, opened 2020-08-23
ShowConnected: Fix tooltip when bluetooth is disabled

infirit:
See #1278 for discussion. This fixes the problem for me and it is an obvious copy paste error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Searching (with Ctrl+F in manager device list) did not work
 * Default PIN lookup
 * Exceptions from asynchronous DBus calls (getting picked up by tools like Apport or ABRT)
+* Fix and properly update tooltip on blueman-tray
 
 ## 2.1.1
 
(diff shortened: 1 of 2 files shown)


cschramm:
See https://github.com/blueman-project/blueman/commit/87880ba4fa71db733703775b864c56a6843e4e1f#diff-868994f6d339fd38450b99bcaa0ec495L52.

It was:
```
try:
  if powered:
    set "Bluetooth Enabled"
except PowerManagerNotLoaded:
  set "Bluetooth Enabled"
```
i.e. it did not change anything if PowerManager was loaded and Bluetooth was not enabled.
You wanted to change that to:
```
if PowerManagerLoaded:
  if powered:
    set "Bluetooth Enabled"
else:
  set "Bluetooth Enabled"
```
Same behavior.
You actually changed it to:
```
if PowerManagerLoaded:
  if powered:
    set "Bluetooth Enabled"
  else:
    set "Bluetooth Enabled"
```
Now it sets "Bluetooth Enabled" even if the Bluetooth is not enabled but does not do anything anymore if PowerManager is not loaded.

infirit:
Ok, I had to think this through a bit and I see what has happened. I will change the original logic a bit but it will have the same result.

Also, I don't think we should show anything when PowerManager is not available. This is just confusing so better not show anything.

infirit:
Updated the PR and added a a commit to handle PowerManager being (un)loaded.

cschramm (approved):

Issue #1371, opened 2020-09-03
Status indicators


cschramm:
This is a follow-up to https://github.com/blueman-project/blueman/issues/1362#issuecomment-686307938.

Linux tray icons / status indicators / whatever you want to call them are a mess. I'm trying to collect what I think to know about that mess.

There are basically two common ways to implement them: The old [XEmbed protocol](https://specifications.freedesktop.org/xembed-spec/xembed-spec-latest.html) based on X11 and the newer [StatusNotifierItem (SNI)](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/) based on DBus.

The two indicator implementations that blueman currently provides are based on GtkStatusIcon and libappindicator. GtkStatusIcon is a nice XEmbed implementation. GTK 4 will drop it (we could still use GTK+ 3 GtkStatusIcon, though, as blueman-tray is a small, mostly decoupled, standalone application now). libappindicator is an unmaintained piece of legacy software that has a bad SNI implementation and a bad fallback XEmbed implementation and evolved into some kind of industry standard for (non-Qt, non-kdelibs) SNI clients :man_facepalming:. There's also Ayatana Indicators which is said to a be drop-in replacement for libappindicator although it's not and it does not seem more advanced than libappindicator itself. :man_shrugging: It does have some adoption though, at least in Debian (see #1217) and MATE (see [mate-indicator-applet](https://github.com/mate-desktop/mate-indicator-applet)).

As XEmbed is tightly coupled to X11 and Wayland is a thing nowadays, mainstream desktops started dropping support for it. KDE, who initially drafted SNI for Plasma 4, dropped support in Plasma 5 (there is an xembed-sni-proxy thingy since 5.5, though). Ubuntu picked up KDE's SNI protocol to build Application Indicator (and the infamous libappindicator) on it and dropped XEmbed support in Unity. Let's not talk about Gnome Shell when it comes to trays...

Instead, let's turn to our target group where it's rather a question of who does not yet support SNI than of who does not support XEmbed (anymore):
* xfce4-panel: xfce4-indicator-plugin since like forever (4.4?), xfce4-statusnotifier-plugin since 1.14
* mate-panel: Builtin support since 1.18 / 1.20, also mate-indicator-applet
* lxqt-panel: plugin-statusnotifier since 0.10.0
* Cinnamon since 2.8
* LXPanel: Doesn't seem like it has support (is LXDE still relevant, though?)
[...]

PR #1382, opened 2020-09-13
Add StatusNotifierItem indicator implementation

cschramm:


infirit:
Let me test this on several desktops to see how it behaves. Did you figure out if making text bold is supported?

cschramm:
> Did you figure out if making text bold is supported?

mate-panel does support our bold tooltip, see https://github.com/blueman-project/blueman/issues/1371#issuecomment-691723632. :slightly_smiling_face:

infirit:
This is not working very well so far when I use xfce4-statusnotifier-plugin. I see a tooltip but no menu. Blueman also crashes the statusnotifier-plugin when updating the tooltip. I'll try to narrow it down next week

cschramm:
https://gitlab.xfce.org/panel-plugins/xfce4-statusnotifier-plugin/-/blob/master/panel-plugin/sn-item.c looks like I'll have to add `ItemIsMenu = False` to make it activatable.

It uses libdbusmenu. Maybe we should use that too for the menu. :man_shrugging: It has GI bindings.

From testing with `gi.repository.DbusmenuGtk3.Menu.new("org.blueman.Tray", "/org/blueman/sni/menu")` it looks like the lib uses `GetGroupProperties`. A quick (and incomplete) solution should be to add

```
self.add_method("GetGroupProperties", ("ai", "as"), "a(ia{sv})",
                lambda ids, props: [(idx, self._render_item(item)) for idx, item in enumerate(self._items, 1)])
```

to `MenuService.__init__`.

infirit:
Thx, I'll try that out.

Using libdbusmenu will indirectly bring the deprecated libdbus-glib back
and we just got rid of it. This also brings me to the maintenance as I
don't expect them to move to gio for this anytime soon.

cschramm:
Why deprecated?

libdbusmenu is actually the goto reference implementation for (Canonical) D-Bus menus. I just avoided it as it would not really reduce the amount of code on our side but I really simplified things (left out methods, which seems to bite me now, ignored recursion parameters and the like) where the lib would sure make a better job.

infirit:
I am talking about dbus-glib what dbusmenu is using , https://dbus.freedesktop.org/doc/dbus-glib/

```
dbus-glib is a deprecated API for use of D-Bus from GLib applications. Do not use it in new code.

Since version 2.26, GLib's accompanying GIO library provides a high-level API for D-Bus, "GDBus", based on an independent reimplementation of the D-Bus protocol. The maintainers of D-Bus recommend that GLib applications should use GDBus instead of dbus-glib.
```

As for maintenance of libdbusmenu, last release is over 4 years old and what I understood debian was moving to AyatanaIndicators which also forks libdbusmenu. And I see neither move to Gio anytime soon so we end up with using a deprecated dbus library.

Just like libappindicator, there is no real maintenance.

infirit:
Wait, this is very confusing. Ubuntu lists dbus-glib as a build dependency [1] but libdbusmenu moved to GDbus in 2011. Is Ubuntu stoned or am I 😕.

[1] https://packages.ubuntu.com/source/focal/libdbusmenu

cschramm:
Ubuntu usually is, yes. 😄

infirit (approved):
Nothing stands out so let's merge and deal with any issues that come up.

infirit (approved):

cschramm:
From #1537 I added a proper `GetGroupProperties` implementation to make this work with snixembed. `GetLayout` and `GetGroupProperties` seem completely redundant to me, but obviously different implementations heavily depend on one or the other.

snixembed also brought my attention to an interesting detail of the SNI registration: https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierWatcher/ actually specifies to pass a service name to `RegisterStatusNotifierItem`. What my implementation as well as libappindicator and Ayatana pass is an object path.

The difference seems to come from the history of the interface: appindicator used to pass a service name and just use the sender of the `RegisterStatusNotifierItem` call as the service to talk to. KNotifications, it seems, used to pass a service name and expect to find a fixed `/StatusNotifierItem` object there. Funny enough the specification says nothing about that object path, so based on that I don't see how a proper implementation could be done.

The implementations I looked at support both by checking if the argument is a service name or an object path:

snixembed: https://git.sr.ht/~steef/snixembed/tree/0cd2d89df586c4fa3cf1d4135ec961f9d88cc501/item/src/proxyserver.vala#L22
mate-panel: https://github.com/mate-desktop/mate-panel/blob/ce8167e1b498e47ca6f667f83acaf21d407135bb/applets/notification_area/status-notifier/sn-host-v0.c#L86
xfce4-statusnotifier-plugin: https://gitlab.xfce.org/panel-plugins/xfce4-statusnotifier-plugin/-/blob/c6f7aeec13dacc931b0bd74644bea8a8a1a47389/panel-plugin/sn-backend.c#L359

mate-panel seems to provide a third way: If the argument contains a `/` it takes everything in front of it as the service name and the rest as the object path. I do not get how passing only the object path works with that, as the service name should end up empty, but I can tell it does. :shrug:

PR #1383, opened 2020-09-15
Add oc.po

cschramm:
This was requested on Weblate (Weblate does not offer to add new translations if the POT file is not in the repository).

### DIFF
--- a/po/oc.po
+++ b/po/oc.po
@@ -1,0 +1,2571 @@
+# Occitan translations for blueman package.
+# Copyright (C) 2020 Copyright © 2008 - 2020 blueman project
+# This file is distributed under the same license as the blueman package.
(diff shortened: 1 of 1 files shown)


infirit (approved):

infirit:
Maybe add the pot file anyway 😜

infirit:
Forgot to add, when we actually get translated string we ned to update LINGUAS.

cschramm:
> Maybe add the pot file anyway :stuck_out_tongue_winking_eye: 

We could keep it up to date manually with every MR that effects it and add a check that it actually is up to date. :man_shrugging: 

> Forgot to add, when we actually get translated string we ned to update LINGUAS.

:thinking: cy has been missing there as well all the time.

PR #1390, opened 2020-09-26
Restore blueman.pot

infirit:
Having blueman.pot has its benefits like,

* Allow weblate to add new translations
* No more magic artifacts to avoid pot creation date/time commits
* No need to squash commits anymore.

I do understand it is not really necessary but it makes the integration with weblate easier.

### DIFF
--- a/.github/workflows/weblate.yml
+++ b/.github/workflows/weblate.yml
@@ -11,17 +11,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - run: sudo apt-get install autopoint python-gi-dev cython libglib2.0-dev libbluetooth-dev gettext
-      - run: ./autogen.sh
-      - env:
-          TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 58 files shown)


cschramm:
> Update translations

We should leave that one to Weblate (or at least merge #1323 right before this one). Otherwise merge conflictst, lost translations, :boom:, ...

I do see the benefits and am fine with it but we should add a check to the check workflow that verifies the POT file is up to date so that it cannot fall behind.

cschramm:
A job like this should work for the check:

```
  potfile:
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2
      - run: sudo apt-get update
      - run: sudo apt-get install -y -qq --no-install-recommends autopoint gettext cython libglib2.0-dev python-gi-dev libbluetooth-dev
      - run: ./autogen.sh
      - run: make -C po blueman.pot-update && git diff-index --exit-code HEAD -- po/blueman.pot || (echo "blueman.pot is not up to date" && exit 1)
```

We'll also have to port this to 2-1-stable as `make blueman.pot` always updates the file there, opposed to `make blueman.pot-update` so that the weblate workflow's caching does not work.

PR #1407, opened 2020-10-14
Update GTK+ 3 dependency

cschramm:
With the use of gtk_menu_popup_at_pointer we actually require 3.22


infirit (approved):
Oops, good to see one of us is paying attention 😁

PR #1413, opened 2020-10-29
Allow cancelling device connection attempts

cschramm:
I realized that this is easy to achieve. It would also be the same for network services but it gets tricky for serial services.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,21 +8,46 @@
 * Drop blueman-assistant
 * Raise minimum Python version to 3.6
 * Raise GTK+ 3 version to 3.22
-* Do not use exitcode 1 when we expect to fail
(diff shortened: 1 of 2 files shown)


infirit (approved):
If I am not mistaken, any error from DisconnectService will bubble up to the UI, right?

cschramm:
As I'm doing a synchronous call, errors would raise exceptions in the `abort` function, but the only error that would be expected (except for generic issues with D-Bus communication) is `org.bluez.Error.NotConnected` according to https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/device-api.txt and I do even get that for disconnected devices, Disconnect just always succeeds.

infirit:
Cool, approved already so feel free to merge :+1:

PR #1414, opened 2020-10-29
Add auto-connect items to device menu

cschramm:
The first four commits are refactoring and improvements for the service interface and menu handling.

Then comes the actual implementation for https://github.com/blueman-project/blueman/issues/239#issuecomment-375131266 - finally :tada:. As the wizard is no more, menu items seem like the only logical place to do this. The last commit is kind of an optional idea to keep the menu lean, but I think it makes sense. The only downside I see is that people might not open the menu a second time after connecting a service when looking for a way to auto-connect.

### DIFF
--- a/blueman/config/AutoConnectConfig.py
+++ b/blueman/config/AutoConnectConfig.py
@@ -1,0 +1,29 @@
+import gi
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk
+
+from blueman.bluez.Device import Device
+from blueman.main.Config import Config
+
+
+class AutoConnectConfig(Config):
+    def __init__(self) -> None:
+        super().__init__("org.blueman.plugins.autoconnect")
+
+    def bind_to_menuitem(self, item: Gtk.CheckMenuItem, device: Device, uuid: str) -> None:
+        data = device.get_object_path(), uuid
+
+        def switch(active: bool) -> None:
+            services = set(self["services"])
+            if active:
+                self["services"] = set(services).union({data})
+            else:
+                self["services"] = set(self["services"]).difference({data})
+
+        def on_change(config: AutoConnectConfig, key: str) -> None:
+            if key == "services":
+                item.props.active = data in set(config[key])
+
+        item.props.active = data in set(self["services"])
+        item.connect("toggled", lambda i: switch(i.props.active))
+        self.connect("changed", on_change)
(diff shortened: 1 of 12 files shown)


cschramm:
@infirit: Should I split off some of the initial commits to bring this forward?

blueman/config/AutoConnectConfig.py
```diff
@@ -0,0 +1,29 @@
...
+
+from blueman.bluez.Device import Device
+from blueman.main.Config import Config
+
+
+class AutoConnectConfig(Config):
+    def __init__(self) -> None:
+        super().__init__("org.blueman.plugins.autoconnect")
+
+    def bind_to_menuitem(self, item: Gtk.CheckMenuItem, device: Device, uuid: str) -> None:
```
infirit, line 13: This is confusing as it is not related to menuitems, more to the services key in gsettings. I would suggest to name it ``bind_service`` instead.

blueman/config/AutoConnectConfig.py
```diff
@@ -0,0 +1,29 @@
...
+            else:
+                self["services"] = set(self["services"]).difference({data})
+
+        def on_change(config: AutoConnectConfig, key: str) -> None:
+            if key == "services":
+                item.props.active = data in set(config[key])
+
+        item.props.active = data in set(self["services"])
+        item.connect("toggled", lambda i: switch(i.props.active))
+        self.connect("changed", on_change)
```
infirit, line 29: My suggestion for all of this is to make ``org.blueman.plugins.autoconnect`` relocatable and use the bluetooth address in the path, similar to org.blueman.plugins.netusages. Have in the relocatable schema keys for each uuid you want to support with a boolean as value.

With that you can use the already existing bind/bind_to_widget :smile:.

infirit (commented):
Have not tested or looked very closely. The only thing I like to discuss is related to how this is currently stored in GSettings. It makes it unnecessarily complicated to bind to widgets.

blueman/config/AutoConnectConfig.py
cschramm: I just tried that but using the UUIDs as GSettings keys gets super ugly. Keys have to start with a letter, so the UUIDs need a prefix and I need to access all in the AutoConnect applet plugin. :thinking:

infirit: That sucks. AFAIK the xml schema does not mandate it, is the error from dconf or gsettings? If you need to manipulate things the previous approach is better then I guess.

infirit:
Let's just merge this and see how things go. Also other thins needed for the next release and ill make time to test what I can.

blueman/config/AutoConnectConfig.py
cschramm: `glib-compile-schemas` gives shows error `names must begin with a lowercase letter` and ignores the file.

infirit (approved):

Issue #1374, opened 2020-09-05
Feature Request: desktop notifications for connect / disconnect


hissssst:
First of all, thank you for this awesome software

I'd like to open discussion about desktop notifications with blueman.
It would be nice to have notifications sent when device connects/disconnects

infirit:
I am not against adding a notification but not without a proper use case. Can you be more specific in which situation this may be useful and how it would help you?

hissssst:
I would like to receive notifications when device connects/disconnects. It's useful because I am connected to multiple bt-headsets at one time and some of them can disconnect when they are out of battery.

It would be nice to receive notification like `Device Headset Pro (12:34:56:78:ab:cd) disconnected`

cschramm:
Makes perfect sense to me. Usually you do not need it as the status icon change is sufficient if you use just one device at a time but you lack that feedback if you use multiple.

Issue #947, opened 2018-12-01
feature request: show battery status of device


treaki:
for the version see my previews issue #945

When i am connected to my wireless headphones using my android phone (running open source lineage os) i can see its battery status (little bar next to the Bluetooth icon in the notification bar). It would be very grate if blueman-manager could also show this next to the three bars at the right of the devie (package lost, receive and send signal (or so))

thanks in advance and keep up the good work

infirit:
This information is not exposed over the bluez dbus api. Recently for LE device https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/battery-api.txt was added, it is however not very reliable for the LE devices (always 100%) I have. I don't see this happen anytime soon.

cschramm:
We should be able to read the value from the GATT API https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/gatt-api.txt, Battery Service (0x180F), Battery Level Characteristic (0x2A19). If I'm not mistaken, I have a mouse that provides it. Will check.

infirit:
Yes, this is what is exposed by the battery api and is LE only :smile: We have nothing exposed for older devices.

shanmukhateja:
I'd like to state that certain wireless headsets which support iOS devices have non-standard AT commands sent in HSP for battery indication.

There are ongoing talks over at PulseAudio for ways to support this feature [link](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/issues/722).

Android vanilla has added support for this recently [link](http://androidxref.com/9.0.0_r3/xref/packages/apps/Bluetooth/src/com/android/bluetooth/btservice/RemoteDevices.java#733). 

Also, bluez-alsa project has it as well. [link](https://github.com/Arkq/bluez-alsa/blob/ed3f513a1f85d38daa4a80a868686bd709d1ac68/src/rfcomm.c#L443).

Fry-kun:
More devices support a reasonable battery level indication nowadays, please consider adding something here.
I'm sure I'm not the only one who's been using BT earbuds with a laptop lately.. they report their charge level to my phone perfectly fine, but there's nothing on my laptop :(

joker234:
This project works very well with my devices, but it would be nice to use the existing connection for these information and display this in the blueman-applet. https://github.com/TheWeirdDev/Bluetooth_Headset_Battery_Level

kred-jon:
still waiting default indicator from blueman, think its a must now.

[...]

Issue #860, opened 2018-04-15
Decide on signal bars


infirit:
With bluetooth LE we do not have signal bars. The problem afaics is that we do not have tpl, lq and rss. We do have an rssi property in the dbus api but it is only available during discovery.

While it has been a big part of blueman the simplest would be drop them and instead have an indicator showing if the device is connected. I can add the indicator on top of the bars as well but this will make the ui inconsistent between LE and non LE devices.

Opinions, suggestions, complaints :-)

cschramm:
My opinions on this:
* Connection indicator definitely makes sense
* Inconsistent UI between LE and non LE is fine

The only way to avoid the inconsistency seems to be to not use the information we have for non LE devices. The signal bars are probably not that useful to most users, but I'd not downgrade that UI for consistency reasons.

infirit:
I used an image upscaler (waify2x) and the result is pretty good. Imho this looks good enough for Gdk scaling. I'll do them all tomorrow and test it out.
![blueman-lq-90_waifu2x_art_scale_tta_1](https://user-images.githubusercontent.com/3465730/98450681-dfb80480-213e-11eb-9b84-586d163ed63e.png)

infirit:
Here is the result with GDK_SCALE=2. It is pretty amazing how well this ai upscaler works :astonished: . I'll open a PR for this.

![afbeelding](https://user-images.githubusercontent.com/3465730/98466265-b3e45f80-21ce-11eb-8b49-a5c593388c7a.png)

infirit:
I'm leaning towards using using a Gtk.LevelBar as seen in #1768. But this requires quite a bit of work replacing the TreeView.

PR #1417, opened 2020-11-02
Add battery levels to manager device list

cschramm:
This is a first approach to #947 and along the way eases the problem in #860 as it does no longer show fake signal levels if a battery level is available.

The pixmaps are super temporary. I hope I'll find an easy way to build this dynamically with icon theme icon + bar + color. Otherwise I'm a little lost as I have no clue how the existing pngs got created.

I'm also working on / thinking about an applet plugin that shows a notification with battery level, probably in combination with #1374. One cannot couple it with the Connected property, though, as the Battery interface is not available initially (the ServicesResolved signal should probably work).


infirit:
I played around with inkscape a bit (to eventually recreate the rssi/tpl/lq pixmaps) and used https://openclipart.org/detail/303646/fully-charged-battery to make below. I have them in all steps. If you are going to use default icons you are at the mercy of icon themes which may be fine but I have seen some broken icon themes 😨 .

cschramm:
Looks nice. :ok_hand:

cschramm:
How shall we proceed with this? I'd wait for #1432 and then rebase this according to the new rendering scheme. Shall we create a common branch here where you can contribute the icons or shall we do two independent PRs with icons and logic?

infirit:
Here are all of them, 10 - 100 zipped. They can be made into png and then we are dealing bytes in size.

[battery.zip](https://github.com/blueman-project/blueman/files/5648061/battery.zip)

cschramm:
Nice. :+1: I added them as PNGs. Shall we add the SVGs to the source as well? Somewhere like `data/icons/pixmaps/sources`?

infirit:
Ignore my previous comment, I forgot how ManagerDeviceList works :eyes: . Using the png's that are there now is good enough. 

And yes, store the svg's somewhere we can find them later. I don't think we should install them, we could later on if we decide to use them in other places.

infirit (approved):

PR #1418, opened 2020-11-02
Add a battery level applet plugin

cschramm:
Another one for #947. The plugin assumes that the battery service interface is available when services got resolved and the battery service UUID is provided, so this requires BlueZ 5.48 that introduced that service interface. I think the plugin manager would handle the resulting error gracefully and if not we could still catch it or check the BlueZ version beforehand but I guess it's fine to require a three year old minimal version of BlueZ.


blueman/bluez/Base.py
```diff
@@ -40,7 +40,7 @@ class Base(Gio.DBusProxy, metaclass=BaseMeta):
 
     _interface_name: str
 
-    def __init__(self, obj_path: str):
+    def __init__(self, *, obj_path: str):
```
infirit, line 43: Why is this needed?

blueman/plugins/applet/Battery.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+from blueman.Sdp import ServiceUUID, BATTERY_SERVICE_SVCLASS_ID
+from blueman.bluez.Battery import Battery as BluezBattery
+from blueman.bluez.Device import Device
+from blueman.gui.Notification import Notification
+from blueman.plugins.AppletPlugin import AppletPlugin
+
+
+class Battery(AppletPlugin):
```
infirit, line 10: Please add author and description :smile:

blueman/bluez/Base.py
cschramm: `BaseMeta.__call__` expects to find `obj_path` in the keyword arguments, so if you pass it as a positional argument, the whole thing fails.

infirit (approved):

PR #1419, opened 2020-11-02
Restore blueman.pot

cschramm:
I picked up #1390 without the manual po update and with the check job.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -78,3 +78,12 @@ jobs:
       - run: make -C module
       - run: touch /dev/rfkill
       - run: PYTHONPATH=module/.libs python3 -m unittest
+
+  potfile:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo apt-get update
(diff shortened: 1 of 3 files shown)


infirit (approved):
lgtm

cschramm:
And I immediate ran into a problem: gettext >= 0.20.2 does not extract Icon names from desktop files anymore and the ubuntu thingy is on 0.19.8, so it still does. Depending on your local gettext version it might or might not detect a difference due to that. I think I'll add a diff output to the job so that it's visible why it detects a difference and check if I can get >= 0.20.2 into the job so that it matches recent versions. That still means one needs to have >= 0.20.2 locally when updating the file (or remove the icon names that get added otherwise by hand afterward).

infirit:
Force the 0.20.2 requirement in configure.ac ``AM_GNU_GETTEXT_VERSION([0.20.2])``

infirit:
> Force the 0.20.2 requirement in configure.ac `AM_GNU_GETTEXT_VERSION([0.20.2])`

And it is not in debian so this may not be a good idea :thinking:

PR #1425, opened 2020-11-05
Fix autoconnect closure and improve plugin

cschramm:


blueman/plugins/applet/AutoConnect.py
```diff
@@ -37,11 +41,11 @@ def _run(self) -> bool:
             if device is None or device.get("Connected"):
                 continue
 
-            def reply() -> None:
-                assert isinstance(device, Device)  # https://github.com/python/mypy/issues/2608
+            def reply(dev: Optional[Device] = device, service_name: str = ServiceUUID(uuid).name) -> None:
+                assert isinstance(dev, Device)  # https://github.com/python/mypy/issues/2608
```
infirit, line 45: I am puzzled why mypy does not flag this if dev is optional :confused:

infirit (approved):
lgtm

blueman/plugins/applet/AutoConnect.py
cschramm: It's not a bug, it's a feature. You can use assertions to narrow down types, so if `a` is `Optional[A]`, `assert a is not None` makes it `A` from that point on. mypy does not prevent `AssertionError`s due to type checks. An `AssertionError` because `a` is actually `None` at the assertion is absolutely expected and allowed. The assertion would not make any sense anyway if it's already clear by static checks that it cannot fail.

PR #1426, opened 2020-11-05
Update python versions

cschramm:


cschramm:
Installing cython currently fails for 3.10-rc, see https://github.com/pypa/wheel/issues/354 / https://github.com/python/cpython/pull/20333 / https://github.com/python/cpython/pull/22858 but I'll leave those jobs optional and disabled the default fail fast configuration that cancels other jobs from the matrix if one fails. Leaves an ugly red cross on the pull requests for now as GitHub does not visually distinguish between required and non-required failed jobs.

infirit (approved):

PR #1427, opened 2020-11-05
Add plugin with disconnect menu items

### LINKED ISSUE #1268: Disconnect from the applet
I love the project!

My only UX suggestion would be to add Disconnect all (or disconnect and a list of connected devices) to the applet.

I often switch my headset between phone and laptop and having a really simple way to disconnect would really improve my day-to-day experience.

cschramm:
Closes #1268


infirit (approved):
lgtm

PR #1431, opened 2020-11-07
Improve pot handling

cschramm:


infirit (approved):
Lol, anyway drop the po/ part and I suspect it will be ok.

infirit (approved):

cschramm:
:thinking: Shall we add `--no-location` again to reduce the necessity for pot updates and save us from merge conflicts and the like?

PR #1432, opened 2020-11-08
Upscale and fix last bits for hidpi displays

infirit:
Ideally we have vector replacements and generate proper icons but for now this is imho good enough


infirit:
So, what to do with the mypy error as they are actually the same :confused:. 
```python
$ MYPYPATH="stubs" python3 -m mypy -p blueman --strict
blueman/gui/manager/ManagerDeviceList.py:561: error: Incompatible types in assignment (expression has type "gi.repository.cairo.Surface", variable has type "cairo.Surface")
```

blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -407,20 +410,23 @@ def on_finished(fader: CellFade) -> None:
...
                 faderhandler = fader.connect("animation-finished", on_finished)
 
+            w = 14 * self.get_scale_factor()
+            h = 48 * self.get_scale_factor()
+
             to_store = {}
             if round(row["rssi"], -1) != round(rssi_perc, -1):
-                icon_name = "blueman-rssi-%d.png" % round(rssi_perc, -1)
-                icon = GdkPixbuf.Pixbuf.new_from_file(os.path.join(PIXMAP_PATH, icon_name))
+                icon_name = "blueman-rssi-%d.png" % round(rssi_perc)
```
cschramm, line 418: The -1 seems to be missing here.

cschramm:
What I did in 9093d0c4de8021d0419d916d73d7b9a15b941fdb is simply replace `from gi.repository import cairo` with `import cairo` in the Gtk stub file. I guess the easiest solution is to do the same in the Gdk file as well (and GdkX11 and Pango if you like).

This is not complete yet, right, as the four new files do not seem to be used?

blueman/gui/manager/ManagerDeviceList.py
infirit: Ooh, good someone is paying attention 😁. I'll fix it later today.

infirit:
> replace `from gi.repository import cairo` with `import cairo` in the 

Yes, I think this is correct. I'll add a commit to deal with this in all stubs. There is cairo-gobject but this is only used in specific cases (like storing a Cairo surface inside a liststore) and I never got these  working with python.

> This is not complete yet, right, as the four new files do not seem to be used?

These are automatically selected and used through gtk's icon handling. I'll post a screenshot with and without so you can see the difference.

cschramm:
> These are automatically selected and used through gtk's icon handling.

Whaaat? How? :thinking:

infirit:
GDK_SCALE=2 with 32x32

manager 32x32 activity

GDK_SCALE=2 with only 16x16

manager 16x16 activity

> > These are automatically selected and used through gtk's icon handling.
> 
> Whaaat? How?

When using icon names, like we do with the activity icons, Gtk will select the best size for a given scale factor. Have you ever noticed that sometimes icons are different when using GDK_SCALE=2?.  That is because many icon themes have different icons based on the size. For example see these 2 below which I posted when I started working on this, notice how the paired icon is different :smile:.

manager_gdk_scale_2_before


manager_gdk_scale_2

cschramm:
Now I got it. Those files already existed in 16x16. :facepalm:

cschramm (approved):

cschramm (approved):

cschramm:
Bildschirmfoto zu 2020-11-12 01-03-40
Something looks off there. :thinking: (they're all 100 %)

infirit:
Now that I force the size it is likely more noticeable. I'll look into this this weekend.

Issue #1363, opened 2020-08-24
Contextualize and internationalize errors


ymollard:
Following https://github.com/blueman-project/blueman/pull/1347#issuecomment-679111971, here's a discussion about error messages such as in the screenshot below.

The point is to figure out how we could make errors more understandable both in terms of content (Contextualize kernel errors) and form (translate them into the language of the UI), with a minimum effort and maintenance need.

![image](https://user-images.githubusercontent.com/3295098/91100243-1bc1ea00-e665-11ea-8577-d02fb45f7b46.png)

Here's a list of common errors reported in [~800 Github issues](https://github.com/blueman-project/blueman/issues):
* `Protocol not available`
* `No such file or directory`
* `Input/output error`
* `Host is down`
* `Resource temporarily unavailable` (e.g. incorrect/cancelled pairing key)
* `Software caused connection abort` (e.g. Bluetooth turned off during connection)
* `Resource not ready`

cschramm:
Those are
* ENOPROTOOPT
* ENOENT
* EIO
* EHOSTDOWN
* EAGAIN
* ECONNABORTED

I fully agree that the standard `strerror` messages are rather misleading in these contexts.

As discussed in #1347 we are left with matching the error messages as we do not have the error codes, so I guess somebody should at least check musl if it uses the same descriptions for those codes.

We also need to gather the contexts that these errors show up in to find proper descriptions. Are they all from that same "Connection Failed" context?

ymollard:
> We also need to gather the contexts that these errors show up in to find proper descriptions. Are they all from that same "Connection Failed" context?

It's hard to say, I only grep-ed a bunch of issues searching for dbus errors. How can we figure out?

I've just edited the initial message by adding `Resource not ready` if you wanna extend the list of errnos in your message ; I got this one today ([context here](https://github.com/blueman-project/blueman/issues/1368))

cschramm:
I dug into this in the context of generic device connection (opposed to network or serial service connection, discovery, cancellation etc.) a bit and collected the following about those messages so far:

`Protocol not available` is the message for ENOPROTOOPT in glibc, uClibc, and musl and means that no audio endpoints are registered to BlueZ (which is typically done by pulseaudio's Bluetooth module).

`Input/output error` is the message for EIO in glibc and uClibc. The musl message for it is `I/O error`. It means all sorts of low level stuff going wrong and oftentimes hints at driver / firmware issues.

[...]

PR #1440, opened 2020-11-22
Translate common error messages and ignore cancelled error

cschramm:
BlueZ obtains most of those messages from strerror, passing more or less appropriate error codes chosen by itself (e. g. ENOPROTOOPT for missing audio endpoints) or obtained from the kernel, which, amongst other things, translates Bluetooth error codes to more or less appropriate error codes in `bt_to_errno` (e. g. page / advertising timeouts to EHOSTDOWN).

By matching those messages to what is expected from glibc, uClibc, and musl we can use our own messages to clarify the context and localize them (strerror and and other BlueZ error messages are not localized).

See #1363 for the selected errors. Others do not seem relevant.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -174,7 +174,8 @@ class ManagerDeviceMenu(Gtk.Menu):
             prog.message(_("Failed"))
             self.unset_op(device)
             msg, tb = e_(result.message)
-            MessageArea.show_message(_("Connection Failed: ") + msg)
+            if msg != "Cancelled":
+                MessageArea.show_message(_("Connection Failed: ") + msg)
 
         def success(_obj: AppletService, _result: None, _user_data: None) -> None:
             logging.info("success")
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -168,14 +168,34 @@ class ManagerDeviceMenu(Gtk.Menu):
             if key in ("Connected", "UUIDs", "Trusted", "Paired"):
                 self.generate()
 
+    def _handle_error_message(self, msg: str) -> None:
+        msg = msg.split(":", 3)[-1].strip()
+
+        # https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/gnu/errlist.h
+        # https://git.musl-libc.org/cgit/musl/tree/src/errno/__strerror.h
+        # https://git.uclibc.org/uClibc/tree/libc/string/_string_syserrmsgs.c
+        if msg == "Protocol not available":
+            # ENOPROTOOPT
+            logging.warning("No audio endpoints registered to bluetoothd. "
+                            "Pulseaudio Bluetooth module, bluez-alsa, PipeWire or other audio support missing.")
+            msg = _("No audio endpoints registered")
+        elif msg in ("Input/output error", "I/O error"):
+            # EIO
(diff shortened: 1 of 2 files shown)


blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -168,13 +168,32 @@ def on_device_property_changed(self, lst: "ManagerDeviceList", _device: Device,
...
 
+    def _translate_error_message(self, msg: str) -> str:
+        msg = msg.split(":", 3)[-1].strip()
+
+        # https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/gnu/errlist.h
+        # https://git.musl-libc.org/cgit/musl/tree/src/errno/__strerror.h
+        # https://git.uclibc.org/uClibc/tree/libc/string/_string_syserrmsgs.c
+        if msg == "Protocol not available":
+            # ENOPROTOOPT
+            return _("No audio endpoints registered")
```
infirit, line 179: If I am not mistaken we connect a single endpoint not multiple.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -168,13 +168,32 @@ def on_device_property_changed(self, lst: "ManagerDeviceList", _device: Device,
...
+        elif msg in ("Input/output error", "I/O error"):
+            # EIO
+            return _("Input/output error")
```
infirit, line 182: Does "Communication error or problem"make sense here? If so how about using that 😄

infirit (commented):
Just a couple comments but lgtm overall.

blueman/gui/manager/ManagerDeviceMenu.py
cschramm: pulseaudio registers at least a source and a sink endpoint with BlueZ. Anyway, I'm actually not sure how much better this is to the user than "Protocol not available". Maybe something like "No Bluetooth audio support registered"?

blueman/gui/manager/ManagerDeviceMenu.py
cschramm: EIO is used for all sorts of shit. Just grep for it in the BlueZ source and you'll find >120 locations where it gets returned and I think about half of them can end up here as a DBus connection error. I'd say the meaning really is just that *something* went wrong more or less unexpectedly.

Occurences in our GitHub issues oftentimes have to do with firmware issues and the like.

I guess the best advice we could give is to check the bluetoothd logs as there's typically a warning or error level message that hints at what actually went wrong. Not sure if and how I'd place such an advice here, though. I wouldn't put it in this message, but we could think about a "More", "Help" or whatever link that gives some advice like that. For ENOPROTOOPT it could be to install the pulseaudio Bluetooth module, possibly even naming distribution specific package names but that sounds more like something that should go to a Wiki page we could link to.

infirit: 👀 ok then 😄. How about we keep it like it is but log something more helpful and refer users to that, logging.warning("Something unexpected happened, possible causes are.....")?

blueman/gui/manager/ManagerDeviceMenu.py
infirit: I forgot we have a source and a sink, so this lgtm 😄 .

blueman/gui/manager/ManagerDeviceMenu.py
cschramm: Good idea. Added. :+1:

infirit (approved):

PR #1457, opened 2020-12-04
Rename typing.py

infirit:
It conflicts with python's own typing module.


infirit:
It caused problems for me on artix now that they moved to python 3.9.

Would probably be a good idea to make the python 3.9 checks required :smile:

cschramm:
The GitHub checks are required and work fine. Referencing the module as `blueman.typing` is always fine. You only get into trouble if you run python in the blueman directory as `import typing` refers to the local module there that shadows the standard library's typing module in that case. I'm fine with changing it, but... what has Python 3.9 to do with that? :raised_eyebrow:

cschramm (approved):

infirit:
It errors bytcompiling on 3.9.

```python
ake[3]: Entering directory '/home/sander/aur/blueman/src/blueman/blueman'
make[3]: Nothing to be done for 'install-exec-am'.
 /usr/bin/mkdir -p '/home/sander/aur/blueman/pkg/blueman-git/usr/lib/python3.9/site-packages/blueman'
 /usr/bin/install -c -m 644 Constants.py DeviceClass.py Functions.py Sdp.py Service.py gobject.py typing.py __init__.py '/home/sander/aur/blueman/pkg/blueman-git/usr/lib/python3.9/site-packages/blueman'
  File "<string>", line 2, in <module>
  File "/usr/lib/python3.9/py_compile.py", line 9, in <module>
    import importlib.util
  File "/usr/lib/python3.9/importlib/util.py", line 2, in <module>
    from . import abc
  File "/usr/lib/python3.9/importlib/abc.py", line 17, in <module>
    from typing import Protocol, runtime_checkable
  File "/home/sander/aur/blueman/src/blueman/blueman/typing.py", line 1, in <module>
    from typing import Dict, Tuple, Union, TYPE_CHECKING
ImportError: cannot import name 'Dict' from partially initialized module 'typing' (most likely due to a circular import) (/home/sander/aur/blueman/src/blueman/blueman/typing.py)
make[3]: *** [Makefile:467: install-bluemanPYTHON] Error 1
make[3]: Leaving directory '/home/sander/aur/blueman/src/blueman/blueman'
make[2]: *** [Makefile:680: install-am] Error 2
make[2]: Leaving directory '/home/sander/aur/blueman/src/blueman/blueman'
make[1]: *** [Makefile:516: install-recursive] Error 1
make[1]: Leaving directory '/home/sander/aur/blueman/src/blueman/blueman'
make: *** [Makefile:558: install-recursive] Error 1
```

cschramm:
Weird. I can easily replicate the `ImportError` with importing any file that imports `typing` from an interactive interpreter in the blueman directory but I absolutely cannot with `py-compile` or `make install` or anything. Could it depend on the automake version and the precise py-compile script it installs? :thinking:

infirit:
Could be but I doubt it as with py3.8 there was no problem. We may just have been lucky it never blew up earlier :exploding_head:

cschramm:
The constraint sounds just super unexpected and wrong to me. It's basically something like "Python files that import standard or third-party modules may not have local modules with equal names next to them" and I do not really see a reason.

It only blows up if something imports those files having their directory in `sys.path` and that just shouldn't ever be the case and I'd expect more issues if that happens, e.g. relative imports will not work (`ImportError: attempted relative import with no known parent package`).

PR #1467, opened 2020-12-13
Improve service interface and Services plugin

cschramm:
This is split off from #1414 to get it rolling.

* Remove serial service submenu from device menu

The submenu seems rather pointless and random. It is only used if a DUN and / or multiple serial services are available. In the most usual case it only holds a connect item for the DUN service and the settings item. Disconnect items do not get added to the submenu.

* Replace connected with connectable as it is already misused as "not connectable" (serial services return False even if they are connected as they are still connectable). (Note that "connectable" is independent of "available" as it is used to show disabled connect items).
* Add connected_instances to provide per-service disconnect items.
* Add common_actions to provide actions that shall be shown if any of the services specifies them.

The additions allow to remove service-specific stuff from the Services plugin. An intentional change is that serial port disconnect items now use the service's blueman-serial icon instead of modem.

### DIFF
--- a/blueman/Service.py
+++ b/blueman/Service.py
@@ -1,8 +1,28 @@
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Optional, Callable, List, Set, Collection
 
 from blueman.Sdp import ServiceUUID
 from blueman.bluez.Device import Device
+
+
+class Instance:
+    def __init__(self, name: str, port: int = 0) -> None:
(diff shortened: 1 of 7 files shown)


infirit:
dun phones are becoming increasingly rare so I care little how it is presented. So feel free to merge and we can deal with any issues that may arise later.

infirit (approved):

Issue #1466, opened 2020-12-13
Bluetooth Group Network automatic connection inconsistent


rhyek:
blueman: 2.1.3-2ubuntu1.1
BlueZ: ?
Distribution: Lubuntu 20.10
Desktop environment: LXDE/LXQT

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

I believe the intended functionality for blueman so far is generally fine, but I have a question regarding my specific use case:

I have a new mini-pc that I intend to sometimes carry around during travel for development purposes mostly. It will be used mostly in a headless manner without any peripherals or monitor connected to it. I also carry around a MacBook Pro and what I need is for the mini-pc to reliably connect to the internet via a tether to my MacBook's bluetooth where I have "Internet Sharing" turned on.

With monitor and peripherals connected to the mini-pc, once the bluetooth is connected I can right click on my MPB device in the device list in blueman, then click on "Group Network" below "Connect:". This works great and internet works. The problem is that the connection to that network is not reliably recovered in cases such as rebooting the mini-pc, turning the bluetooth off and then on on the MBP (with or without disabling Internet Sharing beforehand), though it always reconnects to the network if I just disable/enable Internet Sharing on the MBP while the bluetooth is connected and remains so.

When the connection is not re-enabled in one of these cases, no amount of turning off/on the bluetooth or connecting and disconnecting on the MBP will trigger the "Connect to: Group Network" on the mini-pc and I have to do it manually somehow.

Can anything be done about this? Maybe a script that runs automatically when the bluetooth connects where I can trigger the connection manually via some command?

Thanks!

cschramm:
Duplicate of #239. There is an auto-connect "backend" in the current master (#853) and a pull request for the frontend parts in #1414.

rhyek:
@cschramm thank you for responding. I'll write this here as I don't want to pollute the linked issues.

I am not very familiar with this project, but I managed to somewhat install from source successfully so I could have the autoconnect plugin. After `make install` I was running into a few issues that I guess just had to do with a mismatched desktop environment or distro to your development environment (like the blueman module install location and others). I had given up and just ran `sudo apt install blueman`, but noticed the plugin was now available! So I guess I have it running.

[...]

PR #1469, opened 2020-12-14
AutoConnect: Run when new adapters become available

cschramm:
Triggered when Bluetooth gets unblocked or a new adapter gets added. We do not have to wait for the timer in those situations.

#1466


codeswhite:
Can we get this merged please?
BTW i also would like to see AutoConnect module run on startup

infirit (approved):

PR #1485, opened 2021-01-21
Stop discovery and retry connection on initial EAGAIN

### LINKED ISSUE #1398: Feature request: Connect to device while scanning
In `blueman-manager` GUI version 2.1.3: If you are  explicitly scanning for devices, and you find a connectable device, then attempting to connect to that device will fail since the scan is ongoing. 

Attempting to connect to a device via the GUI while scanning should instead stop the scanning, and attempt a connect instead.

cschramm:
There are drivers that do not support connecting while a discovery is still active.

Closes #1398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Disconnect items in applet menu (plugin)
 * Desktop notifications on connect / disconnect (plugin)
 * Notifications with battery level for connecting devices (applet plugin)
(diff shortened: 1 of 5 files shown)


infirit:
tbh I thought we already stopped discovery when we connect, but this likely is most of my devices take a but of time to pair and then connect. It would be nice if Thalley can confirm this approach works for him but other than that lgtm.

infirit (approved):

infirit (dismissed):

Issue #1495, opened 2021-02-13
GSettings option to hide the status/indicator icon


tari01:
blueman: 2.1.4
BlueZ: 5.55
Distribution: Manjaro
Desktop environment: MATE

I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Currently we have no mechanism that would hide the blueman-applet icon (in either SNI or legacy mode), while keeping the daemon running. It would be really useful if the applet could expose a GSettings option that would allow the user to hide/show the applet.

cschramm:
As you found in the other thread, the blueman-applet process is vital. What is not and actually displays the icon is the blueman-tray process. It gets auto-started by the StatusIcon plugin.

We thought about making the StatusIcon plugin optional in #949. That is some effort, though, as some other parts rely on its presence.

What's the motivation behind this?

tari01:
Thank you for your reply.

> As you found in the other thread, the blueman-applet process is vital.

Yes. I see no reason why anyone would want to disable blueman-applet. It is a very stable all-in-one backend.

> We thought about making the StatusIcon plugin optional in #949

:+1: 

> What's the motivation behind this?

Applications that provide an alternate indication mechanism would need the daemon running, but would also like to have a means of disabling the icon to avoid confusion/double indicators.

cschramm:
Looks easier than I thought. #1496.

> Applications that provide an alternate indication mechanism

To be honest, I'm not sure if I like that as the icon provides functionality and is at least needed to manage applet plugins. We'd have to look into providing that in blueman-manager or something to avoid losing it.

ton:
What is the status of this, is this supposed to work now?

I would like to run `blueman-applet` in the background only (without a systray icon), and use `blueman-manager` for interacting with the tool. 

I've tried several options to disable the systray icon, without success:

* Disabling the `AppIndicator` plugin. It removes the systray icon temporarily but it immediately shows up again. After restarting `blueman-applet`, the `AppIndicator` plugin is listed as disabled, but the systray icon is still there.
* Removing `!AppIndicator` or `AppIndicator` from the dconf key.

I'm using the latest version of blueman.

cschramm:
@ton: You can test #1496.

Issue #949, opened 2018-12-01
Using blueman-manager without blueman-applet


WhyNotHugo:
blueman: 2.0.6
BlueZ: 5.50
Distribution: ArchLinux
Desktop environment: 1.20.3

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Blueman works pretty well for me, but the usability is kinda confusing. If I start `blueman-manager` (which is the actual tool I need), it doesn't really work. For example, to connect home headsets, I need to:

* Run `blueman-applet` (why do I need to run this?).
  * Dismiss a warning that `blueman-applet` shows, see #948
* Run `blueman-manager`.
* Actually connect my headsets.
* After closing `blueman-manager`, `blueman-applet` is still there, but if I close it, everything continues working fine.

So I'm kinda wondering why manager depends on the applet, and if this can be avoided somehow. It kinda seems superfluous to manually have to start the applet for the manager works, when I don't really need to applet and it doesn't provide any use to me (note: I know it has uses, it's just that _I_ don't need it).

WhyNotHugo:
> `blueman-applet` is kind of a daemon that provides all the functionality we need on DBus. If you close it and continue using `blueman-manager`, `blueman-applet` will automatically get started when you commit any action.

I was using the latest stable (rather than master), and the "auto-starting" of `blueman-applet` is kinda a new thing in master, so I wasn't aware of that.  Thanks for pointing this out!

> Regarding the tray icon, we could provide a way to disable the StatusIcon plugin, however that'd be some work as we'd have to change the way other plugins that depend on it work. That's basically what you're asking for, right?

TBH, I'd rather not have to run the applet process entirely. This sounds kinda like running a deamon for a one-shot operation (connecting headphones, in my case). I'd rather open the blueman-manager, use it, close it, and not have anything linger.

However, if it's kinda necessary due to design limitation, I guess having the icon hide-able would be close enough for me. :slightly_smiling_face:

infirit:
> I was using the latest stable (rather than master), and the "auto-starting" of `blueman-applet` is kinda a new thing in master, so I wasn't aware of that. Thanks for pointing this out!

This is how its been in stable and its the core of blueman.

> TBH, I'd rather not have to run the applet process entirely. This sounds kinda like running a deamon for a one-shot operation (connecting headphones, in my case). I'd rather open the blueman-manager, use it, close it, and not have anything linger.

[...]

PR #1496, opened 2021-02-15
Make StatusIcon plugin unloadable

cschramm:
This provides a solution for #1495 / (#949), however, if a user disables the plugin from the UI, there will be no obvious way for him to re-enable it again (GSettings works, of course).

### DIFF
--- a/blueman/gui/applet/PluginDialog.py
+++ b/blueman/gui/applet/PluginDialog.py
@@ -284,50 +284,36 @@ class PluginDialog(Gtk.Window):
 
         deps = self.applet.Plugins.get_dependencies()[name]
         loaded = self.applet.Plugins.get_loaded()
-        to_unload = []
-        for dep in deps:
-            if dep in loaded:
-                to_unload.append(dep)
+        to_unload = [dep for dep in deps if dep in loaded]
 
         if to_unload:
-            dialog = Gtk.MessageDialog(parent=self, type=Gtk.MessageType.QUESTION, buttons=Gtk.ButtonsType.YES_NO)
-            dialog.props.secondary_use_markup = True
-            dialog.props.icon_name = "blueman"
-            dialog.props.text = _("Dependency issue")
-            dialog.props.secondary_text = \
+            if not self._ask_unload(
                 _("Plugin <b>\"%(0)s\"</b> depends on <b>%(1)s</b>. Unloading <b>%(1)s</b> will also unload <b>"
                   "\"%(0)s\"</b>.\nProceed?") % {"0": ", ".join(to_unload), "1": name}
-
-            resp = dialog.run()
-            if resp != Gtk.ResponseType.YES:
-                dialog.destroy()
+            ):
                 return
+        else:
+            conflicts = self.applet.Plugins.get_conflicts()[name]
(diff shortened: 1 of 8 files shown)


infirit:
I'm not sure this is a good idea but I can't really say why. But as long as the default is to show the tray it's ok with me

blueman/gui/applet/PluginDialog.py
```diff
@@ -284,50 +284,36 @@ def on_toggled(self, _toggle: Gtk.CellRendererToggle, path: str) -> None:
...
         self.applet.Plugins.set_config(name, name not in self.applet.Plugins.get_loaded())
+
+    def _ask_unload(self, text: str) -> bool:
+        dialog = Gtk.MessageDialog(parent=self, type=Gtk.MessageType.QUESTION, buttons=Gtk.ButtonsType.YES_NO)
+        dialog.props.secondary_use_markup = True
+        dialog.props.icon_name = "blueman"
+        dialog.props.text = _("Dependency issue")
+        dialog.props.secondary_text = text
+
+        resp = dialog.run()
```
infirit, line 317: run is a blocking operation. Better to do something like bc25b6d3b74b5c6c7a2d54f731d855db27d4bca8

infirit (commented):
Outside the blocking dialog it just needs a rebate.

blueman/gui/applet/PluginDialog.py
cschramm: Not sure if I'm missing something, but blocking seems exactly right here as `on_toggled` needs the dialog's response to continue.

cschramm: Of course we could listen to `GtkDialog::response` instead and somehow pass the `to_unload` data to handle the rest of the `on_toggled` code there but what would be the benefit of that?

infirit: The problem is the whole applet is blocked while the dialog is open. So for example your won't be able to kill it (SIGINT) as that required the main thread to be responding.

cschramm: `killall blueman-applet` while this dialog is open works fine for me, both with this code and in `main` (I did not really change or introduce anything here, just extract and deduplicate the code into a new method).

infirit: Apperantly I am confused :confused: . Not sure why but even with the blocking call the applet keeps responding.

infirit (approved):

blueman/gui/applet/PluginDialog.py
cschramm: We're inside an event handler at this point. I guess blocking one event handler does not block the event loop, at least in some multi-threaded context, right? :shrug:

WhyNotHugo:
> there will be no obvious way for him to re-enable it again (GSettings works, of course).

Just open View > Plug-ins and check it again?

cschramm:
Lol, you're right, blueman-manager provides a way. :sweat_smile:

cschramm:
Weird. I remember having seen this working when I hacked it together, but now it does not as `Menu` is flagged as not unloadable and depends on `StatusIcon`. :thinking:

Issue #1517, opened 2021-04-02
Feature request


jbierkens:
If I right click in the bluetooth manager tray icon, I get a useful menu. There is one feature I often use which isn't in the menu however: to disconnect currently connected bluetooth device. Is it possible to add this? (Also there are many items in the menu that I never use, in fact I only use the "Recent connections" and "Devices..." menu items, to connect and disconnect.)

blueman: 2.12
BlueZ:
Distribution: Xubuntu 20.04.2
Desktop environment: XFCE

cschramm:
Duplicate of #1268, added in #1427.

PR #1519, opened 2021-04-17
Define NetConf attributes in __init__

cschramm:
This NetConf stuff is a mess... I somehow ended up without those attributes in /var/lib/blueman/network.state which can easily happen if store gets called before any call to set_ipv4 and the resulting object is completely unusable (get_default already fails, set_ipv4 would fail, so it can never get a configuration, and so so...)


cschramm:
The better alternative to this would be #1517 of course.

infirit (approved):

PR #1522, opened 2021-04-17
Use alpine for proper Python 3.10 builds

cschramm:


cschramm:
The `test` job fails as Alpine does not have `/usr/lib/girepository-1.0/AppIndicator3-0.1.typelib`. It should be sufficient to just copy it in but I don't know a proper source.

knuxify:
https://gitlab.alpinelinux.org/alpine/aports/-/commit/41548c9e64666123f269378272f08722edb48486 adds the missing file. If the test runners run on Alpine edge, this should work now. If not, we'll need to wait for 3.15 to come out.

cschramm:
@knuxify: Thanks for the hint but the python alpine containers do not use edge, so we still have to wait to get that beauty.

cschramm:
The Debian-based images switched to bullseye now so that automake is sufficient and mypy works. However, bullseye replaced libappindicator with libayatana-indicators so that the tests fail, see #1611. That currently still leaves us with image options with sufficient automake versions (Alpine 3.13, 3.14 and Debian bullseye) disjunctive from images with sufficient libappindicator versions (Debian buster). Alpine 3.15 should finally have both. (An alternative would be to add upstream support for libayatana-indicators to make things work on bullseye but I'm not planning to do so.)

cschramm:
Solved with #1631 after #1382 dropped the libappindicator dependency.

PR #1546, opened 2021-06-01
Fix hard dependency of DBusService on NetworkManager

cschramm:


mgorny:
Thanks! I can confirm that with this patch blueman works again ;-).

Issue #1544, opened 2021-05-31
Blueman provides incorrect DNS info to NAP clients when using dhcpd for DHCP on Ubuntu


justdave:
blueman: 2.1.3-2ubuntu1.1
BlueZ: 5.55-0ubuntu1.1
Distribution: Ubuntu 20.10/groovy
Desktop environment: whatever the default is on Ubuntu - I can't find an About box anywhere to tell me what it is

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

I originally filed this at https://bugs.launchpad.net/ubuntu/+source/blueman/+bug/1845824 almost two years ago, but they've never patched it and never saw fit to upstream the bug report apparently.

Symptom: When configuring blueman to offer a Bluetooth Network Access Point and instructing it to use dhcpd for the DHCP server, it feeds 127.0.0.53 to the clients as the DNS server. This does not work because 127.0.0.0/8 is only accessible from localhost, and the clients route that address to themselves.

The code which generates the DHCP config is in /usr/lib/python3/dist-packages/blueman/main/NetConf.py and the offending code can be found in the get_dns_servers function at line 31 (as of the version listed above)

It pulls the DNS server out of /etc/resolv.conf.  On modern Ubuntu systems, Ubuntu has a DNS cache server installed, which binds to 127.0.0.53, so that's what it puts in /etc/resolv.conf. This is not the real DNS server of course, and it's not reachable from NAP clients.

Two options seem to be available to fix this:

1) Check if NetworkManager is installed, and if it is, use it to get the DNS server instead of looking in resolv.conf
-or-
2) Add a configuration option for the DNS server to feed to clients.

The following bash snippet will grab the default route's primary DNS server from NetworkManager:
`systemd-resolve --status $( netstat -r | grep default | head -1 | awk '{ print $8 }' ) | grep 'DNS Server:' | sed -e 's/.*DNS Server: \(\S\+\).*/\1/'`

I suspect there's a better API for it from python, but I haven't researched it.

cschramm:
That's not NetworkManager. NetworkManager creates a perfectly fine /etc/resolv.conf for me. 127.0.0.53 is `systemd-resolved`. It symlinks `/lib/systemd/resolv.conf` with that static configuration pointing at its internal resolver.

Could you check #1553?

justdave:
[...]

PR #1553, opened 2021-06-04
Try to get DNS servers from systemd-resolved

cschramm:


cschramm:
This uncovers a conceptional flaw in that configuration system: It only reads the nameservers the moment when it configures the DHCP service and does not listen for changes. If e.g. nameservers get configured dynamically and there is no upstream connection yet, systemd-resolved shows an empty array and the logic writes a broken configuration file, see https://github.com/blueman-project/blueman/issues/1544#issuecomment-867007395.

cschramm:
Now based on and includes #1564.

infirit:
I thought about this a while back. There not reason to keep a dhcphandler running all the time. Blueman could just start it up when the connection is being made.

But yes, changing resolvers something Blueman currently can't handle..

blueman/main/DNSServerProvider.py
```diff
@@ -1,19 +1,79 @@
...
+        self._subscribe_systemd_resolved()
+        self._subscribe_resolver()
 
     @classmethod
     def get_servers(cls) -> List[IPv4Address]:
-        return list(cls._get_servers_from_resolver())
+        return list(set(cls._get_servers_from_systemd_resolved()) or cls._get_servers_from_resolver())
+
+    @staticmethod
+    def _get_servers_from_systemd_resolved() -> Generator[IPv4Address, None, None]:
```
infirit, line 26: Wouldn't a Gio.DBusProxy on `org.freedesktop.resolve1.Manager` make things a lot easier here?

cschramm: Would it? Sugar for `org.freedesktop.DBus.Properties.Get` is only in `blueman.bluez.Base.Base` and all three calls in this method have distinct (object path, interface)-tuples, so that they'd need different proxy objects.

infirit: The call method on the dbusproxy takes an optional interface name,  like `org.freedesktop.resolve1.Manager.GetLink`. as long as you are dealing with the same dbus object/name you should be able to access any interface like this.

infirit: Example to get the DNS property but this works on anything on `/org/freedesktop/resolve1`

```python
from gi.repository import GLib, Gio
proxy = Gio.DBusProxy.new_for_bus_sync(
    Gio.BusType.SYSTEM,
    Gio.DBusProxyFlags.DO_NOT_AUTO_START,
    None,
    "org.freedesktop.resolve1",
    "/org/freedesktop/resolve1",
    "org.freedesktop.resolve1"
    )

proxy.call_sync("org.freedesktop.DBus.Properties.Get", GLib.Variant('(ss)', ("org.freedesktop.resolve1.Manager", "DNS")), Gio.DBusCallFlags.NONE, -1, None)
```
which gives 
```python
GLib.Variant('(v)', (<[(3, 2, [byte 0xc0, 0xa8, 0x01, 0x01])]>,))
```

infirit: Ok, here is why I suggest using proxies.

```python
from gi.repository import GLib, Gio

def on_properties_changed(proxy, changed, invalidated):
    changed_properties = changed.unpack()
    if "DNS" in changed_properties:
        print(changed["DNS"])

manager_proxy = Gio.DBusProxy.new_for_bus_sync(
    Gio.BusType.SYSTEM,
    Gio.DBusProxyFlags.DO_NOT_AUTO_START,
    None,
    "org.freedesktop.resolve1",
    "/org/freedesktop/resolve1",
    "org.freedesktop.resolve1.Manager"
    )


manager_proxy.connect("g-properties-changed", on_properties_changed)
dns = manager_proxy.get_cached_property("DNS")
link_path = manager_proxy.GetLink("(i)", 2)

link_proxy = Gio.DBusProxy.new_for_bus_sync(
    Gio.BusType.SYSTEM,
    Gio.DBusProxyFlags.DO_NOT_AUTO_START,
    None,
    "org.freedesktop.resolve1",
    link_path,
    "org.freedesktop.resolve1.Link"
    )

default = link_proxy.get_cached_property("DefaultRoute")
```

cschramm: I think I did not get that last comment and I'm unsure where we're at with this PR. :sweat_smile: Do you suggest additional changes?

infirit (approved):
Let's merge and see what comes up.

PR #1554, opened 2021-06-07
Spelling: Various device classes

comradekingu:


blueman/DeviceClass.py
```diff
@@ -97,9 +97,9 @@
     # translators: device class
     _("Set-top box"),
     # translators: device class
-    _("Hifi audio"),
+    _("Hi-Fi audio"),
```
cschramm, line 100: https://btprodspecificationrefs.blob.core.windows.net/assigned-numbers/Assigned%20Number%20Types/Baseband.pdf says `HiFi`. I'd stick to that.

blueman/DeviceClass.py
```diff
@@ -53,24 +53,24 @@
...
     # translators: device class
-    _("Isdn")
+    _("ISDN")
 )
 
 access_point_minor_cls = (
     # translators: device class
     _("Fully"),
     # translators: device class
-    _("1-17 percent"),
```
cschramm, line 63: These values are probably not very good in general as they lack context. We should probably adapt the ones from https://btprodspecificationrefs.blob.core.windows.net/assigned-numbers/Assigned%20Number%20Types/Baseband.pdf, i.e. include "utilized" and "available".

blueman/DeviceClass.py
```diff
@@ -180,8 +180,8 @@
     11: ('Barcode Scanner', {0: _("Generic Barcode Scanner")}),
     12: ('Thermometer', {0: _("Generic Thermometer"),
                          1: _("Thermometer: Ear")}),
-    13: ('Heart rate Sensor', {0: _("Generic Heart rate Sensor"),
-                               1: _("Heart Rate Sensor: Heart Rate Belt")}),
+    13: ('Heart rate Sensor', {0: _("Generic Heart-rate Sensor"),
```
cschramm, line 183: https://specificationrefs.bluetooth.com/assigned-values/Appearance%20Values.pdf says `Heart Rate`. I'd stick to that.

comradekingu: True, but when there is a third word, is it "a heart-rate sensor", or a "heart rate-sensor". That is also why I added one for the lone "heart-rate", since that is consistent, and I don't think illegal.

blueman/DeviceClass.py
comradekingu: Pedantic person on the Internet says:
"HiFi" is the worst of all possible options. It is neither "Hi-fi" or "Hifi", but it is also the worst option of the two.
Moreover if you wanted to use big F, use "Hi-Fi", from DIN 45500.
http://brakken.no/el2/avgrensa/din45500.pdf

I am also an audio guy, and "HiFi" just looks, wrong.

blueman/DeviceClass.py
comradekingu: I don't understand what is being utilized or alternatively remains available with the context. It says "Minor Device Class field – LAN/Network Access Point Major Class"

Is it available clients, signal strength, battery level, bandwidth, … ?

blueman/DeviceClass.py
cschramm: The spec says `Heart Rate Sensor`. Not saying that we absolutely need to follow it, of course, it just seems reasonable to me.

blueman/DeviceClass.py
cschramm: I don't know either. Probably clients or upstream bandwidth. :shrug:

PR #1562, opened 2021-07-01
Fix read only filesystem error.

paledega:
Screenshot_2021-07-01_10-13-47

We must move /var/lib/blueman to /var/run/blueman

if We mount root as ro, We cannot write into /var/lib/blueman but we can always write /var/run/blueman because /var/run will mount as tmpfs.

### DIFF
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -309,7 +309,7 @@ class NetConf:
             return NetConf.default_inst
 
         try:
-            with open("/var/lib/blueman/network.state", "rb") as f:
+            with open("/var/run/blueman/network.state", "rb") as f:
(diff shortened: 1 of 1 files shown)


cschramm:
I'm far from an FHS expert but that does not seem plausible to me.

First, `/var` is the goto hierarchy for variable data. "variable" as in "written to during system operation (as opposed to installation and software maintenance)". Its purpose is actually to keep variable data out of `/usr` so that `/usr` can be read-only. `/var` thus must not be read-only.

`/var/lib` is for "variable state information", a perfect fit for our `network.state`.

`/var/run` is just for compatibility to older versions of FHS and was intended for run-time variable data, now served by `/run`.

Contents of `/run` must be cleared on boot as they are considered transient. `network.state` is not transient.

PR #1563, opened 2021-07-05
Fix power level bars

cschramm:
39866e059 mixed things up so th39866e059 mixed things up so that the tpl bar is too short and the lq value may exceed 100 percent leading to an exception as there is no icon for that.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -418,13 +418,13 @@ class ManagerDeviceList(DeviceList):
             except ConnInfoReadError:
                 bars["rssi"] = 50
             try:
-                bars["lq"] = max(50 + float(cinfo.get_lq()) / 127 * 50, 10)
+                bars["lq"] = max(float(cinfo.get_lq()) / 255 * 100, 10)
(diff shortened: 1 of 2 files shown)


blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -418,11 +418,11 @@ def _update_power_levels(self, tree_iter: Gtk.TreeIter, device: Device, cinfo: c
             except ConnInfoReadError:
                 bars["rssi"] = 50
             try:
-                bars["lq"] = max(50 + float(cinfo.get_lq()) / 127 * 50, 10)
+                bars["lq"] = max(float(cinfo.get_lq()) / 255 * 100, 10)
             except ConnInfoReadError:
                 bars["lq"] = 50
```
collares, line 423: Should this be changed to 0 (and the corresponding line in the `tpl` except to `0`)?

cschramm: Good catch. Mostly right, except that 0 is invalid in any case and it needs to be 10. :see_no_evil:

cschramm:
Included in 2.2.2 and #1583

PR #1572, opened 2021-07-23
Fix error message in blueman-mechanism

cschramm:
Gtk.IconTheme.get_default(), triggered by some imports, cannot work in a typical environment that blueman-mechanism is run in and emits an error message.


cschramm:
Included in 2.2.2 and #1583

Issue #1461, opened 2020-12-06
blueman-applet logs exception in GameControllerWakelock on wayland


jokeyrhyme:
blueman: 2.1.4
BlueZ: 5.55
Distribution: Archlinux
Desktop environment: swaywm (wayland, no X11)

* [X] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Howdie, I hope you are well, thanks for this extremely useful project :+1:

I've been noticing some python exceptions in my logs, and finally tracked them down to `blueman-applet`: 

```
blueman-applet version 2.1.4 starting
Stale PID, overwriting
blueman-applet 07.51.52 ERROR    PluginManager:91 load_plugin: Unable to load plugin module GameControllerWakelock
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/blueman/main/PluginManager.py", line 89, in load_plugin
    importlib.import_module(self.module_path.__name__ + ".%s" % plugin)
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 790, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/lib/python3.9/site-packages/blueman/plugins/applet/GameControllerWakelock.py", line 14, in <module>
    raise ImportError('This is not an X11 screen')
ImportError: This is not an X11 screen
blueman-applet 07.51.52 WARNING  PluginManager:148 __load_plugin: Not loading DhcpClient because its conflict has higher priority
blueman-tray version 2.1.4 starting
blueman-applet 07.51.52 WARNING  PluginManager:148 __load_plugin: Not loading PPPSupport because its conflict has higher priority
```

This is not at all urgent for me, as blueman is doing it's job very well for my use case

[...]

PR #1601, opened 2021-09-22
Be quieter about plugin failures on Wayland

RPigott:
I get the errors from #1461 a lot so I thought I'd try to quiet them a little bit.

There is no sensible way to implement this feature on Wayland atm, so the behavior stays the same. The only difference is the logging, which will change this

```
blueman-applet 13.56.12 ERROR    PluginManager:90 load_plugin: Unable to load plugin module GameControllerWakelock
  File "/usr/lib/python3.9/site-packages/blueman/main/PluginManager.py", line 88, in load_plugin
    importlib.import_module(self.module_path.__name__ + f".{plugin}")
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/lib/python3.9/site-packages/blueman/plugins/applet/GameControllerWakelock.py", line 15, in <module>
    raise ImportError('This is not an X11 screen')
```

to this:

```
blueman-applet 16.40.30 WARNING  PluginManager:92 load_plugin: Failed to start plugin GameControllerWakelock: Only X11 platform is supported
```

### DIFF
--- a/blueman/main/PluginManager.py
+++ b/blueman/main/PluginManager.py
@@ -13,6 +13,7 @@ from blueman.gui.CommonUi import ErrorDialog
 from blueman.main.Config import Config
 from blueman.plugins.BasePlugin import BasePlugin
 from blueman.bluemantyping import GSignals
+from blueman.plugins.errors import PluginException
 
 
 class LoadException(Exception):
@@ -88,6 +89,8 @@ class PluginManager(GObject.GObject):
                 importlib.import_module(self.module_path.__name__ + f".{plugin}")
             except ImportError:
                 logging.error(f"Unable to load plugin module {plugin}", exc_info=True)
+            except PluginException as err:
+                logging.warning(f"Failed to start plugin {plugin}: {err}")
 
         for cls in self.plugin_class.__subclasses__():
             self.__classes[cls.__name__] = cls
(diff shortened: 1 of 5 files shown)


cschramm:
Makes perfect sense. :+1: Something similar might make sense for other `ImportError`s that we raise explicitly as well.

Can you please adapt the line breaks to make flake8 pass? The test and mypy problems seem to be due to some issue with GitHub's `ubuntu-latest` environment which currently actually points to a Debian system. Not sure what is going there... :thinking:

cschramm:
Strike that. They are related to the Python Docker images defaulting to bullseye now. Fixed in #1611 (approval very welcome). For the potfile check you'll need to update the pot file using `make -C po blueman.pot-update` and add the changes to the commit. With the 3.10-rc tests are expected to fail and thus not required.

cschramm:
#1612 adds a logic to automagically update blueman.pot afterward, so that we do not have to care here and elsewhere anymore. Approval welcome. :grin:

RPigott:
Thanks for the direction. I'm not too familiar with gettext/i18n so I hope that worked.

cschramm (approved):

PR #1614, opened 2021-10-06
Drop PIN database

### LINKED ISSUE #1475: Broken PIN lookup
blueman: 2.1.4
BlueZ: 5.55
Distribution: LFS
Desktop environment: XFCE

Pairing of audio device (Teufel Bluetooth Headset) fails 
Blueman tries to read the pin from pin-code-database, there it returns the pin value for the one audio device defined there (max:4) - based on the value it creates a random value and uses this to pair with the headset.
But the Teufel Bluetooth Headset expects 0000 as pin.
I guess the code in BluezAgent.py using PIN_SEARCHES is not that ideal - because it gets down to just look for same type and there it gets the wrong value in this case
How is this supposed to work correctly? 
If i add the value for the new device to the pin-code-database its working as expected - but i would either expect a hint to the user to add the device to the database or some kind of fallback mechanism i.e. if code is not found show the Pin-Request dialog. In this case here the dialog is not shown because max:4 is returned as pin due to the one existing audio device

* [ x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

<!--

ℹ Please use some service like [Pastebin](http://pastebin.com/) or [GitHub Gist](https://gist.github.com/) to post your logs and keep the thread clean and readable. Make sure to describe the exact steps you took when producing it.

->

cschramm:
This clearly seems to do more harm than good:

* The special values NULL, KEYBOARD, and ICADE do not actually get handled but provided as strings.
* The lookups are too unspecific, so that any devices that are classified as mouse, audio, or network matches an entry typically not made for it.
* On a match, the user cannot enter a PIN even if the automatic one turns out to be wrong.

On the other hand it does not seem very relevant:

* The whole mechanism is only for Legacy Pairing (Bluetooth <= 2.0) with devices that use a fixed or random PIN.
* Providing 0000 (or any 4-digit PIN) for keyboard, printer, joypad and other devices is handled by the autopair plugin of BlueZ 5.51 and newer.
* Most entries with names refer to very specific devices, typically with an uncategorized major device class, that caused issues in gnome-bluetooth over a decade ago as it was not possible to enter the correct PIN for them. It is possible in blueman and if anybody even still uses such a device, it should be fine to ask for the PIN.

Closes #1475


infirit:
If it causes problems then yeah, drop it.

infirit (approved):

PR #1625, opened 2021-10-28
Create release drafts on git tag pushes

cschramm:


infirit (approved):
Interesting, not tested but merge it and see.

Issue #1620, opened 2021-10-23
UI: Show legends for those progress bars on the right side.


WhyNotHugo:
blueman: 2.2.2
BlueZ: n/a
Distribution: ArchLinux
Desktop environment: swaywm (wayland)

This is what my main blueman windows looks like:

![2021-10-23T21:58:20,653283504+02:00](https://user-images.githubusercontent.com/730811/138569874-0b3a9624-ce1e-4ab0-9076-121a6736ab6b.png)

There's three bars on the right side, but I never remember what they are. At the same time, there's a huge amount of empty/unused space.

I'd like to propose rotating those progress bars (so they're horizontal), and adding a readable label to each one. Actually, the entire contents of the tooltip could be included there. There's two issues going about here: lots of unused space, and information too condensed and hard to read. I think they can both cancel each other out.

This is mostly a UI/UX improvement, but also allows keyboard-only users to figure out those bars -- currently only hovering them gives an indicator of what they are.

It doesn't necessarily have to be progress bars either, small tags with percent indicators might work too (keeping colour coding as well):

![2021-10-23T22:10:07,941533497+02:00](https://user-images.githubusercontent.com/730811/138570107-d3fb22cc-4ab3-47bd-9aa9-6b19c4bed2cf.png)

I'm not entirely sure what layout is best though; if they have "reserved" positions (so they layout matches on all devices), or only render ones with data (devices with different bars will have them in different orders, but they flow would be more natural).

What are your thoughts? Would you rather see a PR and start of discussion there?

cschramm:
First thoughts first: Empty / unused space is not an argument as it's just due to the window size. :smile: 

I absolutely do not like the current state of the indicators both visually and functional and personally find your tags quite appealing.

Regarding functional: In many cases we only show more or less "default" and thus pointless values now. I just created #1642 to make some changes visible: The pointless link quality value will go away completely and the others will probably be absent in a lot of cases as well. I'd still keep the values for cases where they make sense. For other cases we'll probably need another connection indicator.

If I remember correctly (it aged well :grimacing:), the code in #1642 will also drop the weird spacing where the battery indicator of your MX Vertical floats in open space by putting all bars into a single tree cell. I'd definitely go without "reserved" positions.

A PR is always welcome. Feel free to pick changes from #1642 as they fit you.

Also curious about @infirit's thoughts on this.

varlesh:
My concept. Idea this style - move all status info too right with symbolic icons.
[...]

PR #1633, opened 2021-11-14
Symbolic icons

varlesh:
added symbolic icons for GUI in Adwaita style:
```
blueman-active
* blueman-tray

blueman-disabled
* blueman-tray-disabled

blueman-tray
* blueman-tray

blueman-device
* bluetooth-devices-symbolic

blueman-down-active
* rssi-active-symbolic

blueman-down-inactive
* rssi-inactive-symbolic

blueman-plugin
* plugin-symbolic

blueman-serial
* device-serial-symbolic

blueman-trust
* trust-symbolic

blueman-untrust
* untrust-symbolic

blueman-up-active
* tpl-active-symbolic

blueman-up-inactive
* tpl-inactive-symbolic

image-missing
* image-missing-symbolic

blueman-lq-10-100
* lq-10-symbolic
* lq-20-symbolic
etc

blueman-tpl-10-100
* tpl-10-symbolic
* tpl-20-symbolic
etc

blueman-rssi-10-100
* rssi-10-symbolic
* rssi-20-symbolic
etc
```
Also other icon names need rename with `-symbolic` suffix

```
application-exit
rename to: application-exit-symbolic

audio-card
rename to: audio-card-symbolic

battery
rename to: battery-symbolic

blueman-send-file
rename to: document-send-symbolic

blueman-x
rename to: window-close-symbolic

camera-photo
rename to: camera-photo-symbolic

camera-video
rename to: camera-video-symbolic

computer
rename to: computer-symbolic

dialog-error
rename to: dialog-error-symbolic

dialog-information
rename to: dialog-information-symbolic

dialog-password
rename to: dialog-password-symbolic

dialog-warning
rename to: dialog-warning-symbolic

document-new
rename to: document-new-symbolic

document-open
rename to: document-open-symbolic

document-open-recent
rename to: document-open-recent-symbolic

document-properties
rename to: document-properties-symbolic

edit-clear
rename to: edit-clear-symbolic

edit-copy
rename to: edit-copy-symbolic

edit-delete
rename to: edit-delete-symbolic

edit-find
rename to: edit-find-symbolic

gnome-power-manager
rename to: gnome-power-manager-symbolic

go-down
rename to: go-down-symbolic

go-up
rename to: go-up-symbolic

gtk-preferences
rename to: preferences-system-symbolic

help-about
rename to: help-about-symbolic

info
rename to: user-info-symbolic

input-gaming
rename to: input-gaming-symbolic

input-keyboard
rename to: input-keyboard-symbolic

input-mouse
rename to: input-mouse-symbolic

input-tablet
rename to: input-tablet-symbolic

list-remove
rename to: list-remove-symbolic

menu-editor
rename to: open-menu-symbolic

modem
rename to: modem-symbolic

network
rename to: network-transmit-receive-symbolic

network-offline
rename to: network-offline-symbolic

network-transmit-receive
rename to: network-transmit-receive-symbolic

network-wireless
rename to: network-wireless-symbolic

network-workgroup
rename to: network-workgroup-symbolic

phone
rename to: phone-symbolic

preferences-desktop
rename to: preferences-system-symbolic

preferences-system
rename to: preferences-system-symbolic

printer
rename to: printer-symbolic

process-stop
rename to: process-stop-symbolic

system-shutdown
rename to: system-shutdown-symbolic

view-refresh
rename to: view-refresh-symbolic

blueman-battery-10-100
rename to: battery-level-10-charging-symbolic
rename to: battery-level-20-charging-symbolic
etc
```


varlesh:
preview
image
not shure about tray icons, need this added -symbolic suffix or not... also i'm not change code on application. It's only SVG's checked with Adwaita icon names.

cschramm:
Hi @varlesh, to be honest, I'm not sure what to do with these files. I can more or less map them to icons that we ship with blueman but symbolic versions are typically provided by symbolic icon themes. I get the point of adding some to blueman instead of adding blueman-specific icons to any symbolic icon theme out there but is there even a way to ship two sets side by side and pick one or the other depending on whether or not the system theme is symbolic?

varlesh:
Mostly GNOME/GTK-apps used symbolic icons. It looks modern and does not stand out from the style. Also symbolic icons can be colorize from GTK-theme and this good solutions for darker themes (dark panel, light menu and window).
About themes - icons can be installed to hicolor scalable directory, it's already doing on this PR.
This PR - only monochrome icons. Your right to change the icon style of the application or not...

jfly:
@cschramm, you don't actually have to do anything with these files other than ship them alongside the non-symbolic icons. GTK actually has logic to conditionally look up `*-symbolic` versions of icons. See https://gitlab.gnome.org/GNOME/gtk/-/blob/b2c227e9c57839a2a4e24462a71ae0bad9a95264/gtk/gtkicontheme.c#L2376. One of the ways to trigger this behavior is to set the `-gtk-icon-style` css property mentioned here: https://mail.gnome.org/archives/gtk-devel-list/2014-May/msg00020.html.

This behavior is different and arguably better than the "symbolic icon theme"s you mentioned, where people provide simple, flat, monochrome icons with names like `bluetooth-adapter.svg`. In that scenario, gtk doesn't realize it's drawing a symbolic icon, whereas when it renders an icon that it knows is symbolic (the name ends in `-symbolic`), then it can do clever stuff like recoloring the icon according to the `_NET_SYSTEM_TRAY_COLORS` hint.

For more information:

- See this PR to polybar to add proper support for the `_NET_SYSTEM_TRAY_COLORS` hint: https://github.com/polybar/polybar/pull/2552
- This change to my dotfiles: https://github.com/jfly/dotfiles/commit/f6043a72081f639ff8ec318bfa8b68d3932d8dc0, where I set the `-gtk-icon-style: symbolic;` css attribute and added `*-symbolic` versions of many of the blueman icons. If this PR gets merged up, then I could remove those extra `-symbolic.svg` files.

cschramm:
@varlesh: Actually this PR is not installing anything as the file names get added neither to Makefiles nor to the meson build file but I can take care of that once I figured out how this can work overall.

@jfly: Thanks a lot for the explanation. So there is a way to ship both, assuming that desktops somehow set `-gtk-icon-style: symbolic;` if desired. That means to me, for every foo-symbolic.svg there should be a flanking foo.svg.

That does not seem to be the case at all, so let's get into the set then:
* [ ] I guess blueman-tray2 should actually be blueman-tray-symbolic.
* [ ] blueman-tray-disabled should be blueman-disabled-symbolic.
* [ ] I cannot make sense of bluetooth-adapter.
* [ ] bluetooth-devices should probably be blueman-device-symbolic.
* [ ] device-serial-symbolic should probably be blueman-serial-symbolic.
* [ ] image-missing is a standard icon from the [Icon Naming Specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html). We do not ship it.
* [ ] The lq, rssi, and tpl icons will not work as we load pixmaps there, not icons. That's an open UI topic and we might end up with no icons at all. You might want to join #1620.
* [ ] I guess rssi-active-symbolic, rssi-inactive-symbolic, tpl-active-symbolic, and tpl-inactive-symbolic should be blueman-down-active-symbolic, blueman-down-inactive-symbolic, blueman-up-active-symbolic, and blueman-up-inactive-symbolic.
* [ ] plugin-symbolic, trust-symbolic and untrust-symbolic lack a blueman- prefix.
* [ ] This should cover most of our icons but at least a blueman-symbolic (can be just a copy of blueman-tray-symbolic) and blueman-send-file-symbolic seems to be missing.

infirit:
Just so I'm understanding the symbolic stuff properly. What makes GTK use the symbolic icons? Is the the (icon)theme?

cschramm:
An override like [@jfly's](https://github.com/jfly/dotfiles/blob/f6043a72081f639ff8ec318bfa8b68d3932d8dc0/homies/config/gtk-3.0/gtk.css). No clue if desktops actually set that, though. :shrug:

infirit:
Ah, so it's theme based. Then if we can cover all icons this should be a nice addition.

infirit:
On top of what @cschramm listed, I updated your list :-)

* We use default icons for trust/untrust. If there was a full colour one I would be happy to use that instead.
* bluetooth-device-symbolic what is this supposed to be used for?

I moved and renamed the icons so they in the proper location and name in #1716.

cschramm:
(@clefebvre and me already discussed this on IRC I think in February)

Personally I'm fine with fully adopting symbolic icons within the UI. I see the downsides of maintaining two variants and I'm not attached to the current fullcolor look. My main (only?) exception is the application and tray icon which I'd like to keep in fullcolor. I think we discussed that and had the same idea there but I do not remember the exact solution that you proposed (desktops still want to have at least the tray icon in mono or even symbolic).

@infirit, what do you think?

@infirit added the missing bits to this PR in #1716. If I'm not mistaken that structure goes straight into the `-gtk-icon-style` direction, though, so if we decide against it, it does not make much sense to merge it, right?

clefebvre:
Yes, I think for the status icons within the app (connectivity) the colors really add something. The only thing we need to fix there is HiDPI support. The idea proposed by @whynothugo was also pretty cool:

image

He retains the colors while making the status more easy to understand and avoids HiDPI issues.

Regarding the tray I think the best option is to make it a setting in gsettings. This lets users decide whether they want a fullcolor BT icon or a symbolic one, and more importantly it lets distros set their preferred default for what integrates best with their desktop. There's definitely merit in both the traditional immediately identifiable blue BT icon and the symbolic one which mingles well with other status icons (battery, network etc...).

PR #1649, opened 2021-12-14
Free mypy

infirit:
Tested with version 0.910

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: apt-get update
       - run: apt-get install -y -qq --no-install-recommends automake autoconf libtool autopoint gettext libglib2.0-dev python-gi-dev libbluetooth-dev iproute2
-      - run: python3 -m pip install cython mypy==0.812 pycairo>=1.16.3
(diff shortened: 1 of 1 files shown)


infirit:
This takes care of mypy on python 3.11

cschramm:
I'm eagerly monitoring https://github.com/python/mypy/issues/11158 as I would have bumped to 0.920 once it's available, but freeing it is fine for me as well. I'm actually surprised that 0.910 works with Python 3.11 as I expected https://github.com/python/mypy/pull/11531 / https://github.com/python/mypy/pull/11560 to be required (to get typed_ast 1.5.1 with https://github.com/python/typed_ast/pull/176) but that's not part of a release yet. :thinking:

cschramm (approved):

infirit:
If it starts causing issues again I'll pin it again to a version. Happy to see those failed checks gone 😁

hauntsaninja:
mypy only requires typed-ast if you're running with Python 3.7 or lower, or if you want to check Python 2 code :-)

cschramm:
Right. https://github.com/python/mypy/commit/4adb5e4631916ea0dee625cdfaf18ac7ef4a1c58 is the crucial change. The compilation error actually only affects 0.812 while 0.900 and newer just do not even try to install typed-ast on Python 3.11. :+1:

PR #1652, opened 2021-12-19
Put recent connections in toplevel applet menu

### LINKED ISSUE #1558: UI improvements suggestions
Here are a few UI suggestions which will greatly improve the usability of the app.

1) Add a connect button next to each item in devices so the user does not need to right click and open a context menu. This reduces 1 click and is simpler

2) Replace "search" button with "refresh" or "Search/Refresh". Its not clear that search is searching for devices and not searching within the already existing devices. 

3) Show the device list in the drop down menu on the toolbar. Why make the users open another window when they want to connect? Its really annoying

4) Don't show "Bluetooth devices ready" popup. Only show an error if they are not ready.

Thanks

cschramm:
Other applets like NetworkManager's do not try to keep menu small, so why should blueman...

Closes #1558

This was lying around for a while now. I'm not happy with the presentation yet and don't have any better idea for it but in general this should be good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Drop auto-power feature. BlueZ now has the AutoEnable setting for even better auto-powering.
 * Do not use pointless link quality value
 * Do not re-use dbusmenu item identifiers; avoids issues at least with gnome-shell-extension-appindicator.
+* Recent connections in toplevel applet menu
 
 ### Bugs fixed
 
--- a/blueman/main/indicators/StatusNotifierItem.py
+++ b/blueman/main/indicators/StatusNotifierItem.py
@@ -43,16 +43,17 @@ class MenuService(DbusService):
     def _get_layout(self, parent_id: int, _recursion_depth: int, _property_names: List[str]
                     ) -> Tuple[int, Tuple[int, Dict[str, GLib.Variant], List[GLib.Variant]]]:
         if parent_id == 0:
-            return self._revision, (0, {}, self._render_menu(self._items.items(), self._render_submenu))
+            return self._revision, (0, {}, self._render_menu(((item["id"] << 8, item) for item in self._items.values()),
+                                                             self._render_submenu))
         else:
-            item = self._items[parent_id]
+            item = self._items[parent_id >> 8]
             if "submenu" in item and _recursion_depth != 0:
                 return self._revision, (parent_id, self._render_item(item), self._render_submenu(item, parent_id))
             return self._revision, (parent_id, self._render_item(item), [])
 
     def _render_submenu(self, item: "MenuItemDict", idx: int) -> List[GLib.Variant]:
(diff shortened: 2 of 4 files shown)


infirit:
I tried it and outside of duplicated entires for devices and it being mixed in with the recent connections I worry about the exit menu running off the screen. How are we going to make sure when someone has 24 set as the max recent connections exit is still accessible?

I was thinking of adding an expander when we reach, let's say 6 recent, and hide older ones. This however is going to complicate the current design so yeah.. not sure what's worse.

infirit:
I'll test with 24 and see how the menu behaves.

cschramm:
When I considered #1558 I thought about an adapting expander as well, but came to the conclusion that just adding everything to the top level should be fine. Some arguments for "don't worry about large menus":
* The default maximum for recent connections is 6.
* If I remember correctly, back when I wrote the code I looked at nm-applet and found that they have no limit for their menu at all (or least something huge like 30 or something), so they do not seem to worry either.
* At least GTK-based menu implementations typically show a concept of scrolling, so if all goes wrong, all we get is a scrolling but still usable menu.

I can imagine that it might get rather nasty on low resolutions, though, but I guess at least as people do not increase the default maximum, they should be fine.

Duplicate devices? 🤔 Ah, you mean standard connect vs recent connection items?

cschramm:
Just checked a very low resolution and indeed multiple status icon menus - including our current one - started scrolling (mate-panel). Also, it looks like I did *not* remember correctly regarding nm-applet. It shows 5 networks and puts more into a submenu.

In any case, if we talk about adding 5 items on toplevel, plus a submenu item, so that it's up to 6 vs. just adding all items on toplevel that's the same number in 99.9 % of all cases as 6 is exactly our default maximum of recent connections and I don't think it's worth addressing the very special case of an increased maximum *and* a low resolution just to avoid scrolling in the menu if we're lucky.

infirit:
Thanks for checking.

This is what I am seeing. The duplicate Devices (Apparaten) was because I used the tray from master instead of this PR. But it still shows up in the middle of all the recently connected devices.
afbeelding

cschramm:
Ouch, yeah, stupid me. The position indexes start to overlap at some point. :facepalm:

infirit:
it's still duplicating entries with the latest changes.
image

Other than that it looks ok to me.

cschramm:
I cannot replicate any duplication. :thinking:

> The duplicate Devices (Apparaten) was because I used the tray from master instead of this PR.

again?

infirit:
> again?

Eeuh, wait, i get deja Vu, i may have done "it" again 😮‍💨

infirit:
Can you rebase this one last time :-). I'll make sure to test this weekend.

infirit (approved):
Made sure to properly install and restart :grin: . All looks good now.

PR #1653, opened 2021-12-21
Update de.po

odysseusPolytropos:
'Friendly Name' means a human-readable name in contrast to some incomprehensible ID. The current german translation 'Freundlicher Name' does hardly reflect this and is only a literal translation, therefore I propose 'Menschenlesbarer Name' as an alternative.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -85,7 +85,7 @@ msgstr "Vorübergehend sichtbar"
 
 #: data/ui/adapters-tab.ui:101
 msgid "<b>Friendly Name</b>"
-msgstr "<b>Freundlicher Name</b>"
+msgstr "<b>Menschenlesbarer Name</b>"
 
 #: data/ui/applet-passkey.ui:8
 msgid "Pairing request"


cschramm:
I fully agree but our authoritative source for translations is Weblate, see https://github.com/blueman-project/blueman#translate. Please contribute there or indicate that you do not want to (I'll do it then :smile:).

odysseusPolytropos:
Thanks for the hint. I'm not familiar with weblate, but added a proposal there.

For main:
https://hosted.weblate.org/translate/blueman/blueman/de/?checksum=7e551e7f3b43289e

For 2.2:
https://hosted.weblate.org/translate/blueman/2-2/de/?checksum=7e551e7f3b43289e

cschramm:
:smiley: No worries, you can actually just make edits without the proposal loop.

That will show up automagically in #1547 / #1626 at some automagic point in time and we'll merge it latest before a release. Thanks! :+1:

infirit:
I don't think human readable is the right choice. Unix time in second is
human readable but not friendly. The Bluetooth address is human readable
but not friendly.

It's one if those words that don't translate well to German or Dutch (my
native tongue). I don't have a better alternative in Dutch.

odysseusPolytropos:
Well, technically you are correct, that the Bluetooth addresses are human readable as binary code is also human readable, if you print it out. But generally I don't think this is the common usage of this term, because I would expect something attributed as 'human readable' to be intended to be read by humans in an easy way.

'Friendly name' is not the best choice in the first place (albeit established in English) and the German translation 'Freundlich' not very understandable imho, if you don't have a technical background. 'Benutzerfreundlicher Name' (=user friendly name) would be more precise, but unnecessary long.

Maybe 'Anzeigename' (=diplay name) would be a better translation. What do you think?

cschramm:
Makes sense but I wonder ... couldn't it just be Name? :thinking:

infirit:
+1 for just name. It requires code changes, if you're up to it @odysseusPolytropos let me know. Otherwise I'll handle it next week

PR #1675, opened 2022-02-10
Use a recent automake for release packages

cschramm:
Otherwise the configure script will not work for Python 3.10.

I already re-built and re-released https://github.com/blueman-project/blueman/releases/tag/2.2.4 with this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,8 @@ on:
 jobs:
   release:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


cschramm:
On second thought, we could also just use a bullseye container. Will give it a try.

cschramm:
There we go. That's much nicer. Successfully tested in https://github.com/blueman-project/blueman/actions/runs/1822828966.

infirit (approved):

PR #1693, opened 2022-03-21
Minor fix ru.po

goosesima:


cschramm:
Thanks for the contribution, @goosesima but the "authoritative" source for translations is Weblate. I applied the suggestion to https://hosted.weblate.org/translate/blueman/blueman/ru/?checksum=590b1a4a1b52a1b6 and the repository will get it automagically in #1547.

PR #1699, opened 2022-04-02
meson: add pythoninstalldir option

ChenQi1989:
In case of cross build, using host python to determine the python
site-packages directory for target is not feasible, add a new option
pythoninstalldir to fix the issue.

Signed-off-by: Chen Qi <Qi.Chen@windriver.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -26,7 +26,12 @@ pkgdatadir = join_paths([prefix, get_option('datadir'), package_name])
 bindir = join_paths([prefix, get_option('bindir')])
 libexecdir = join_paths([prefix, get_option('libexecdir')])
 schemadir = join_paths([datadir, 'glib-2.0', 'schemas'])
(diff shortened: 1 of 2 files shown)


eli-schwartz:
The proper solution is probably to move away from the deprecated `python3` module and switch to the `python` module. Either `install_sources()` or `get_install_dir()`.

This supports cross compilation, since you can select the right python via a cross file (and Meson will extract the necessary info from there), and also this module already comes with a builtin Meson option `-Dpython.purelibdir` for overriding the site-packages directory with an absolute path.

infirit:
I forgot the reason I haven't moved to ``python`` in meson.

@ChenQi1989 I don't do any cross building and we still have autotools. Is ``--with-python_prefix`` the autotools equivalent?

ChenQi1989:
https://github.com/blueman-project/blueman/pull/1700 solves the issue.
@eli-schwartz @infirit Thanks for your feedback and quick fix.

PR #1700, opened 2022-04-03
Port meson from deprecated python3

infirit:
@ChenQi1989 try this and then set  `-Dpython.purelibdir`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -133,16 +133,65 @@ install_data(
     install_dir: join_paths([datadir, package_name, 'pixmaps'])
 )
 
+icon_makefiles = [
+    'Makefile.am',
+    '128x128/apps/Makefile.am',
+    '128x128/status/Makefile.am',
+    '128x128/Makefile.am',
+    '16x16/actions/Makefile.am',
+    '16x16/apps/Makefile.am',
+    '16x16/devices/Makefile.am',
+    '16x16/status/Makefile.am',
+    '16x16/Makefile.am',
+    '192x192/apps/Makefile.am',
+    '192x192/status/Makefile.am',
+    '192x192/Makefile.am',
+    '22x22/apps/Makefile.am',
+    '22x22/status/Makefile.am',
+    '22x22/Makefile.am',
+    '24x24/actions/Makefile.am',
+    '24x24/apps/Makefile.am',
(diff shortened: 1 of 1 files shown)


infirit:
Fixed bunch of other stuff like prefix.

meson.build
```diff
@@ -25,8 +25,8 @@ datadir = join_paths([prefix, get_option('datadir')])
 pkgdatadir = join_paths([prefix, get_option('datadir'), package_name])
 bindir = join_paths([prefix, get_option('bindir')])
 libexecdir = join_paths([prefix, get_option('libexecdir')])
-schemadir = join_paths([datadir, 'glib-2.0', 'schemas'])
```
eli-schwartz, line 28: You don't need to care about making schemadir absolute. It is only used as the install_dir to a function. Meson functions for installing always accept relative paths (e.g. `share/glib-2.0/schemas`) and install the files relative to the prefix.

You do need to ensure this is an absolute path if you intend to do something else with schemadir, such as configure it into a file.

infirit: Make sense, fixed. In this case I made a big mess as datadir already has prefix :fearful: 

At some point I'll rework the directory handling and pass relative paths and absolute when configuring.

ChenQi1989:
@infirit I've tested these patches in my cross compilation environment. They worked well. Thanks.

meson.build
```diff
@@ -47,10 +47,11 @@ conf_data.set('icondir', join_paths(datadir, 'icons'))
 conf_data.set('dhconfig', get_option('dhcp-config-path'))
 conf_data.set('POLKIT', have_polkit)
 conf_data.set('GETTEXT_PACKAGE', package_name)
-conf_data.set('PYTHON', pyprog.path())
+conf_data.set('PYTHON', pyinstall.path())
 
 # Check for build dependencies
-pythonlib = dependency('python-' + python.language_version(), required: true)
+#pythonlib = dependency('python-' + pyinstall.language_version(), required: true)
```
cschramm, line 53: Drop?

cschramm (approved):

meson.build
infirit: Yes, forgot to remove. Fixed

PR #1716, opened 2022-05-13
Move things around and add to Makefile.am

infirit:
Meson just installs the whole directory.

I moved the rss/tpl/lq version to the pixmap directory for now.


infirit:
@cschramm shall we just merge this and see handle the fallout (if any)?

PR #1725, opened 2022-05-26
UI: Switch to symbolic icons

clefebvre:
- Switch to symbolic icons in menus, toolbars, tray menus.
- Switch to emblems in the device list.
- Replace fullcolor action icons in net plugin with folder icon.

image


clefebvre:
Unrelated note: While messing with the blocked emblem, I noticed I could no longer summon the context menu on my device after blocking it. This made it impossible to unblock it without removing it completely.

cschramm:
Looks good to me. I like it. :slightly_smiling_face: 

I skimmed through the icon themes that we claim to support in [Dependencies.md](https://github.com/blueman-project/blueman/blob/main/Dependencies.md) and unfortunately it looks like we lose compatibility with MATE and Elementary as they do not have any bluetooth-disconnected icons, no application-x-addon-symbolic (MATE does have a non-symbolic version of it) and Elementary has only a non-symbolic gnome-power-manager. :confused: I think I'd prefer network-offline-symbolic and a blueman-plugin-symbolic for compatibility and report a bug to Elementary regarding gnome-power-manager-symbolic.

flake8 fails due to some lines that got too long with the `-symbolic` suffix.

clefebvre:
Thanks @cschramm,

I shipped the block and trust icons instead of just using existing names to ensure the emblems looked similar. That said we can make them look different. Would you prefer a forbidden sign for the block?

bluetooth-disconnected and application-x-addon-symbolic are part of Adwaita. Themes don't need to contain these icons they can just inherit them from gnome/adwaita/hicolor. Do you get a missing icon when switching to these themes?

We could of course ship a copy under a blueman name but we shouldn't need to.

I'll fix the flake8 issues in a commit tomorrow.

clefebvre:
Something like this for block?

dialog-error-symbolic

cschramm:
Hm, right, at least the Debian and Arch GTK+ 3 packages depend on Adwaita packages, so Adwaita should be in place and anything that's in it should be fine. :+1:

I like the sign for block much better than the strike-through icon, yes. :pray:

cschramm (approved):

cschramm:
Looks all good to me now. Still waiting for @infirit's opinion on dropping fullcolor. Nobody seems to be against it so far, but I definitely want to involve him.

infirit:
ngl, I like colours in my applications. Unfortunately the rest of the world disagrees with me so :shrug: .

But with this PR we at least have a more consistent look which is more important.

@clefebvre can you rebase and resolve the conflicts?

clefebvre:
@infirit sure, sorry for the delay, I'll try to do it today.

infirit (approved):

PR #1735, opened 2022-06-06
RFC: Remove translatable string location

infirit:
I don't know if this is a good idea or not but it will get rid of most of the updates to blueman.pot

We loose `Source string location` in weblate but I wonder, how useful is that for translators?

### DIFF
--- a/po/Makevars
+++ b/po/Makevars
@@ -8,7 +8,7 @@ subdir = po
 top_builddir = ..
 
 # These options get passed to xgettext.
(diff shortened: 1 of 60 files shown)


cschramm:
Well, in theory it makes sense to be able to check the context. I'm not sure if translators actually look at the locations and even if, I wonder if it really enables them to get the context.

We actually had locations disabled once and reverted that in #1317 after we switched to squashing PR automations for both the pot and the po files. If it bothers you anyway, I'm fine with disabling again. It does not really bother me (well, reviewing those PRs obviously gets a little easier without those line changes, but those reviews are not super critical anyway).

infirit:
Ah, forgot about that. No it's just slightly annoying, i just need to live with it 😄

PR #1736, opened 2022-06-06
Add Notificaton.set_message

cschramm:
The method can be used to replace the message especially of already shown notifications. _NotificationBubble.show() does keep track of the notification ID and uses the D-Bus protocol's replacing capabilities, so that we can just re-call it to replace an already shown notification. In case of _NotificationDialog it is sufficient to just set the property, no matter if the dialog is already shown or not.

### DIFF
--- a/blueman/gui/Notification.py
+++ b/blueman/gui/Notification.py
@@ -87,6 +87,9 @@ class _NotificationDialog(Gtk.MessageDialog):
         self.connect("enter-notify-event", on_enter)
         self.connect("leave-notify-event", on_leave)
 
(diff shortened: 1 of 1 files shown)


infirit:
Yeah this is a good start. We will want to change the icon as well. So we'll not to move that part out if ``__init__``.

cschramm:
Do we? I actually started with a more generic `update` method with optional keyword arguments that also supports changing the summary as a low hanging fruit, but I thought that the text content would be enough. 🤔

infirit (approved):

PR #1747, opened 2022-06-20
Tray: Support for symbolic icons

clefebvre:
This allows the tray to use symbolic icons instead of fullcolor icons.

Note this feature is configurable in gsettings and disabled by default.

To enable it, kill blueman, run the command below and restart blueman:

`gsettings set org.blueman.general symbolic-status-icons true`

Note: It would be nice to let users configure this setting graphically. However I didn't add UI configuration in this PR, for the following reasons:

- Although users might like to do this, the main use case I wanted to solve here was distro configuration. With this key in place, distributions can provide a gsettings override to properly integrate blueman to their desktop environment.
- Although the tray is a core part of Blueman, it's implemented via optional plugins. I wasn't sure how/where to expose such a setting within the UI.
- I wasn't keen on introducing a new msgid that late. If we're close to release this would mean part of the UI wouldn't be properly translated in many languages.
- Because the setting isn't exposed to the UI I didn't think it needed to be listened to for changes, so changing it has no effect until blueman is restarted.


infirit:
It's really easy to make this instant apply so I would prefer something like this.
```python
Config.connect("changed::symbolic-status-icons", lambda: self.icon_should_change())
```

As for the UI, if people want to control this then we could make it part of the first start wizard and expose it somewhere. But for now this this should good enough.

clefebvre:
It doesn't work for me @infirit, I'm not sure why. Not only does the icon not refresh when I change the gsettings value, but it doesn't even switch to or from symbolics when I disable/enable bluetooth. It's stuck either in fullcolor mode or symbolic mode until I restart blueman.

clefebvre:
arf, don't mind me.. it's late, I wasn't focused. I got it working. I'll push another commit to add it.

clefebvre:
@cschramm @infirit is this OK?

Is it possible to get a tag after this PR? We're getting really close to a BETA for Mint 21, we'd love to include Blueman in it.

cschramm (approved):

Issue #1696, opened 2022-03-30
Blueman-applet wrong battery percentage in Notification


Chr1s70ph:
blueman: 2.2.4-1
BlueZ: 5.63-2
Distribution: Linux manjaro 5.16.14-1-MANJARO
Desktop environment: i3wm

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Connecting my XBOX-Controller displays a wrong battery percentage in the notification. As seen in the logs below, blueman does fetch 50% first (thats what is sent in the notification) and only afterwards fetches the *correct* battery percentage.

```console
blueman-applet 15.39.31 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0/dev_F4_6A_D7_E6_73_BF {'Connected': True}
blueman-applet 15.39.32 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0/dev_F4_6A_D7_E6_73_BF {'Percentage': 50}
blueman-applet 15.39.32 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0/dev_F4_6A_D7_E6_73_BF {'ServicesResolved': True}
blueman-applet 15.39.36 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0/dev_F4_6A_D7_E6_73_BF {'Percentage': 78}
```

This is not a Controller driver issue, already discussed in XPADNEO-Issue [#342](https://github.com/atar-axis/xpadneo/issues/342)

cschramm:
Sounds like that's what the device provides via Bluetooth. You can double-check where BlueZ gets its values from by enabling debugging output for bluetoothd using `-d src/battery.c,profiles/battery/battery.c` (or even look into the raw data with `btmon` if you feel adventurous).

Our Battery applet plugin currently shows the percentage on connection and that is obviously the 50 %. I think there's nothing we can do to avoid it. What we could do is change the concept of the plugin and send notifications on changes as well, so that at least the 78 % will get reported (~4 seconds later).

infirit:
I don't have an Xbox controller but I'll check how hid-nintendo handles this. I don't think it currently reports battery level at all but if it does I'll report back.

Chr1s70ph:
<details>

<summary>This is the output of bluetoothd</summary>

```console
bluetoothd[46166]: Bluetooth daemon 5.63
[...]

PR #1751, opened 2022-06-24
Update instead of delaying connection notifications

cschramm:
I really don't like the generic delay of 5 seconds for all connection notifications. This is a quick draft to show instant notifications and update them with battery levels for up to five seconds instead.

It works fine in a quick test, however, when the device just got connected, it does not have the battery (percentage) yet, so that line 37 always fails. `_on_battery_created` does update the message as expected but this breaks the icon logic as the battery icon will never be used. Not sure if a way to update the icon is the way to go there or not. I also could not actually test the `property-changed` signal.

### DIFF
--- a/blueman/gui/Notification.py
+++ b/blueman/gui/Notification.py
@@ -90,6 +90,9 @@ class _NotificationDialog(Gtk.MessageDialog):
     def set_message(self, message: str) -> None:
         self.props.secondary_text = message
 
+    def set_notification_icon(self, icon_name: str) -> None:
+        self.set_icon_from_icon_name(icon_name, 48)
+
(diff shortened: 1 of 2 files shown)


cschramm:
Added a second commit that updates the icon. Works fine and does not feel awkward to me.

infirit:
While this works for the devices I have it won't for the xbox one controller in  #1696. The notification will be gone before it's updated. So a timeout is needed of at least 4 seconds.

cschramm:
Depends on the notification daemon's default timeout but from some quick tests I don't expect it to break in general as updating an already closed notification simply leads to a new notification. Might seem a little awkward but well the controller's behavior is just exactly that. 😆

infirit (approved):
Ok then. Hopefully folks don't get confused 🤣

PR #1755, opened 2022-06-25
Fix duplicated bluetooth adapter menu entries

infirit:


cschramm:
Hmm. I guess it's not such a good idea that the `Manager` class' `populate_devices` calls `_on_object_added` no matter if the object actually got added or not. It obviously leads to duplicate events and we need to check all handlers for resilience against that. From a quick survey, I'm a bit worried about `ConnectionNotifier._on_battery_created` and `DeviceList.add_device`.

cschramm (approved):

infirit:
The bug was actually somewhere else and my fix was wrong. DisplayKnownDevices never handled adapters.

PR #1774, opened 2022-07-05
Hide recent connections associated with unavailable adapters

### LINKED ISSUE #1771: Recent connections being greyed out
blueman: 2.2.4
BlueZ:
Distribution: manjari
Desktop environment: xfce4

I have moved my disc into a new laptop (still a thinkpad) and the items in the recent connections lists are greyed out. My guess is, the adapter id or something just doesn't match up with the id of the bluetooth device on my new laptop hence it gets greyed out. I figured out to use `dconf-editor` to modify `org.blueman.plugins.recentconns.recent-connections` (didn't make a copy or screenshot, sorry) to remove the old items.
The expected behaviour would be to just skip those items instead of greying them and wasting space in the list.

cschramm:
The current behavior dating back to blueman 1 is that connections for unavailable adapters get shown insensitive while connections for available adapters but unavailable devices get hidden. There does not seem to be any good reason for that difference.

To be able to use an assignment expression for a nice solution, I preceded a raise of the Python version to 3.8. We can separate that, of course.

Closes #1771


infirit:
Ah the controversial walrus operator 😃. Bumping to 3.8 is fine with me. Can you drop 3.7 and 3.6 in the checks so ci doesn't fail?

infirit (approved):

PR #1776, opened 2022-07-08
Replace MessageArea with standard Gtk.InfoBar

infirit:
We don't really need one, Gtk.InfoBar has everything we need already. The rest is a small fix and rework.


blueman/main/Manager.py
```diff
@@ -72,17 +71,16 @@ def do_activate(self) -> None:
...
             self.window.connect("configure-event", self._on_configure)
 
-            grid = self.builder.get_widget("grid", Gtk.Grid)
-
             toolbar = self.builder.get_widget("toolbar", Gtk.Toolbar)
             statusbar = self.builder.get_widget("statusbar", Gtk.Box)
 
+            self._infobar = self.builder.get_widget("message_area", Gtk.InfoBar)
+            self._infobar.connect("response", self._infobar_response)
+            self._infobar_bt: str = "empty"
```
cschramm, line 79: Shouldn't this just be an empty string?

cschramm (approved):

blueman/main/Manager.py
infirit: Yeah, leftover from testing. I'll fix that.

infirit: Fixed.

cschramm (approved):

Issue #1759, opened 2022-06-28
Failed to apply network settings


alexandremsouza1:
blueman: 2.2.5
BlueZ:
Distribution: Ubuntu 22.04 LTS x86_64
Desktop environment: GNOME

I just updated and now I'm having this problem. My headphones are currently working fine. I'm just getting this dialog every time I go into plugins or when the system boots.

![image](https://user-images.githubusercontent.com/38253342/176175072-8ba441ef-9f23-4ccb-9e53-022eff1cc0c7.png)

cschramm:
Ubuntu 22.04 does not provide blueman 2.2.5. Are you using the official 2.2.5-1 Debian package?

Please stop blueman-applet and start it manually with `blueman-applet --loglevel debug` to get proper output.

infirit:
From the message it most likely means you haven't installed the dbus service file correctly. Are you building blueman yourself?

alexandremsouza1:
no, I installed from apt-get `deb http://cz.archive.ubuntu.com/ubuntu kinetic main universe`

clefebvre:
This same error was reported in Mint 21 BETA (blueman 2.3): https://github.com/linuxmint/mint21-beta/issues/98

cschramm:
I'm able to provoke a similar exception with

```python
from blueman.main.DBusProxies import Mechanism
m = Mechanism()
```

```sh
sudo killall blueman-mechanism
```

```python
m.AnyMethod()
```

i.e. when our proxy object for the mechanism service survives the service itself. The Gio DBusProxy seems to resolve the well-known bus name `org.blueman.Mechanism` into a unique connection name and remembers just the latter, so that when it's gone everybody is lost.

That situation is totally unexpected, though. As the initialization of the proxy succeeds, we can conclude that blueman-mechanism started successfully and its name appeared on the bus. The method call then gets committed immediately, so that there's no chance that the service already reached its 30 seconds timeout. It seems like it's dying unexpectedly somewhere in `blueman.main.MechanismApplication.MechanismApplication.__init__`.

`journalctl -u blueman-mechanism.service` should be helpful.

cschramm:
Not really, `ReloadNetwork` gets called when the `Networking` applet plugin gets loaded. In main, #1564 actually avoids that by returning early from `_apply_nap_settings`.

clefebvre:
[...]

PR #1795, opened 2022-07-23
Networking: Don't load if nap is not enabled

infirit:
It *may* cause errors to show even when no bluetooth is available.
see #1759 for refference.


cschramm:
Not sure if we should do this as it just covers an unexpected issue with blueman-mechanism that we should definitely try to find.

Issue #1248, opened 2020-05-10
blueman-services does not allow a user to save transfer settings


S4qFBxkFFg:
blueman: 2.1.2-1 (focal)
BlueZ: 5.53-0ubuntu3 (focal)
Distribution: Ubuntu 20.04 (focal)
Desktop environment: XFCE 4.14

When the blueman applet is enabled, right-clicking and selecting "Local Services..." then "Transfer" does not allow a new destination folder to be selected and saved.

Instead, the new folder can be selected, but not saved as the "Apply" button is greyed out.

When running the command as root, the "Apply" button is available, but this only affects the root user.

I just tried upgrading to the latest version of Ubuntu, but the problem remains.  I also tried creating a new user on this machine, and the same problem resulted.

terminal log from starting the applet, attempting to change destination folder, and exiting:
https://pastebin.com/7kz9fhdy

Please let me know if any logs are required; also reported at: https://bugs.launchpad.net/ubuntu/+source/blueman/+bug/1877131

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

<!--

ℹ Please use some service like [Pastebin](http://pastebin.com/) or [GitHub Gist](https://gist.github.com/) to post your logs and keep the thread clean and readable. Make sure to describe the exact steps you took when producing it.

->

cschramm:
> blueman: 2.1.5-1 (focal)
Should be 2.1.2 I guess.

As a first try, please apply this patch:

```diff
--- /usr/lib/python3/dist-packages/blueman/plugins/services/Transfer.py
+++ /usr/lib/python3/dist-packages/blueman/plugins/services/Transfer.py
@@ -42,6 +42,7 @@
         self.widget.props.visible = False
 
     def on_property_changed(self, config, key):
+        logging.debug(key)
         value = config[key]
 
         if key == "shared-path":
```

run `blueman-services --loglevel debug`, replicate the issue, and provide the output.

S4qFBxkFFg:
> Should be 2.1.2 I guess.

Apologies - typo when updating the post after upgrading Ubuntu - corrected.

The requested output is:
```
blueman-services --loglevel debug
blueman-services version 2.1.2 starting
blueman-services 11.43.04 INFO     Services:90 load_plugins: ['Network', 'Transfer']
blueman-services 11.43.05 INFO     Services:128 set_page  : Set page Network
[...]

PR #1798, opened 2022-07-23
Drop Config class

infirit:
The only use it had was backward compatibility with pygobject 3.11.2. Which we don't support any more so drop it.


blueman/main/Config.py
```diff
@@ -1,22 +0,0 @@
...
-        else:
-            super().__init__(schema=schema_id, path=path)
-
-    def bind_to_widget(self, key: str, widget: GObject.Object, prop: str,
-                       flags: Gio.SettingsBindFlags = Gio.SettingsBindFlags.DEFAULT) -> None:
-        self.bind(key, widget, prop, flags)
-
-    def __setitem__(self, key: str, value: Any) -> None:
-        if not self.is_writable(key):
-            logging.error("GSetting not writable, settings not saved")
```
cschramm, line 21: Just to point it out: We loose this log message that we added in the course of #1248.

cschramm (approved):

blueman/main/Config.py
infirit: Yeah. I don't think it's all that useful, certainly not useful enough to keep the Config class for. But thanks for pointing it out 👍

PR #1799, opened 2022-07-25
Recent connections menu is missing in lxqt-panel

niknah:
In lxqt-panel.  GetLayout is called twice.  Once with parent_id = 0, then parent_id = 7(recent connections).  But it doesn't rebuild the recent connection submenu only the menu item.
I am using lxqt-panel v1.1.0

### DIFF
--- a/blueman/main/indicators/StatusNotifierItem.py
+++ b/blueman/main/indicators/StatusNotifierItem.py
@@ -37,10 +37,9 @@ class MenuService(DbusService):
             return self._revision, (0, {}, self._render_menu(self._items, 1, self._render_submenu))
         else:
             item = self._items[parent_id - 1]
-            if "submenu" in item:
+            if "submenu" in item and _recursion_depth != 0:
                 return self._revision, (parent_id, self._render_item(item), self._render_submenu(item, parent_id ))
-            else:
-                return self._revision, (parent_id, self._render_item(item), [])
+            return self._revision, (parent_id, self._render_item(item), [])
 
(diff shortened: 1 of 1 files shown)


blueman/main/indicators/StatusNotifierItem.py
```diff
@@ -36,7 +36,11 @@ def _get_layout(self, parent_id: int, _recursion_depth: int, _property_names: Li
         if parent_id == 0:
             return self._revision, (0, {}, self._render_menu(self._items, 1, self._render_submenu))
         else:
-            return self._revision, (parent_id, self._render_item(self._items[parent_id - 1]), [])
+            item = self._items[parent_id - 1]
+            if "submenu" in item:
+                return self._revision, (parent_id, {}, self._render_submenu(item, parent_id ))
```
cschramm, line 41: Are you sure `{}` is correct here? Shouldn't it just be
```
return self._revision, (parent_id, self._render_item(self._items[parent_id - 1]), self._render_submenu(item, parent_id))
```
for all `parent_id != 0`?

niknah: Yes, you're right.  The audio profile menu had disappeared with that change.

cschramm: There's no need for the `else` branch then as `_render_submenu` handles items without submenus just fine.

Also there's a whitespace error that makes mypy unhappy.

niknah: Thanks, I have changed this and added a check for the _recursion_depth parameter

cschramm (approved):

PR #1800, opened 2022-07-25
Update nl.po

Pjotr123:
Almost complete Dutch translation. Revised.


cschramm:
Very nice but our authoritative source for translations is https://hosted.weblate.org/projects/blueman/. Changes made here would get overridden from or cause conflicts there. Please apply your changes on Weblate.

infirit:
Yeah, please use weblate. I do however not agree with your translation of adapter to stekker (plug). 

Geüpload and gedownload are in my opinion the right translation.

PR #1801, opened 2022-07-26
Re-register StatusNotifierItem on name owner changes

cschramm:


infirit (approved):
I have no experience with this interface but lgtm :smile:

cschramm:
Well, it makes sense as a new host does not know about existing items and the patch makes the item re-appear at least when reloading xfce4-panel or mate-panel.

PR #1804, opened 2022-07-28
Polish readme a little bit

cschramm:
* Remove explicit cell phone technologies.
* Remove Launchpad reference.
* Use badges for Repology and Weblate


infirit (approved):
Badges look nice 👍

PR #1812, opened 2022-08-03
Add emblem for connected state

cschramm:
This could solve the problem mentioned in #1642 as an alternative to the #1620 concept.

I think it's fair to just use the spot of the "paired" emblem and override it with the "connected" emblem. It's not very distinguishable right now, though. One might probably want a different color for it. 

The icon's content is basically stolen from Adwaita.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -258,15 +258,16 @@ class ManagerDeviceList(DeviceList):
 
         return icon_info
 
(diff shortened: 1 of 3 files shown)


infirit:
Yeah, this is a reasonable approach for now. I think the icon is fine so ok with me.

infirit (approved):

cschramm:
Still not satisfied with this, but I tend to merge it to go on with #1642. Maybe @clefebvre, you'd want to have an artistic look at this? 😅

PR #1841, opened 2022-08-30
Drop link quality

cschramm:
This value is really pointless, especially as it is vendor specific. Also it's only available via libbluetooth, so BlueZ 6 will not have it anymore.

This is more or less a subset of https://github.com/blueman-project/blueman/pull/1642.


infirit (dismissed):
lgtm. I hope to get back into blueman in the coming month :-)

Issue #1555, opened 2021-06-13
Bluetooth doesn't turn on automatically after suspend/resume


hamkg:
blueman: 2.2.1
BlueZ: 5.58
Distribution: Slackware64 -current
Desktop environment: MATE 1.24

After upgrading to blueman-2.2.1, bluetooth is turned off when suspending and remains so after resume. No big deal since I can easily turn it on with the blueman applet, but I would prefer it turning on automatically.

A bug, or intentional change? Something I can change in /etc/bluetooth/main.conf?

See lowlevel debug output got after starting blueman-applet, then going into suspend, then resume: https://pastebin.com/Jn9ZAssT

* [ ] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

<!--

ℹ Please use some service like [Pastebin](http://pastebin.com/) or [GitHub Gist](https://gist.github.com/) to post your logs and keep the thread clean and readable. Make sure to describe the exact steps you took when producing it.

->

cschramm:
Are you saying this behavior differs between 2.1.4 and 2.2.1 (without any other changes on them system!)? That's should not be intentional, no.

Am I right that there's actually no output when resuming?

hamkg:
Yes, I tried reverting to 2.1.4 and after suspend/resume Bluetooth was active at once. Upgraded once again to 2.2.1 and BT stays off after resume (red X on the systray icon, menu items greyed out). After clicking "Turn Bluetooth On" everything works as expected.

cschramm:
I guess the behavior highly depends on the system. For me, I always see the Powered property of the adapter changed to False on suspend and back to True on resume. In your log, it only gets changed to False but not changed back to True. Blueman can auto-power the adapter for you if you enable that option in the settings of the PowerManager plugin.

I do not get why 2.1 and 2.2 should show different behaviors here, though, as when the option is not enabled, they do not interfere with the power states at all. :thinking: Maybe the icon in 2.1 just tells you lies and it's actually not powered. You can check the state e.g. with `bluetoothctl show`.

hamkg:
The icon in 2.1.4 is truthful: no trouble at all getting a BT connection after suspend/resume.
I can see the same difference as you noticed when comparing logs from 2.1.4 and 2.2.1, but that doesn't help me understand why this is happening.
Maybe hardware-related? 
I've put a question in the Slackware section on LQ and noone has mentioned the same experience.
[...]

PR #1843, opened 2022-09-03
Drop auto-power

cschramm:
I think this feature does more harm than good nowadays.

BlueZ introduced a similar `AutoEnable` option in 5.35. At least Debian and its derivatives enable it by default. Distros that love upstream configurations like Arch Linux did not but [BlueZ 5.65 finally changed the default](https://git.kernel.org/pub/scm/bluetooth/bluez.git/commit/?id=180cf09933b2d8eb03972c8638063429fe5fece5).

Compared to `AutoEnable` `auto-power` has some non-desirable differences in behavior: At least it does not cover suspend and resume, see #1555, and together with the `KillSwitch` plugin it overrides killswitches as the plugin tries to unblock them when powering is requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Hide recent connections associated with unavailable adapters
 * Store network configuration in GSettings instead of /var/lib/blueman/network.state.
 * Replace custom MessageArea widget with regular Gtk.InfoBar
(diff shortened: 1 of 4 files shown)


infirit (approved):
omg, hell froze over and they actually changed the default :scream: .

Issue #1845, opened 2022-09-06
System tray menu looks wrong


resiehnnes:
blueman: 2.3.2
BlueZ: 5.65
Distribution: Fedora 36
Desktop environment: Gnome 42.4, X11

System tray menu looks kinda messy when I turn the bluetooth on. When it's off it looks alright.

Bluetooth OFF:

![Screenshot from 2022-09-06 13-53-14](https://user-images.githubusercontent.com/52158526/188630083-f4aa7625-c156-47e3-8295-a4774b0e7a0b.png)

Bluetooth ON:

![Screenshot from 2022-09-06 13-53-30](https://user-images.githubusercontent.com/52158526/188630078-71170850-085c-4b66-935d-097b635c0116.png)

cschramm:
What's rendering those? To my knowledge Gnome itself does not have the concept of tray icons anymore. :raised_eyebrow:

resiehnnes:
I am using [Dash to Panel](https://github.com/home-sweet-gnome/dash-to-panel) + [KStatusNotifierItem](https://github.com/ubuntu/gnome-shell-extension-appindicator) First extension gives a 'Windows' like panel, where's the other one is for tray icons.

cschramm:
Ok, sounds like we're having a compatibility issue with [gnome-shell-extension-appindicator](https://github.com/ubuntu/gnome-shell-extension-appindicator)
Public then. I'll try to have a look. Can you confirm that blueman 2.2 works fine with it?

resiehnnes:
Yep, looking good with 2.2.3

infirit:
I can't reproduce this on my fedora 36 vm

cschramm:
Weird, I can, without doing anything fancy. :confused:

Looks like that thing has a fundamental flaw at least when changing items between separators and standard ones. Providing it with unique identifiers solves the problem for me. -> #1847

kuraga:
blueman 2.3.2, XFCE 4.16

![image](https://user-images.githubusercontent.com/1063219/189484742-12553e31-35cd-41e3-9be6-f219ad27d6d6.png)

(sometimes without `Label Empty`)

PR #1847, opened 2022-09-08
Do not re-use dbusmenu item identifiers

cschramm:
Re-using identifiers e.g. for separator and standard menu items triggers bugs at least in gnome-shell-extension-appindicator (#1845).

As a workaround we can just use the toplevel items' priority values instead of sequential identifiers to get unique identifiers for all items.

To do that the MenuItem's priority value gets added as `id` to the MenuItemDict serialization and picked up by the IndicatorInterface implementations. Note that the semantics of the first argument ot the MenuItemActivator change accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Replace custom MessageArea widget with regular Gtk.InfoBar
 * Drop auto-power feature. BlueZ now has the AutoEnable setting for even better auto-powering.
 * Do not use pointless link quality value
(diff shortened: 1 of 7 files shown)


infirit (approved):
I really haven't looked too closely but nothing stands out so let's merge and see what comes up next.

cschramm:
mypy has us covered. :grimacing:

cschramm:
#1652 contains a fix (for GetLayout with id != 0 as used by LXQt) and a cleanup.

Issue #1864, opened 2022-09-27
blueman-manager fails to start from tray icon


fdservices:
bluez 5.65-3
blueman 2.3.2-1
arch linux 5.19.10-arch1-1

Bluetooth experimental is set to false

Starting blueman-manager from a terminal produces some error output, **before its starts correctly**

```python
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/blueman/gui/manager/ManagerDeviceList.py", line 518, in on_finished
    callback()
  File "/usr/lib/python3.10/site-packages/blueman/gui/manager/ManagerDeviceList.py", line 308, in <lambda>
    self._prepare_fader(row_fader, lambda: self.__fader_finished(device))
  File "/usr/lib/python3.10/site-packages/blueman/gui/manager/ManagerDeviceList.py", line 312, in __fader_finished
    super().device_remove_event(device)
  File "/usr/lib/python3.10/site-packages/blueman/gui/DeviceList.py", line 175, in device_remove_event
    assert tree_iter is not None
AssertionError
```

I suspect that it is this error that causes the problem.

infirit:
I can't replicate this exact issue but I can reproduce it in another way. Quick fix (not solution) is below


```python

diff --git a/blueman/gui/DeviceList.py b/blueman/gui/DeviceList.py
index a0af7e06..37f757fd 100644
--- a/blueman/gui/DeviceList.py
+++ b/blueman/gui/DeviceList.py
@@ -172,7 +172,9 @@ class DeviceList(GenericList):
     def device_remove_event(self, device: Device) -> None:
         logging.debug(device)
         tree_iter = self.find_device(device)
-        assert tree_iter is not None
+
+        if tree_iter is None:
+            return
 
         if self.compare(self.selected(), tree_iter):
             self.emit("device-selected", None, None)
```

cschramm:
I don't think that error is a problem as it's in the context of the `animation-finished` event and the event handler can cope with such errors so that they are non-fatal. Thus blueman-manager starts correctly anyway.

Starting from the tray icon is generally possible in two way: Left-click on the icon (depends on the icon implementation and server) or via the "devices" item in its (right-click) menu. I assume that both ways do not work for you, correct?

To get behind that problem, you can

- check `~/.xsession-errors` for errors.
- try the left-click mechanism manually e.g. with `dbus-send --print-reply --dest=org.blueman.Applet /org/blueman/Applet org.blueman.Applet.Activate`.
[...]

PR #1865, opened 2022-09-28
Do dbus activation like Gedit

infirit:
There are plenty of GtkApplication apps that are dbus activatable, basically all of gnome. Gedit and other Gnome apps do it like this so let's copy it. Maybe this helps with missing DISPLAY variable (:eyes:) but this is how it supposed to be done anyway so here we go.

I added this only for the actual Gio/GtkApplications

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -22,7 +22,7 @@ if __name__ == '__main__':
     parser = parser = create_parser()
     parser.add_argument("--socket-id", dest="socket_id", action="store", type=int, metavar="ID")
     parser.add_argument("adapter", nargs="?", metavar="ADAPTER NAME")
-    args = parser.parse_args()
+    args, other_args = parser.parse_known_args()
 
     if args.LEVEL.upper() == "DEBUG":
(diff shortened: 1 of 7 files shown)


cschramm:
Trying to understand things... So, the argument tags the GApplication a service and, as far as I understand, a service has an extra 10 seconds timeout in which the activating message may arrive and get processed before it quits when its use count became zero.

I think that should not be relevant for blueman-applet as we call `hold` in its `do_activate` method and never release it. blueman-manager's application's `hold` method probably gets called by the GtkApplicationWindow and I guess it releases when it gets closed or something. The application seems to terminate immediately anyway. I can both close the window within the first 10 seconds and see the process die immediately and still replicate the `DISPLAY` misconfiguration as in https://github.com/blueman-project/blueman/issues/1864#issuecomment-1260712168 with `GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying (4)` from blueman-applet.

cschramm (approved):
I don't think this really changes anything, but it does not hurt and does follow GLib recommendations.

PR #1866, opened 2022-09-29
Update types for mypy 0.981

cschramm:
The typeshed update in 0.981 introduces the _Pointer type and makes pointer a function. Using _Pointer seems like the easiest solution here but note that it is not available in older versions.


infirit (dismissed):
Should be fine relying on this or later versions. If it does become an issue we can pin it again later.

infirit:
Is sonarcloud complaining about _Pointer not being python2 compatible?

Issue #1870, opened 2022-10-06
Duplicate tray icon


clefebvre:
blueman: 2.3.2
BlueZ: 5.64-0ubuntu1
Distribution: Mint 21
Desktop environment: Cinnamon 5.4

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

This is an issue which seems to happen randomly. We've seen people talk about it on the Mint forums. I got lucky here on this computer, I have it in a way that it seems to happen all the time on every reboot.

Blueman provides two tray icons in the Cinnamon panel:

![image](https://user-images.githubusercontent.com/1138515/194301534-f46b3638-944c-46c0-851a-c86e0848431e.png)

The StatusItemNotifier plugin is enabled:

![image](https://user-images.githubusercontent.com/1138515/194301863-39fe1d1e-b1ed-473d-bc7c-457ca38e96d5.png)

Both the applet and the tray processes are running:

![image](https://user-images.githubusercontent.com/1138515/194301959-000ee880-dcc4-4709-95f1-b0cadcdb3801.png)

Here are the settings used by blueman:

![image](https://user-images.githubusercontent.com/1138515/194302169-e8fc79a8-fbb7-440e-801d-d891c168719f.png)

@mtwebster the left tray icon has the following characteristics:

- Its tooltip is "Blueman"
- Its context menu is blank
- It shows/hides the Blueman manager on left click
- It's provided by XappStatusApplet (I can see that by pressing Ctrl+Right-Click -> About...)

The right tray icon seems to be an xembed:

- Its tooltip is "Bluetooth Enabled"
- Its context menu is properly populated
- It shows/hide the Blueman manager on left click
- Cinnamon doesn't show which applet provides it.... when clicking Ctrl+Right-Click, we still get the Blueman context menu.

Cinnamon's logs:

![image](https://user-images.githubusercontent.com/1138515/194303222-45e5fed8-55d7-481d-a027-f1d74204f3f3.png)

We see both an XappStatusIcon and a systray here:

```
info t=2022-10-06T12:30:08Z Adding XAppStatusIcon: blueman (:1.47/org/x/StatusIcon/Icon_1)
info t=2022-10-06T12:30:08Z Adding systray: blueman-tray (24x24px)
```
[...]

PR #1871, opened 2022-10-06
Improve handling for new StatusNotifierWatcher

cschramm:
Earlier we just re-registered our item when a new StatusNotifierWatcher showed up but there are actually three cases where this happens:

* In case the watcher got replaced. This is what #1801 actually targeted and works fine for.
* Initially if a watcher is available. This means we actually register the item twice: Directly from within the constructor and from the handler. Does not hurt but is unnecessary.
* In case a watcher actually just showed up. In this case, we are already using the GtkStatusItem fallback, though, and this thus leads to two icons, see #1870.

Instead of re-registering, this change causes blueman-tray to reload if a watcher appears after we failed to register the item or a watcher appears for the second time.

### DIFF
--- a/blueman/main/Tray.py
+++ b/blueman/main/Tray.py
@@ -38,8 +38,7 @@ class BluemanTray(Gio.Application):
         for indicator_name in applet.GetStatusIconImplementations():
             indicator_class = getattr(import_module('blueman.main.indicators.' + indicator_name), indicator_name)
             try:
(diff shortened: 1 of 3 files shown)


infirit (approved):

clefebvre:
@cschramm I can confirm it fixes https://github.com/blueman-project/blueman/issues/1870

PR #1872, opened 2022-10-06
Cherries and preparation for 2.3.3

cschramm:


cschramm:
:thinking: I don't get the problem with mypy on Python 3.6. According to the docs, it *does* have `ctypes._Pointer` (it has always been present since 2.5).

infirit (approved):
Maybe just override it this time. I also have no clue why it's having a fit.

cschramm:
Oh, sure. mypy 0.981 dropped support for 3.6 so that on 3.6 the CI installs 0.971 and 0.971 does not have `ctypes._Pointer` in its typeshed.

infirit:
should we pin to the last version it's supported?

cschramm:
We could pin to 0.971 and revert https://github.com/blueman-project/blueman/pull/1872/commits/f2ee33b866468670b53b549ae0eccd24aadccd7f.

PR #1877, opened 2022-10-07
ManagerDeviceMenu: Do not assume all instances have been generated

infirit:
If the right click instance has never been used and the user double clicks to connect it will not have a SelectedDevice yet.


cschramm (approved):

sudoaza:
Wouldn't it be better to skip the instances that have not been generated with `next` instead of `return`ing from the loop?

sudoaza:
Thank you @infirit <3

cschramm:
@infirit: We fail to lock the "Device" menu with this.

If I double-click a device to connect, the right-click menu gets locked, i.e. it shows only a single, disabled "Connecting..." item. The same is expected for the "Device" menu in the menu bar but it's not the case. If I select a different device and then switch back, the menu is locked as expected as it got re-generated in the meantime.

infirit:
ok, generating all instances it is then. I'll open a new PR for this.

PR #1880, opened 2022-10-10
Notify user when we fail to reach blueman-manager via D-Bus

cschramm:


cschramm:
Can be tested with `DISPLAY= dbus-update-activation-environment DISPLAY; dbus-send --print-reply --dest=org.blueman.Applet /org/blueman/Applet org.blueman.Applet.Activate; dbus-update-activation-environment DISPLAY`

infirit (approved):

PR #1891, opened 2022-10-14
Define relevant plugin types for BluemanApplet.Plugins

cschramm:
This is rather ugly, especially as it breaks the dynamic nature of the PluginManager and involves a cast, but it is definitely an improvement as it would have caught #1887 early (would need backporting to 2-3-stable for that of course - RecentConns is ugly).

There might be some generic way of defining this, like `__getattr__(self, key: Literal[_T]) -> _T` but I did not find a way to actually express something like "literal name matching a typevar's value".

### DIFF
--- a/blueman/main/Applet.py
+++ b/blueman/main/Applet.py
@@ -1,4 +1,7 @@
-from typing import Any
+from gi.repository import Gio, GLib
+import logging
+import signal
+from typing import Any, cast
 
 from blueman.Functions import *
 from blueman.bluez.Manager import Manager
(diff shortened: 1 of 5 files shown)


infirit (approved):
I'll be honest, most of the more complex typing stuff in blueman I don't understand. But if you feel this is going to help :+1:

cschramm:
Let me add some description then:

Key here is to change the return value of `PluginManager.__getattr__` from `Any` to `object`. That will be the type that mypy resolves e.g. `BluemanApplet.Plugins.Menu` to 

With `Any` that value may get used in any way as no specific type is assumed. That's why mypy did not catch that https://github.com/blueman-project/blueman/pull/1872/commits/1297d8806e416f90a4dd0bfcef6e76849ef65675 uses `BluemanApplet.Plugins.Menu.add` with types that were not supported by the actual method in the `2-3-stable` branch.

With `object` mypy assumes that the value does not even *have* an `add` property. I thus add an explicit
```python
    @property
    def Menu(self) -> Menu:
```
so that it's defined that `BluemanApplet.Plugins.Menu` is of the `Menu` type and the call gets correctly checked. Anything else still gets handled by the `__getattr__` method and thus gets treated like an arbitrary `object`.

Adding this to 2.3.3 and running mypy 0.971 gives
```
blueman/plugins/applet/PulseAudioProfile.py:80: error: Argument 2 to "add" of "Menu" has incompatible type "Tuple[int, int]"; expected "int"
blueman/plugins/applet/DisconnectItems.py:40: error: Argument 2 to "add" of "Menu" has incompatible type "Tuple[int, int]"; expected "int"
```
which is exactly the problem in #1887 that makes 2.3.3 completely useless.

PR #1893, opened 2022-10-15
Use Gio Resource for UI, Icon and Pixmaps

infirit:
My end goal is to drop Constants.py. But also this makes it nicer to deal with in general.

- A single file for all our UI, Icons and Pixmaps
- Register once and it will be available throughout the application
- GtkIconTheme handles out hicolor icons transparently
- Less global variables to deal with from Constants.py

It currently breaks
- ~~StatusNotifier~~
- ~~Desktop file(s)~~


infirit:
I forgot to add. Although this works I don't have it tested well and I may have missed something. I'll deal with the errors when I undraft it. And any comments or criticisms welcome.

infirit:
This is good to go. Any objections, suggestions, changes?

cschramm:
Is this expected to work with autotools from source? I just gave it a try and the apps look for `data/org.blueman.gschema` but there's only the xml file. I guess that'd currently need a manual step, right?

cschramm:
This breaks things where we provide icon names other than `blueman` externally, at least:

- The `blueman-adapters.desktop` file where we reference `blueman-device`.
- The `StatusNotifierItem` implementation where we forward whatever icon name the applet provides, currently at least `blueman-tray`, `blueman-active`, `blueman-disabled` and `-symbolic` counterparts.

PR #1897, opened 2022-10-16
GameControllerWakelock: Use dbus to inhibit screensaver

infirit:
This actually works for me on KDE. I'll test XFCE later.

### DIFF
--- a/blueman/plugins/applet/GameControllerWakelock.py
+++ b/blueman/plugins/applet/GameControllerWakelock.py
@@ -1,25 +1,10 @@
 from gettext import gettext as _
 import logging
-from typing import Any
+from typing import Any, Dict, Optional
 
 from blueman.bluez.Device import Device
-from blueman.Functions import launch
 from blueman.plugins.AppletPlugin import AppletPlugin
-from blueman.plugins.errors import UnsupportedPlatformError
-
-import gi
-gi.require_version('Gdk', '3.0')
-try:
-    gi.require_version('GdkX11', '3.0')
-except ValueError:
-    raise ImportError("Couldn't find required namespace GdkX11")
-
-from gi.repository import Gdk
-from gi.repository import GdkX11
-
-
-if not isinstance(Gdk.Screen.get_default(), GdkX11.X11Screen):
-    raise UnsupportedPlatformError('Only X11 platform is supported')
+from gi.repository import Gio
 
 
 class GameControllerWakelock(AppletPlugin):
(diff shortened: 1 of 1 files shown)


infirit:
Joy, Xfce's screensaver doesn't implement `org.freedesktop.ScreenSaver`

infirit:
But luckily it does use the same arguments and function names. Looked at Gnome in a VM and it also implements `org.freedesktop.ScreenSaver`. Any other relevant desktops I need to check?

cschramm:
I'm not sure if that's a proper way. At least it's compatibility hell. There's `org.xfce.ScreenSaver`, `org.mate.ScreenSaver`, `org.cinnamon.ScreenSaver`, `com.deepin.ScreenSaver`, ... and other lockers do not have such an interface at all.

infirit:
If they expose `org.freedesktop.ScreenSaver` I don't have to specifically handle them. xdg-screensaver doesn't work for me on plasma and I doubt it still work on gnome. I'll have a look at the others and see what they put on dbus.

Inhibit from GtkApplication does absolutely nothing for me.

infirit:
> > If they expose org.freedesktop.ScreenSaver I don't have to specifically handle them.
> 
> They don't. The names I mentioned are both the bus and interface names. Object paths differ as well. I think I saw `Inhibit` methods on all of those interfaces.
> 
> > xdg-screensaver
> 
> You mean the implementation that we currently have in place? Shoot. I thought we're only talking about simplifying things and extending to Wayland, not fixing. sweat_smile

Yup, I just tested this on the fresh debian 11 vm with xfce4, it's not working. Screen blanks after the timeout of 1 minute that I set.

> [xdg-screensaver](https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-screensaver.in) does support KDE, though. It actually uses the same interface as you do here for KDE 4 thinking and targets the Gnome, MATE, and Cinnamon equivalents as well as other handlers.

If I monitor the destination with `gdbus monitor -e -d org.freedesktop.ScreenSaver` and call `xdg-screensaver susped wid` nothing shows up.

> > Inhibit from GtkApplication does absolutely nothing for me.
> 
> Awesome... Could be the same reason as for `xdg-screensaver suspend` or the system needs to implement some handler. shrug

I don't know. I'll test the hacky script on the debian 11 vm and see if it actually works.

infirit:
I'm really tempted to just open a PR that removes it. If someone want to resurrect it with a method that works on various desktops they can open a PR.

cschramm:
I still don't get why it doesn't. If you look at https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-screensaver.in, you see that it should do exactly what you're doing here in its `screensaver_freedesktop` function. If your change works but `xdg-screensaver` does not, either we fail to call it correctly or its broken on your system. Could it be that it simply does not get `KDE_SESSION_VERSION` so that it tries its KDE 3 handling?

infirit:
I got a clue `$ gdbus call --session -d org.freedesktop.ScreenSaver -o /org/freedesktop/ScreenSaver --method org.freedesktop.ScreenSaver.Inhibit "test" "test"` also doesn't work while in blueman it does. It looks like the calling process has to remain running for the Inhibit to persist. `xdg-screensaver` for sure doesn't.

infirit:
It doesn't support xfc4's screensaver, wip  https://gitlab.freedesktop.org/xdg/xdg-utils/-/merge_requests/38

Ah, and I was right about the process needing to be running, here is why it doesn't work on plasma https://gitlab.freedesktop.org/xdg/xdg-utils/-/issues/114 (only 5 years old).

All these are old af so I don't think we should rely on it anymore. I rather just read `XDG_CURRENT_DESKTOP` as I doubt we'll see any fixes from the xdg folks anytime soon.

cschramm:
Oh wow, now that makes a lot of sense... :unamused:

So, ideally we'd have a library that we can use from within our process to make the KDE Inhibit calls work properly and with Xfce support and probably a lot more.

I don't want to keep you from building that monster, but I'm not sure if this plugin is such an important feature. To be honest I don't fully understand the use case at all. A screensaver / locker should not jump in during gaming sessions anyway, at least due to regular input, no?

infirit:
> A screensaver / locker should not jump in during gaming sessions anyway, at least due to regular input, no?

Controller and joysticks are not considered regular input and won't reset the timer.

> I don't want to keep you from building that monster, but I'm not sure if this plugin is such an important feature.

Nope, now that I start to think about it'll never end. So my suggestion is to drop the plugin for now. And I'll update #1896 restoring the plugin and if it ever starts working it can be merged rebased and merged.

I really thought, cool, standard freedesktop interface, this should be simple :sob: .

cschramm:
Thanks für the explanation. :+1: I thought controllers usually do show up as input devices.

I'm fine with the removal, just want to add: If I'm not mistaken, the current implementation does not work with KDE 4's screensaver and Xfce's but it does seem to work for others, including MATE's, Cinnamon's, xscreensaver, xautolock, and anything that respects the X screensaver options set with `xset`.

PR #1901, opened 2022-10-17
NetConf: Provide DNS again

infirit:
Let's just do this for now and get this fixed quickly for debian 11.

### DIFF
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -91,7 +91,7 @@ class DnsMasqHandler:
             assert self.netconf.ip4_address is not None
 
             ipiface = ipaddress.ip_interface('/'.join((self.netconf.ip4_address, '255.255.255.0')))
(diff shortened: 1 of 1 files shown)


cschramm:
Why not #1565? 😄

We can't fix anything for Debian 11. It does not get updates like this.

cschramm:
Debian and Ubuntu releases do not get any updates except security patches.

infirit:
kinda sad that it's now broken on debian 11. Is backports still a thing? If so are you planning on having blueman in there instead?

cschramm:
Sure, there is https://backports.debian.org/. I can upload 2.4 there when it's ready.

I stopped caring, to be honest. There are so many systems out there that basically ship a snapshot and never get any updates. Versions with DNS disabled for dnsmasq are in Debian 11, three currently supported Ubuntu releases (two LTS), with a fourth one in two days and three supported Fedora releases, to just name the most important cases.

PR #1908, opened 2022-10-20
Make AnyBase garbage-collectable

cschramm:
Its DBusConnection held references to it but at least for AnyNetwork some plugin unload methods even expect it to get destroyed properly.

The `close` methods seems completely unused - or is it some kind of GLib magic?

### DIFF
--- a/blueman/bluez/AnyBase.py
+++ b/blueman/bluez/AnyBase.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional
+import weakref
 
 from gi.repository import GObject, GLib
 from gi.repository import Gio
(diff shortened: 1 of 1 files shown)


infirit:
> Its DBusConnection held references to it but at least for AnyNetwork some plugin unload methods even expect it to get destroyed properly.

They should call close. 

> The `close` methods seems completely unused - or is it some kind of GLib magic?

No magic. The close method is meant to be called when it's not needed anymore.

I'm fine with using weakref instead but in the future I'd like to hande property changes differently. Not having any objects is one thing that i would like to get rid of. 

I'll do some testing over the weekend.

cschramm:
>> some plugin unload methods even expect it to get destroyed properly.
> 
> They should call close.

Fair enough, but it's currently unused. Should work as well, yes.

cschramm:
I tried it and can confirm that adding `self._any_network.close()` before `del self._any_network` makes it work. `del` is still necessary there as the plugin instances themselves do not get destroyed, but that's a different story.

I think the implicit handling gives a nicer interface as you can do things like
```
inst = Object()
weakref.finalize(
    self,
    inst.disconnect_signal,
    inst.connect_signal("property-changed", handler)
)
# don't care about inst anymore
```

Sure, you cannot control unsubscribing the signal precisely, just via GC, but I do not really see an actual use case for that anyway.

Regarding plugin instances: https://github.com/blueman-project/blueman/commit/5a0e13052f5123655cc1d26d804195c22fd71b9d mentions that problem, but I don't even see how the `weakref` part can lead to the debug message and the `on_delete` mechanism. I don't think `weakref.finalize(self, x)` where `x` holds a strong reference to `self` can work as that reference prevents `self` from being destroyed at all.

PR #1911, opened 2022-10-21
List devices and battery levels in status icon tooltip

cschramm:
Issues: `BatteryWatcher`'s users expect it to get destroyed and for that `BatteryWatcher` expect the `Manager` and `Battery` objects to get destroyed. The latter requires #1908 to actually happen and for `Manager` it does not work due to the `self`-references it passes to its signal handlers to be able to call its `emit` method. I'm not sure if and how that could get resolved.

Apart from that I'm not sure how good the idea of using that Unicode symbol actually is... :sweat_smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Enable dnsmasq DNS service if possible and add DNS servers otherwise
 * Reconfigure DHCP service on local nameserver changes
 * Support for systemd-resolved for getting nameservers for NAP clients
+* List connected devices in status icon tooltip
 
 ### Changes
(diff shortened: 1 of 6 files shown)


infirit (approved):
Works ok. Let's see if the unicode battery character will cause issues :-).
image

cschramm:
I'm not sure why but it seems like `BatteryWatcher` instances can now get properly destroyed. :thinking:

kuraga:
Show the active profile, maybe?

kuraga:
@cschramm , show the active profile, maybe?

PR #1913, opened 2022-10-22
Strip surrounding whitespace from device names

cschramm:
I encountered an interesting device that has a trailing CR on its Name and default Alias, causing weird looking empty lines at least in our device list and the status icon menu items. Of course there could be more such garbage (like control characters somewhere in the middle of a name), but stripping surrounding whitespace in all places where an Alias shows up in UI seems like a low hanging fruit here.

shivr

### DIFF
--- a/blueman/bluez/Device.py
+++ b/blueman/bluez/Device.py
@@ -32,6 +32,11 @@ class Device(Base):
     ) -> None:
         self._call('Disconnect', reply_handler=reply_handler, error_handler=error_handler)
 
+    @property
+    def display_name(self) -> str:
+        alias: str = self["Alias"]
+        return alias.strip()
+
 
 class AnyDevice(AnyBase):
     def __init__(self) -> None:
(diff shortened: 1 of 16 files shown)


infirit:
Does it make sense to have this handled by the device class? Now we have to remember to strip on anything we do with the alias

cschramm:
The idea behind doing this everywhere where the value gets displayed is that you might still want to get the actual value without the cleanup if your purpose is not just to show the name to the user.

We actually have a case (that I failed to exclude from the cleanup :sweat_smile:): Consider you want to remove the whitespace from the alias, then blueman has to show it in the rename dialog, so https://github.com/blueman-project/blueman/blob/f4f747e8b5d6fdaad6c0640f8745b1471ccb5e18/blueman/gui/manager/ManagerDeviceMenu.py#L404 actually should not strip.

I'm undecided if that's worth scattering the cleanup everywhere. :shrug: It's not super important in many cases, e.g. it does not really matter in log messages. The most relevant places are those in my screenshot and maybe notifications and that's probably it.

infirit:
How about we make the intention explicit and have a dedicated method for displaying purposes? Something like below perhaps...

Also curious how gnome bluetooth/bluedevil handles this :smile: .

```diff
diff --git a/blueman/bluez/Device.py b/blueman/bluez/Device.py
index 9d87bdc5..0d30a568 100644
--- a/blueman/bluez/Device.py
+++ b/blueman/bluez/Device.py
@@ -32,6 +32,9 @@ class Device(Base):
     ) -> None:
         self._call('Disconnect', reply_handler=reply_handler, error_handler=error_handler)
 
+    def display_name(self) -> str:
+        return self.get("Alias").strip()
+
 
 class AnyDevice(AnyBase):
     def __init__(self) -> None:
```

infirit (approved):
lgtm

PR #1915, opened 2022-10-26
Make compatible with Python 3.11

t-puetz:
I installed Python 3.11 and compiled your project with Cython 3.0.0a11 since Cython moved a frame header file to an internal directory and blueman could not be compiled with an old Cython version anymore. Afterwards I noticed that when trying to connect to a device, that Python  3.11 will not accept accessing class members via self anymore, but it has to be self.\_\_class\_\_

### DIFF
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -21,8 +21,10 @@ from blueman.Sdp import (
     HID_SVCLASS_ID)
 
 import gi
+
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
+
 gi.require_version("Gdk", "3.0")
 from gi.repository import Gdk
 from gi.repository import GLib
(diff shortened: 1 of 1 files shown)


cschramm:
I cannot find anything about such a change and it would clearly require more changes in our code. What issue did you actually run into? :thinking:

t-puetz:
Hi,

I am not a very experienced dev, so I apologize if this PR is nonsense.

When connecting to a device in blueman-manager I got the following error and that let me to create this patch.

  File "/usr/lib/python3.11/site-packages/blueman/gui/manager/ManagerDeviceList.py", line 243, in on_event_clicked
    self.menu.connect_service(row["device"])
  File "/usr/lib/python3.11/site-packages/blueman/gui/manager/ManagerDeviceMenu.py", line 144, in connect_service
    self.set_op(device, _("Connecting…"))
  File "/usr/lib/python3.11/site-packages/blueman/gui/manager/ManagerDeviceMenu.py", line 105, in set_op
    if inst.SelectedDevice == self.SelectedDevice and not (inst.is_popup and not inst.props.visible):
       ^^^^^^^^^^^^^^^^^^^
AttributeError: 'ManagerDeviceMenu' object has no attribute 'SelectedDevice'

PR #1918, opened 2022-10-31
ManagerDeviceLost: Fix right click device menu

infirit:


cschramm:
I think it's bad enough for a 2.3.5, yes. We should also merge #1807. I do not have any candidates from the main branch.

PR #1927, opened 2022-11-12
Use GAction where possible

infirit:
Eventually we will need to drop the current menu's for something else. In preparation I wanted to move all "regular" menuitem callbacks to GAction as pretty much all replacements will use them . Sadly RadioMenuItems don't work with GAction :disappointed:.

I think it would be good to do as much as possible so marked it for review. Also included accelerators for closing windows and apps.


infirit:
I have reworked this a bit and added wip ManagerDeviceList actions. I am thinking now how to handle plugins...

PR #1935, opened 2022-11-19
Small fixes

infirit:
gettext in meson now behaves the same as under autotools.

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -1,7 +1,6 @@
 #!@PYTHON@
 import os
 import sys
-import signal
 import logging
 import gettext
 
--- a/apps/blueman-applet.in
+++ b/apps/blueman-applet.in
@@ -1,7 +1,6 @@
 #!@PYTHON@
 import sys
 import os
-import signal
 import logging
 import gettext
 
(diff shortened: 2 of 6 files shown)


cschramm (approved):
No idea what's wrong with the mypy jobs but it's not related to these changes. I retried a successful job at https://github.com/blueman-project/blueman/actions/runs/3500156611/jobs/5871427226 and it fails as well. The python image is the same. I did not see the mypy version in the logs, but 0.991 had been available for a couple of days, so it should have been that in both cases. It's definitely not mypy 0.991 that causes it as it works fine for me (with Python 3.10).

infirit:
Same here, fine for me on 0.991. It's tripping over cairo.Context so perhaps there is something in pycairo :thinking: 


```
$ MYPYPATH=stubs python3 -m mypy -p blueman --strict
Success: no issues found in 130 source files
[sander@infirit blueman]$ mypy --version
mypy 0.991 (compiled: no)
```

infirit:
It is, https://pycairo.readthedocs.io/en/latest/changelog.html. Maybe for now drop installing pycairo with pip to avoid this until we get a proper fix. It also looks like pycairo provides typing so we could drop it from blueman stubs right?

cschramm (approved):
Ah, that makes sense. :+1:

PR #1937, opened 2022-11-20
Fix cairo.Context generic types

infirit:
This works but isn't backwards compatible with pycairo earlier pycairo which is in most distribution.

```
$ PYTHONUSERBASE=~/pycairodev/ MYPYPATH=stubs python3 -m mypy -p blueman --strict
Success: no issues found in 130 source files
[sander@infirit blueman]$ MYPYPATH=stubs python3 -m mypy -p blueman --strict
stubs/gi/repository/Gdk.pyi:12: error: "Context" expects no type arguments, but 1 given  [type-arg]
stubs/gi/repository/Gtk.pyi:14: error: "Context" expects no type arguments, but 1 given  [type-arg]
blueman/gui/GtkAnimation.py:18: error: "Context" expects no type arguments, but 1 given  [type-arg]
Found 3 errors in 3 files (checked 130 source files)
```


cschramm:
Not really a problem, is it? We're installing pycairo via pip anyway (for similar reasons; needed their stubs added in 1.16.3 not available in most distributions at that time).

infirit:
Yeah, it's not really. When I'm back I'll add a commit to force pycairo =>1.22.0

cschramm (approved):

PR #1962, opened 2022-12-06
Fix icons

infirit:
Todo

- ~~Add something similar for IconPixmap~~
- Handle errors in loading the icon
- Update scalefactor
- ~~Wayland has no primary monitor, find another way to get scale factor.~~
- Fix MATE (sort of fixed, it's slightly transparent :eyes: )
- ~~fix typing~~

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: sudo apt-get update
-      - run: sudo apt-get install -y -qq --no-install-recommends meson gettext cython libglib2.0-dev python3-dev python-gi-dev libbluetooth-dev
+      - run: sudo apt-get install -y -qq --no-install-recommends meson gettext cython3 libglib2.0-dev python3-dev python-gi-dev libbluetooth-dev
       - run: meson --warnlevel 3 --buildtype debug -Druntime_deps_check=false builddebug
       - run: ninja -v -C builddebug/
 
--- a/blueman/main/Applet.py
+++ b/blueman/main/Applet.py
@@ -21,7 +21,7 @@ from blueman.plugins.applet.StandardItems import StandardItems
 from blueman.plugins.applet.StatusIcon import StatusIcon
 
 
-class BluemanApplet(Gio.Application):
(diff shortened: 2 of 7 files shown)


cschramm:
Quickly tested with mate-panel:

* Does not fix the tray icon itself.
* Fixes custom icons for menu items, especially `blueman-send-symbolic`.
* Sets correct scale factor 1 but...
* ...shows icons twice the size as usual. Changing the size to 16 brings back the expected icon sizes.

infirit:
> That is really strange, Xfce and Plasma are fine. I'll add it to the list.

It wasn't me this time :rofl: . It's also broken for me with 2.3.5 from arch repos. It seems like IconName is not picked up by MATE panel.

cschramm:
Yes, I can confirm that. The icons only become available in the default icon theme after installing the package a second time. :see_no_evil:

cschramm:
> > Quickly tested with mate-panel:
> > ```
> > * Does not fix the tray icon itself.
> > ```
> 
> That is really strange, Xfce and Plasma are fine. I'll add it to the list.

The problem with the tray icon itself is that our `StatusNotifierItemService` provides the `IconName` (`blueman-tray`, `blueman-active`, or `blueman-disabled`) and obviously the server cannot find that icon if it is not installed. This PR currently fixes the same problem with icon names (now icon-data so that the server does not have to find anything) in `MenuService` but not in `StatusNotifierItemService`.

I'm surprised it works in Xfce and Plasma. Are they able to find the icons from the GResource? :thinking:

infirit:
> I'm surprised it works in Xfce and Plasma. Are they able to find the icons from the GResource? thinking

I doubt it. It somehow find a regular bluetoot/blueman icon. When I implement IconPixmap it should be fixed.

infirit:
> When I implement IconPixmap it should be fixed.

Done that and the right icons are showing. It's not quite right yet but getting there, slowly..

cschramm:
Not sure if you're seeing the same but some icons are somehow very transparent for me now.

Bildschirmfoto zu 2022-12-08 20-59-34

infirit:
Yes. It doesn't happen with "our" icons and also not with adwaita. I'm going to give up on this for now and revert the whole thing.

PR #1966, opened 2022-12-08
Revert the GResource stuff as it causes too much issues + couple small fixes

infirit:
I did include some fixes. Let me know if you want those in a separate PR


cschramm (approved):
Fine for me, but still renders a problem for #1951, right?

infirit:
I had it working with that so not a problem. Not as clean as the files in the gresource was most annoying and messy.

PR #1972, opened 2022-12-20
UI maintanance and a few fixes

infirit:
The bulk of this is just modernizing and updating deprecated properties. We target Gtk 3.24 now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Changes
 
 * Raise minimum Python version to 3.8
+* Raise minumum Gtk+ version to 3.24
 * Hide recent connections associated with unavailable adapters
 * Store network configuration in GSettings instead of /var/lib/blueman/network.state.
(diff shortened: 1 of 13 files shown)


cschramm:
> We target Gtk 3.24 now.

We need to reflect that at least in Dependencies.md and a changelog entry about it would probably make sense as well.

cschramm:
Well, GTK is not actually a build-time requirement. Not sure why it's in meson.build. If we add it, we either need to disable the check in the pipelines or make it available there.

infirit:
> Well, GTK is not actually a build-time requirement. Not sure why it's in meson.build.

Having thought about it, it doesn't make sense to check for this as most ditro's have the pkgconfig bits in the dev package. I dropped it from both meson and autoconf.

> > We target Gtk 3.24 now.
> 
> We need to reflect that at least in Dependencies.md and a changelog entry about it would probably make sense as well.

Add CHANGELOG entry and will update Dependencies when merged.

cschramm (approved):

Issue #1954, opened 2022-11-29
Unnamed pairable keyboards


cschramm:
Hiding devices was implemented under the assumption that unnamed devices are generally not pairable and that they get hidden on other systems like Android and Windows as well, see #1258.

We seem to encounter some cases where unnamed devices are pairable:

- #1824 - Goldtouch Keyboard
- #1898 - Logitech K480 Keyboard
- #1953 - Keyboard

I'm wondering if those devices show up (unnamed) on systems that we expect to hide them.

infirit:
All the unnamed devices I have around me are hidden by my android phone and windows. To give an example for 
61:E8:A8:73:D6:04 doesn't show up normally on my phone, only when I go into developer options and enable showing unnamed devices. I'm thinking to change the default for 2.4 as this clearly is confusing lot's of people.

I also don't have one of these devices. I'm willing to get one but paying over €100 for a goldtouch keyboard is too much for me.

![image](https://user-images.githubusercontent.com/3465730/204662079-06219889-85bc-48e8-bf60-d1187d30a838.png)
![Screenshot_20221129-232617_Instellingen](https://user-images.githubusercontent.com/3465730/204662130-6122b9ba-913b-4f9d-b19e-91d9eb2caf3a.png)

cschramm:
> I'm willing to get one but paying over €100 for a goldtouch keyboard is too much for me.

:smile: I hope we can gather some more data from affected users. The question remains how those devices (keyboards) work with other systems that hide unnamed devices. I see basically two possible answers:

- It is possible to get a name from the devices but BlueZ either does not know it or just not expose it on the API.
- Hiding on other systems depends on more properties than just the missing name.

cschramm:
In #1824 it actually looks like the name was missing due to some system issue. It might all just be a symptom of broken drivers or something.

infirit:
I found an open box `Logitech K480 Wireless ` which I ordered. I'll report back when I tried it.

infirit:
> I found an open box `Logitech K480 Wireless ` which I ordered. I'll report back when I tried it.

This is how this devices behaves. It is unnamed for a good while (minute+) and at some point BlueZ discovers the name. This also matches on my android phone. Interestingly Windows does show "keyboard" instead of nothing at all.

What is clear (to me), we should rethink how we handle discovery. 60 seconds is definitely too short for the name to show up.

cschramm:
BlueZ needs to be scanning to discover the name?

[...]

PR #1988, opened 2022-12-30
Increase device scan timeout

infirit:
10ish seconds is not enough for certain devices to be found.

### DIFF
--- a/blueman/gui/DeviceList.py
+++ b/blueman/gui/DeviceList.py
@@ -246,7 +246,7 @@ class DeviceList(GenericList):
         self.clear()
         self.manager.populate_devices()
 
-    def discover_devices(self, time: float = 10.24,
+    def discover_devices(self, time: float = 60.0,
                          error_handler: Optional[Callable[[BluezDBusException], None]] = None) -> None:
         if not self.discovering:
             self.__discovery_time = 0


cschramm (approved):
I expected it to be a minute already. :see_no_evil: 

Is this enough for https://github.com/blueman-project/blueman/issues/1954#issuecomment-1336153757?

I'm still considering to drop the timeout completely, https://github.com/blueman-project/blueman/issues/1954#issuecomment-1338941119, but raising it for now definitely does not hurt. :+1:

infirit:
> Is this enough for [#1954 (comment)](https://github.com/blueman-project/blueman/issues/1954#issuecomment-1336153757)?

It usually is ok. 

> I'm still considering to drop the timeout completely, [#1954 (comment)](https://github.com/blueman-project/blueman/issues/1954#issuecomment-1338941119)

Yeah, I have to try that out and see how things work. The thing is, I usually have blueman-manager always open (it starts/is restored with my session) and it would then in theory keep scanning, not sure this is ok.

cschramm:
> it would then in theory keep scanning

We might have a misunderstanding there: I'd still provide the stop button and I'd most probably not auto-start discovery, so the idea is that the user stops it either explicitly within blueman-manager *or* by closing it. Compared to other Bluetooth tools it seems just super unusual to have a scan timeout and it should not have any actually bad effects if the user just keeps scanning except for broken adapters / drivers that e.g. have issues when pairing while still scanning but that's even a problem for short-running scans.

PR #1989, opened 2022-12-30
Use F-Strings everywhere

infirit:
Also fix a bug where we never showed the progressbar text when transferring files with blueman-sendto


blueman/plugins/applet/DhcpClient.py
```diff
@@ -57,7 +57,7 @@ def err(_obj: Mechanism, result: GLib.Error, _user_data: None) -> None:
 
                 self.quering.remove(device)
 
-            Notification(_("Bluetooth Network"), _("Trying to obtain an IP address on %s\nPlease wait…" % device),
+            Notification(_("Bluetooth Network"), _(f"Trying to obtain an IP address on {device}\nPlease wait…"),
```
cschramm, line 60: https://github.com/blueman-project/blueman/pull/1306#issuecomment-665910288 still applies, doesn't it?

blueman/plugins/manager/PulseAudioProfile.py
```diff
@@ -83,7 +83,7 @@ def on_selection_changed(self, item: Gtk.CheckMenuItem, device: Device, profile:
 
             def on_result(res: int) -> None:
                 if not res:
-                    self.parent.infobar_update(_("Failed to change profile to %s" % profile))
+                    self.parent.infobar_update(_(f"Failed to change profile to {profile}"))
```
cschramm, line 86: Translated f-string.

blueman/plugins/services/Network.py
```diff
@@ -100,16 +100,16 @@ def ip_check(self) -> None:
...
                 entry.props.secondary_icon_name = "dialog-error"
                 entry.props.secondary_icon_tooltip_text = tooltip_text
                 raise ValueError(error_message)
 
             elif nap_ipiface.network == ipiface.network:
                 tooltip_text = _(
-                    "IP address overlaps with subnet of interface %s, which has the following configuration  %s/%s\n"
-                    "This may cause incorrect network behavior" % (iface, ipiface.ip, ipiface.netmask))
+                    f"IP address overlaps with subnet of interface {iface}, which has the following configuration"
+                    f"{ipiface}/{ipiface.netmask}\n This may cause incorrect network behavior")
```
cschramm, line 112: Translated f-strings.

blueman/plugins/services/Network.py
```diff
@@ -100,16 +100,16 @@ def ip_check(self) -> None:
 
         for iface, ipiface in self.interfaces:
             if nap_ipiface.ip == ipiface.ip:
-                error_message = _("IP address conflicts with interface %s which has the same address" % iface)
+                error_message = _(f"IP address conflicts with interface {iface} which has the same address")
```
cschramm, line 103: Translated f-string.

cschramm (changes requested):

infirit:
Forgot about the fstring issue with gettext 😞 . I do know `"{_('string'}"` support was added in 3.10 but we target 3.8 right now. I'll revert those translated strings.

cschramm:
> I do know `"{_('string'}"` support was added in 3.10

Oh, I'd expect that to work anyway. I guess you're referring to https://bugs.python.org/issue36310. That's only for pygettext but we use xgettext for extracting the string at least with autotools. It looks like xgettext does not support it at all: https://savannah.gnu.org/bugs/?61596 :disappointed:

cschramm:
There are still a couple of `f"...{_(...)}..."` now. Unless I got something wrong xgettext does not extract those (neither would pygettext < 3.10).

cschramm (approved):

PR #1992, opened 2022-12-31
Bump min Gtk+ version to 3.24

infirit:


infirit:
Wait, I didn't know this file is actually in the repo :sweat:, I could've made it part of the PR. I always thought it was like the wiki :smile:

cschramm (approved):
:smile:

PR #1993, opened 2022-12-31
blueman-manager: Handle device removal events gracefully

cschramm:
As seen in https://bugzilla.redhat.com/show_bug.cgi?id=2148624, it looks like at least our fader can cause the device removal event to get handled multiple times, so that the tree_iter might actually be gone and the assertion causes an exception. If there is no tree_iter, there is nothing left that we could or should do, so we can just return.


cschramm:
Paths that I can easily think of to trigger this: Call `clear` multiple times or call `clear` and let a device removal signal arrive from D-Bus. I think this leads to multiple faders for the same device and the first one will clear the data that the later one will look for.

infirit (approved):

PR #1994, opened 2022-12-31
More for unnamed devices

infirit:


blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -141,8 +141,12 @@ def search_func(self, model: Gtk.TreeModel, column: int, key: str, tree_iter: Gt
...
 
     def filter_func(self, _model: Gtk.TreeModel, tree_iter: Gtk.TreeIter, _data: Any) -> bool:
-        no_name = self.get(tree_iter, "no_name")["no_name"]
-        if no_name and self.Config["hide-unnamed"]:
+        row = self.get(tree_iter, "no_name", "device")
+        no_name = row["no_name"]
+        device = row["device"]
+        klass = get_minor_class(device["Class"]) if device is not None else None
+
+        if no_name and self.Config["hide-unnamed"] and klass not in (_("Keyboard"), _("Pointing"), _("Combo")):
```
cschramm, line 149: That's a weird check. It's basically just the Peripheral major class. I'm not sure, though, if we should actually include pointing devices or just target keyboards (and possibly combos) specifically.

blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -315,6 +315,13 @@ def make_caption(name: str, klass: str, address: str) -> str:
         return "<span size='x-large'>%(0)s</span>\n<span size='small'>%(1)s</span>\n<i>%(2)s</i>" \
                % {"0": html.escape(name), "1": klass, "2": address}
 
+    @staticmethod
+    def make_display_name(alias: str, klass: int, address: str) -> str:
+        if alias.replace("-", ":") == address:
+            return _("Unnamed device")
```
cschramm, line 321: That means the same name for all unnamed devices, causing issues at least with "hide unnamed devices" disabled, but I guess there's more in your head as you're currently not using `klass`. :smile:

cschramm (commented):
Happy new year! :clinking_glasses:

blueman/gui/manager/ManagerDeviceList.py
infirit: I also experimented with naming it based on the device class description, but then I thought why as it's already shown. It can be any name, how about
 ```python
f"{alias} (Unnamed device)"
```

> causing issues at least with "hide unnamed devices" disabled

What issues specifically are you thinking of?

cschramm: Never mind, I forgot that we're showing the actual address as well, so it's not a mess of unnamed devices but you still see the addresses in any case.

blueman/gui/manager/ManagerDeviceList.py
infirit: dropped pointing from the list

infirit:
I think this is ready now.
[nohideunnamedkeyb.webm](https://user-images.githubusercontent.com/3465730/224553398-98e83cb2-bb99-4d60-8a8c-50c00de3ab68.webm)

cschramm (approved):
Deserves a changelog entry

infirit:
Ah, yeah changelog, i keep forgetting 👍.

PR #1995, opened 2022-12-31
Always have a row selected

infirit:


blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -100,7 +100,7 @@ def set_op(self, device: Device, message: str) -> None:
         ManagerDeviceMenu.__ops__[device.get_object_path()] = message
         for inst in ManagerDeviceMenu.__instances__:
             if not hasattr(inst, "SelectedDevice"):
```
cschramm, line 102: I think this condition should be added to the next `if` with `or` to a) show the log message and b) avoid generating twice.

cschramm (changes requested):

infirit:
Finally got to the core of the problem, not having a row selected. As long as this is true everything else just works ™. Do you want me to open a PR with just the revert and the fix?

cschramm (approved):
Looks good like it is. :+1:

PR #1997, opened 2023-01-02
Properly activate Manager

infirit:
Also included the 2 commits from #1865


cschramm (dismissed):
As mentioned in #1865 I do not think that `--gapplication-service` has an actual effect for us.

cschramm (approved):

PR #2009, opened 2023-01-15
bluez.Base: Don't subclass from Gio.DBusProxy

infirit:
So we can get to the invalidated properties.

Base will now add the list of invalidated properties like this.
```python
Base:67 __properties_changed: /org/bluez/hci0/dev_76_DD_19_54_A8_42 {} ['TxPower', 'RSSI'] <Device.Device object at 0x7f777e155e80 (blueman+bluez+Device+Device at 0x55ef15d47da0)>
```

### DIFF
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -26,12 +26,10 @@ class BaseMeta(GObjectMeta):
         return instance
 
 
-class Base(Gio.DBusProxy, metaclass=BaseMeta):
-    connect_signal = GObject.GObject.connect
-    disconnect_signal = GObject.GObject.disconnect
-
+class Base(GObject.Object, metaclass=BaseMeta):
     __name = 'org.bluez'
     __bus_type = Gio.BusType.SYSTEM
+    __proxy: Gio.DBusProxy
 
     __gsignals__: GSignals = {
         'property-changed': (GObject.SignalFlags.NO_HOOKS, None, (str, object, str))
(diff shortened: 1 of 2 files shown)


infirit:
Wait, I just realized I did property-changed in two ways. It should be at least consistent between them.

I am leaning to `key, None` so we can check the value and if it's None then it's invalidated. But sending the list is also fine with me.

cschramm:
Is it even worth knowing that a property got invalidated? Can't we just provide `None` for it instead?

cschramm:
> I am leaning to `key, None` so we can check the value and if it's None then it's invalidated.

That would make perfect sense to me. 👍

infirit:
Updated and also slightly changed how it emit's the signal.

blueman/bluez/Base.py
```diff
@@ -131,6 +142,10 @@ def get_properties(self) -> Dict[str, Any]:
 
         return props
 
+    def destroy(self) -> None:
```
cschramm, line 145: It's "on purpose" that this does not get used (yet), right?

cschramm (approved):

blueman/bluez/Base.py
infirit: It's used here :-) https://github.com/blueman-project/blueman/blob/ac6d54668c4d5d13e510c5f4cc28f7dd643af25b/blueman/gui/manager/ManagerDeviceList.py#L134

cschramm: Ah, sure, this is an override. :facepalm:

PR #2018, opened 2023-01-26
Handle rfcomm_list failing

infirit:
Just quick draft so we can discuss if we need to display something to the user or not. And logging the exception may be too much, perhaps just the error message with or without the error number?

### DIFF
--- a/blueman/plugins/applet/SerialManager.py
+++ b/blueman/plugins/applet/SerialManager.py
@@ -6,7 +6,7 @@ from blueman.gui.Notification import Notification
 from blueman.Sdp import SERIAL_PORT_SVCLASS_ID
 from blueman.plugins.applet.DBusService import RFCOMMConnectedListener
 from blueman.services.Functions import get_services
-from _blueman import rfcomm_list
+from _blueman import rfcomm_list, RFCOMMError
 from subprocess import Popen
 import logging
 import os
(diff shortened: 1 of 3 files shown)


cschramm:
I would not display anything and not put much thought in the log message either. The more important question to me is if this is enough to actually make things work on RFCOMM-disabled systems.

By the way, I don't think the `rfcomm_list` call in `on_device_disconnect` can get reached on an RFCOMM-disabled system anyway, but the error handling does not hurt, of course.

infirit:
It's a kernel module on arch so I should be able to easily test this.

> By the way, I don't think the `rfcomm_list` call in `on_device_disconnect` can get reached on an RFCOMM-disabled system anyway, but the error handling does not hurt, of course.

Yeah, hadn't put much thought into it and just grepped for the command :-)

infirit:
Yeah, this is enough to handle rfcomm missing. Pushed update with 

```
blueman-manager 20.28.49 DEBUG    ManagerDeviceMenu:292 generate  : Mi Bluetooth Speaker
blueman-manager 20.28.49 ERROR    SerialService:38 connected_instances: rfcomm_list failed with: ERR_SOCKET_FAILED
```

infirit:
eeeuh, this just showed up in the infobar when I tried to connect serial port and made the window 2million pixels wide :rofl: . It's probably a good idea to hide it :-)

```
Verbinding mislukt: File "/usr/lib/python3.10/site-packages/blueman/services/meta/SerialService.py", line 89, in connect
    port_id = create_rfcomm_device(Adapter(obj_path=self.device["Adapter"])['Address'], self.device["Address"],
  File "_blueman.pyx", line 172, in _blueman.create_rfcomm_device
_blueman.RFCOMMError: ERR_SOCKET_FAILED
```

cschramm (approved):

PR #2022, opened 2023-02-02
Move widget creation to UI file and give stats enough space to grow and shrink

infirit:
Before it would jump back and forth depending on the rate. It is still possible for it to grow when enough data is transferred but no jumpy stats every couple of seconds.
image

edit: gave it a bit more space, hopefully this is enough for it to never jump even when a lot of data has transferred.


cschramm (dismissed):

infirit:
I noticed I had the placeholder for the rate tooltips marked as translatable which is wrong. While at it I also moved most of the menu.

cschramm (approved):

PR #2035, opened 2023-02-26
Manager: Rework list sorting

infirit:
Our list got very confusing with newly found device rows being inserted in the list based on name or datetime added. When you have lot's of devices it becomes a pain to find them so we need a better way.

We primarily now sort whether a device is paired. The logic behind this is that these are the devices the user most likely is interested in. Next we sort all connected devices below and the remaining ones at the bottom.

image


cschramm:
One could argue if the most important group should be at the top or at the bottom, but I guess top is best as you might need to scroll to it otherwise.

Also adding these groups is not really a reason to remove the options and always sort by name in ascending order within the groups. Should not hurt, though, I don't think the sort options are a commonly used feature and anybody will miss them (might be completely wrong :laughing:).

One thing that bugs me, though, but I'm not sure what to do with that, is that an item moves now when getting paired or connected. :thinking:

infirit:
> One could argue if the most important group should be at the top or at the bottom, but I guess top is best as you might need to scroll to it otherwise.

I can make ascending/descendinng an option again.

> Also adding these groups is not really a reason to remove the options and always sort by name in ascending order within the groups. Should not hurt, though, I don't think the sort options are a commonly used feature and anybody will miss them (might be completely wrong 😆).

Sure, I can expand and readd the options. A magic default sort, always alphabetical and always by added.

> 
> One thing that bugs me, though, but I'm not sure what to do with that, is that an item moves now when getting (un)paired or (dis)connected. 🤔

I can make it a timed sort, let's say an initial sort and then every couple minutes? It would still move at some point but not immediately anymore. Would that help?

infirit:
Added back in the existing sort options. Will poke at the delayed sort idea I have next.

PR #2053, opened 2023-03-26
Make search button a toggle

infirit:
Still need to clean it up and drop the stop from the progressbar but putting it here for some feedback.
edit: updated screenhot and hid the cancel button

image


cschramm:
I like it. :+1: Brings us a little closer to also drop the timeout mechanism, but it would need some animation other than the progress bar, I guess. That stop icon's border looks like it could be spinning. :sweat_smile:

infirit:
How about a spinner like this.
[spinner.webm](https://user-images.githubusercontent.com/3465730/232230559-63f3e3f2-f52a-4eca-8d74-dbbb24cab0b2.webm)

PR #2074, opened 2023-05-08
Meson paths cleanup and install polkit rules

infirit:


meson.build
```diff
@@ -21,11 +21,11 @@ endif
...
-pkgdatadir = join_paths([prefix, get_option('datadir'), package_name])
-bindir = join_paths([prefix, get_option('bindir')])
-libexecdir = join_paths([prefix, get_option('libexecdir')])
-schemadir = join_paths(['share', 'glib-2.0', 'schemas'])
+datadir = get_option('datadir')
+pkgdatadir = join_paths([datadir, package_name])
+bindir = get_option('bindir')
+libexecdir = get_option('libexecdir')
+schemadir = join_paths([datadir, 'glib-2.0', 'schemas'])
 pythondir = pyinstall.get_install_dir()
```
eli-schwartz: At least some of these are being used to write out a configure_file.

"prefix is already used for relative paths" is certainly true for meson's install_dir handling, but e.g. I would not expect gtk IconTheme.prepend_search_path to do so when getting a relative path from Constants.py

cschramm (approved):

Issue #2101, opened 2023-08-02
blueman-manager crashed with AttributeError in make_device_icon(): 'NoneType' object has no attribute 'load_surface'


tkrah:
blueman: 2.3.5+mint1+vera
BlueZ: 5.64-0ubuntu1
Distribution: Mint 21.1
Desktop environment: Cinnamon

* [ X] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Using blueman-manager it is crashing regularly with apport telling me this as a title:

`blueman-manager crashed with AttributeError in make_device_icon(): 'NoneType' object has no attribute 'load_surface'`

I have no idea what's wrong here, anyone got an idea how to debug / fix that problem?

cschramm:
Please provide the traceback (or at least the location of the error).

From looking at the code I think that could be possible if you lack a device icon. Can you start blueman-manager again right away? I'd expect it to crash again as the same device should still be in the list.

tkrah:
Does this help:

```
PythonArgs: ['/usr/bin/blueman-manager']                              
Traceback:                                                            
Traceback (most recent call last):                                   
   File "/usr/lib/python3/dist-packages/blueman/gui/manager/ManagerDeviceList.py", line 648, in _set_cell_data
      surface = self.make_device_icon(row["icon_info"], row["paired"], row["trusted"], row["blocked"])
    File "/usr/lib/python3/dist-packages/blueman/gui/manager/ManagerDeviceList.py", line 265, in make_device_icon
      target = icon_info.load_surface(window)                          
  AttributeError: 'NoneType' object has no attribute 'load_surface'
```

I can restart it right away, it does not always crash with this, but it happens here and there, sometimes 2 times a day, sometimes its crashing more than 10 times a day.

cschramm:
Perfect, thanks.

So, `row["icon_info"]` seems to be `None`. Assuming that it gets explicitly set to that, let's try to catch it with an assertion. Can you please apply the following patch to `/usr/lib/python3/dist-packages/blueman/gui/manager/ManagerDeviceList.py` and see where that gets us:

```diff
--- /usr/lib/python3/dist-packages/blueman/gui/manager/ManagerDeviceList.py
+++ /usr/lib/python3/dist-packages/blueman/gui/manager/ManagerDeviceList.py
@@ -360,6 +360,7 @@ class ManagerDeviceList(DeviceList):
[...]

PR #2103, opened 2023-08-03
Bunch of cleanups

infirit:
The last commit addresses #2101. I understand it's very unlikely to fail but I prefer blueman-manager to not crash to desktop. Many exceptions in the our lists are not fatal but this one is.

edit: flake had some suggestions

### DIFF
--- a/blueman/gui/GenericList.py
+++ b/blueman/gui/GenericList.py
@@ -1,4 +1,4 @@
-from typing import Dict, Optional, TYPE_CHECKING, Iterable, Mapping, Callable, Tuple, Union, Collection, List, Any
+from typing import Dict, Optional, TYPE_CHECKING, Iterable, Mapping, Callable, Tuple, Union, Collection, Any
 
 import gi
 gi.require_version("Gtk", "3.0")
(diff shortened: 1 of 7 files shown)


blueman/gui/GenericList.py
```diff
@@ -101,54 +94,32 @@ def prepend(self, **columns: object) -> Gtk.TreeIter:
...
-
-            if matches == len(cols):
-                ret.append(i)
-
-            matches = 0
+        for tree_row in self.liststore:
+            row_data = self.get(tree_row.iter)
+            for key in cols:
+                if row_data[key] == cols[key]:
+                    ret.append(tree_row.iter)
```
cschramm, line 103: This seems wrong to me.

We used to append a single entry for the row if `all(row_data[k] == v for (k, v) in cols.items())` (one for each row if all columns match). Now we're appending the row multiple times for each matching column. So basically this changes from logical AND to logical OR and duplicates results.

It's not actually a problem as the only use case seems to be with only the `name` key and the result is the same in case of a single column. Plus, the use case only uses the first result, so there's also no point in collecting at all.

We should probably just strip it down to something like this directly in `PluginDialog`:

```python
for tree_row in self.list.liststore:
    if self.list.get(tree_row.iter)["name"] == name:
        self.list.set(tree_row.iter, active=loaded)
        break
```

cschramm (changes requested):

cschramm (approved):

PR #2104, opened 2023-08-03
Wrap cairo.Surface in a GObject

infirit:
Ignore the other commit (it's bed time so I left it in).

How about we handle the surface like this? When I added the scaling I wasn't smart enough to figure out how to get the surface in the store. It's not perfect but it works well enough and simplifies things.


cschramm:
Sure, why not? :+1:

cschramm (approved):

PR #2105, opened 2023-08-04
Allow blueman-applet to be controller by systemctl --user

infirit:
Not entirely sure how this van be enabled by default but for now just make it possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Changes
 
+* Allow blueman-applet to be controlled by systemd (systemctl --user). The autostart desktopfile still has priority over this and will have to be removed.
 * Raise minimum Python version to 3.8
 * Raise minumum Gtk+ version to 3.24
 * Hide recent connections associated with unavailable adapters
--- a/data/configs/blueman-applet.service.in
+++ b/data/configs/blueman-applet.service.in
@@ -1,7 +1,11 @@
 [Unit]
 Description=Bluetooth management applet
+PartOf=graphical-session.target
 
 [Service]
 Type=dbus
 BusName=org.blueman.Applet
 ExecStart=@BINDIR@/blueman-applet
+
+[Install]
+WantedBy=default.target


cschramm:
To my understanding that is completely up to distributions, e.g. Debian packages typically enable units on installation while on Arch things need to get enabled by the user after installation.

infirit:
Ok, let me add a changelog entry and then it can be reviewed.

cschramm:
Did you actually get that running?

For me it tries to start after a system reboot but fails loading the icon scheme. The problem looks exactly like the problem folks have with an improperly configured systemd D-Bus activation environment with no `DISPLAY` set.

When I just re-login, even after restarting my desktop manager, I do not see a start attempt at all, although I'd expect one.

infirit:
I could start and stop it from an existing session. I haven't tried a fresh boot, let me do that and get back to you.

infirit:
Can you try adding `ExecStartPre=sleep 3` under the unit section?

I'm at work rn so can't test this on xfce but will do later today.

infirit:
Gonna give up on this. Tried several things but it refuses to start at XFCE session start. Once it's started it's fine.

infirit:
Gonna give up on this. Tried several things but it refuses to start at XFCE session start. Once it's started it's fine.

cschramm:
I'm currently mainly using Cinnamon, with lightdm.

Whatever I tried, systemd always tries to start the unit when I first log in, no matter if I'm in a Desktop session or not. When lightdm is enabled, the unit already gets started when lightdm gets (i.e. before I actually log in). I have no idea why but somehow lightdm seems to open an initial user session for me right away. 🤷‍♂️

systemd does have an autostart mechanism for user applications but that needs to be requested by the desktop via the xdg-desktop-autostart.target. XDG autostart files can then be translated into systemd units with the systemd-xdg-autostart-generator tool.

In the end it seems like XDG autostart is the only universal system to autostart desktop applications.

PR #2111, opened 2023-08-08
Various little things

infirit:


blueman/main/applet/BluezAgent.py
```diff
@@ -94,21 +91,11 @@ def on_insert_text(editable: Gtk.Entry, new_text: str, _new_text_length: int, _p
...
-                    err: Callable[[Union[BluezErrorCanceled, BluezErrorRejected]], None]) -> None:
-        ...
-
-    @overload
-    def ask_passkey(self, dialog_msg: str, is_numeric: "Literal[False]", device_path: str, ok: Callable[[str], None],
-                    err: Callable[[Union[BluezErrorCanceled, BluezErrorRejected]], None]) -> None:
-        ...
-
-    def ask_passkey(self, dialog_msg: str, is_numeric: bool, device_path: str, ok: Callable[[Any], None],
+    def ask_passkey(self, dialog_msg: str, is_numeric: bool, object_path: str, ok: Callable[[Any], None],
```
cschramm, line 98: The overloads make sense as they restrict the arguments to either `is_numeric = true` and a `Callable[[int], None]` or `is_numeric = false` and a `Callable[[str], None]`. Without, any `Callable[[Any], None]` is fine independent of `is_numeric`.

blueman/main/Adapter.py
```diff
@@ -18,16 +18,14 @@
...
-    from typing_extensions import TypedDict
-
-    class Tab(TypedDict):
-        grid: Gtk.Grid
-        hidden_radio: Gtk.RadioButton
-        always_radio: Gtk.RadioButton
-        temparary_radio: Gtk.RadioButton
-        visible: bool
-        label: Gtk.Label
+class Tab(TypedDict):
```
cschramm, line 21: Makes sense. Out `TypedDict`s are from a time where we still supported Python < 3.8.

blueman/main/applet/BluezAgent.py
infirit: I completely missed the second part that is different. I'll back this out.

cschramm (approved):

PR #2123, opened 2023-08-20
blueman-services: Make it a notebook

infirit:
We really don't need a custom list for this. Also opens up binding settings to widgets.

image

image

edit: updated screenshots

### DIFF
--- a/blueman/main/Services.py
+++ b/blueman/main/Services.py
@@ -1,11 +1,10 @@
-from gettext import gettext as _
 import os
 import logging
 import importlib
 import signal
-from typing import List, Optional
+from typing import Optional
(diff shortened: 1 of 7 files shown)


infirit:
Anything else you like changed while I'm working on this? I am thinking of removing the "Services" label as it only NAP these days. Also it would look better if the left margin on the  check/radio buttons is reduced bit.

infirit:
Made the whole thing more compact. Added (translatable) tooltips to the dun/pan labels to avoid people explaining inside the label like the person who did the dutch translation (which I removed in weblate).

cschramm (approved):

PR #2128, opened 2023-08-25
Update PulseAudioProfile.py

localevil:
On my system, the active profile is not bold.
I suggest highlighting it with another icon.


infirit:
@cschramm on plasma it's also not bold, should we just drop it and use an icon like `dialog-ok`?

cschramm:
Perfectly fine for me. Markup support in `com.canonical.dbusmenu`s seems super rare. I actually don't have any recent example.

localevil:
@infirit I changed icon to `dialog-ok`.

infirit (approved):

PR #2132, opened 2023-09-03
Allow to toggle the Bluetooth status from the manager menu

### LINKED ISSUE #2044: Where is the actual Bluetooth on/off button?
blueman: 2.3.5
BlueZ:
Distribution: Linux Mint 21 Cinnamon
Desktop environment: Cinnamon

I've always used the Bluetooth button on my taskbar to turn Bluetooth on and off, as well as connect to devices. Since upgrading from Linux Mint 20 to 21, now I have Blueman there instead (not sure what it was before exactly):

image

Where is the actual on/off button to just turn off Bluetooth? It's nowhere to be found (including in System Settings, which just opens the same Blueman dialog.)

cschramm:
This also does not disable the whole "Adapter" menu, so that the exit item is now reachable when powerd off.

Closes #2044

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Support for systemd-resolved for getting nameservers for NAP clients
 * List connected devices in status icon tooltip
 * Support for nautilus 43 and later
+* Allow to toggle the Bluetooth status from the manager menu
 
 ### Changes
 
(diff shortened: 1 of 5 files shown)


infirit:
The menu entry doesn't update so initially I was little confused :laughing: . You'll want to update it in `on_adapter_property_changed` to reflect the power state change.

cschramm:
🤦 My Thinkpad adapter gets removed when rfkill blocked. Will add that.

infirit:
The conflict is easy to fix but for some reason most of CI failed and I see no option to restart it :eyes: .

infirit (approved):

PR #2137, opened 2023-09-10
Listbox for plugindialog

infirit:
I still see some weirdness with enabling/disabling plugins but I suspect I messed up my gsettings in the process. Putting it out here as a possible option to use for blueman-manager's device list.

The idea here is that we make our own widget in the factory function and then optionally bind the GObject property to widgets. Then updating the `GObject` property will automatically update the UI. The plugin dialog is not very dynamic and used stateful `SimpleAction` for the checkbox but binding the enabled property is also an option.

mypy was very unhappy so some of the typing annotation looks weird and I had to cast the `PluginItem` properties to convince it.

The other option is to use `ListBoxRows` and insert them manually into the `ListBox`.

### DIFF
--- a/blueman/gui/applet/PluginDialog.py
+++ b/blueman/gui/applet/PluginDialog.py
@@ -1,8 +1,7 @@
 from gettext import gettext as _
 import logging
-from typing import TYPE_CHECKING, List, Type, Dict
+from typing import TYPE_CHECKING, Type, Dict, cast, Optional
 
-from blueman.gui.GenericList import GenericList, ListDataDict
 from blueman.main.Builder import Builder
 from blueman.main.PluginManager import PluginManager
 from blueman.plugins.AppletPlugin import AppletPlugin
@@ -11,10 +10,37 @@ from blueman.plugins.BasePlugin import Option
 import gi
 gi.require_version("Gtk", "3.0")
 gi.require_version("Gdk", "3.0")
-from gi.repository import Gtk, Gdk, Gio
+from gi.repository import Gtk, Gdk, Gio, GLib, GObject
 
 if TYPE_CHECKING:
     from blueman.main.Applet import BluemanApplet
+
+
+class PluginItem(GObject.Object):
+    __gtype_name__ = "PluginItem"
+
+    class _Props:
+        icon_name: str
+        plugin_name: str
+        description: str
+        enabled: bool
+        activatable: bool
+
+    props: _Props
+
+    icon_name = GObject.Property(type=str)
+    plugin_name = GObject.Property(type=str)
+    description = GObject.Property(type=str)
+    enabled = GObject.Property(type=bool, default=False)
+    activatable = GObject.Property(type=bool, default=False)
+
+    def __init__(self, icon_name: str, plugin_name: str, description: str, enabled: bool, activatable: bool):
+        super().__init__()
+        self.props.icon_name = icon_name
+        self.props.plugin_name = plugin_name
+        self.props.description = description
+        self.props.enabled = enabled
+        self.props.activatable = activatable
 
 
 class SettingsWidget(Gtk.Box):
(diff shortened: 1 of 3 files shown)


infirit:
The weirdness was because I messed up gsettings. Cleaned it up and it's ready for review.

Hopefully based of this I can do the same for the ManagerDeviceList to make it much easier to implement a new design.

blueman/gui/applet/PluginDialog.py
```diff
@@ -159,45 +169,97 @@ def __init__(self, applet: "BluemanApplet") -> None:
...
-                          ) -> int:
-        a = self.list.get(iter1, "activatable", "name")
-        b = self.list.get(iter2, "activatable", "name")
+    def _add_plugin_action(self, name: str, state: bool, activatable: bool) -> None:
+        logging.debug(f"adding action: {name}")
+        action = Gio.SimpleAction.new_stateful(
+            name, None, GLib.Variant.new_boolean(False)
+        )
+        action.set_property("enabled", activatable)
+        action.set_state(GLib.Variant.new_boolean(state))
```
cschramm, line 183: Why not initialize with `state` right away?

blueman/gui/applet/PluginDialog.py
```diff
@@ -159,45 +169,97 @@ def __init__(self, applet: "BluemanApplet") -> None:
...
-            else:
-                return 0
+            return 0
 
-    def _on_close(self, _widget: Gtk.Widget, _event: Gdk.Event) -> bool:
-        self.applet.Plugins.disconnect(self.sig_a)
-        self.applet.Plugins.disconnect(self.sig_b)
-        return False
+    def _on_plugin_toggle(self, action: Gio.SimpleAction, state: GLib.Variant) -> None:
+        action.set_state(state)
```
cschramm, line 225: Not sure if this should not better be deferred until after the below branches, so that in case `_ask_unload` gets called and returns `False`-ish the state does not get changed here and then back there.

cschramm (approved):
Didn't test but the code lgtm.

blueman/gui/applet/PluginDialog.py
infirit: I ended up with widget and action out of sync in certain situations.

For example, let's say the user wants to enable the `PPPSupport` which conflicts with the `NMDUNSupport`. In the callback the user decided not to enable plugin. The action is  `False` and setting it to `False` again does not send a signal to the widget to update it's state as there is no change. It's most likely the reason I messed up my GSettings.

blueman/gui/applet/PluginDialog.py
infirit: Don't think there is a reason. I'll double check to be sure and update the commit.

infirit: Fixed.

blueman/gui/applet/PluginDialog.py
```diff
@@ -159,45 +169,96 @@ def __init__(self, applet: "BluemanApplet") -> None:
...
-            else:
-                return 1
-
+            return 1
         else:
-            if a["activatable"] and not b["activatable"]:
+            if item1.props.activatable and not item2.props.activatable:
                 return -1
-            elif not a["activatable"] and b["activatable"]:
+            if not item1.props.activatable and item2.props.activatable:
```
cschramm, line 219: Sonarcube complains that `not item1.props.activatable` is always `True` here. Makes some sense, but fine for me to leave the structure. :shrug:

cschramm (approved):

blueman/gui/applet/PluginDialog.py
infirit: I didn't change (or think about) the logic but I'll have a look to see if this can be handled better.

cschramm: I think it makes sense to put these two checks for different `activatable` values first, then the two for different `plugin_name`s and finally a single `return 1` and save what's currently in lines 209 and 210. There should not be any redundancy as caught by Sonarcube then.

infirit: Pushed update that works for me and it's happy now.

cschramm (approved):

PR #2144, opened 2023-09-28
Black autoformatter

infirit:
I was curious and the result is mostly quotes. The command used was `black -l 120 -t py38 -t py39 -t py310 -t py311 -t py312`.

It can also run as part of the CI. Is it an idea to just have black manage this?

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,13 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: psf/black@stable
+        with:
+          options: "--check --verbose --line-length 120 -t py38 -t py39 -t py310 -t py311 -t py312"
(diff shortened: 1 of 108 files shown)


cschramm:
Sure :+1:, just flake8 isn't happy with the result yet. :sweat_smile:

infirit:
> Sure 👍, just flake8 isn't happy with the result yet. 😅

black is "opinionated" and incompatible with flake8, it's a replacement. I added a new workflow for black and dropped flake8 from checks.

sonarcloud is incorrectly unhappy with hardcoded "IP" addresses. It does have some codesmells I sorta agree with but definitely not all. I'll follow up with another PR to address them.

cschramm:
Fine for me, but we should consider adding pylint then.

I think what's going on in SonarCloud is just that all or at least most of those things already existed and were set to ignored, so that they did not show up in the analysis and due to the (whitespace) changes they are now considered new. I just set the security hotspots to "safe", so that it should pass.

infirit:
I'll add pylint to run after black. Hopefully they'll be nice to each other 😰

PR #2146, opened 2023-10-02
Add services-window.ui to Makefile

cschramm:


infirit (approved):
How did this never cause issues for people?

cschramm:
I was surprised as well. :sweat_smile: Well, there are couple of preconditions: You have to build from git, use make, not meson, fetched an update in the last ~5 weeks since #2123 got merged, and try to use blueman-services (which existing users typically do not do). I would have guessed that a user of https://aur.archlinux.org/packages/blueman-git comes up with it, though. :man_shrugging:

Issue #1119, opened 2019-10-14
Wrong output "Failed to enable bluetooth" when enable bluetooth device


parr0tr1ver:
blueman: 2.1.1
BlueZ: 5.50
Distribution: WRLinux
Desktop environment: Xfce

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

I add a debug info in plugins/applet/PowerManager.py: get_bluetooth_status()
  def get_bluetooth_status(self):
        logging.warning("self.current_state: %s" % self.current_state)
        return self.current_state

Steps to reproduce:
1 Start blueman-applet with '--loglevel debug'
2 right click bluetooth-applet icon and disable bluetooth device
3 launch blueman-manager from terminal, and click 'Enable Bluetooth'. It shows error message but the bluetooth device enabled sooner.

sh-5.0# blueman-manager 
blueman-manager version 2.1.1 starting
blueman-manager 09.07.04 
Failed to enable bluetooth

And the applet outputs please see:
https://gist.github.com/parr0tr1ver/c07f9ce43ab19ee63c2d3e7098884a02

get_bluetooth_status() runs before "check     : callbacks done" and then output error message. But the final callback enables the bluetooth device.

Any suggestion is appreciate. Thanks.

parr0tr1ver:
Why the description are missing? :(

parr0tr1ver:
@cschramm what more infomation should I provide, please?

cschramm:
The description was missing. I fixed your markup and can replicate the behavior. Thanks for the report! :slightly_smiling_face:

parr0tr1ver:
@cschramm thanks. I have been reading the source code for a while but didn't get root cause. Do you have any thought how to fix it? Thanks. :)

infirit:
@parr0tr1ver it is a timing issue. Delaying ``if not applet.GetBluetoothStatus():`` in ``check_bluetooth_status`` by a second will solve this. Now, having sleep(1) is a bit ugly so it would probably be nicer to schedule callback and have it check the status every 0.2 seconds or so. But timeout after a few seconds..

If you want to have a go at fixing this look at [Glib.timeout_add](https://lazka.github.io/pgi-docs/#GLib-2.0/functions.html#GLib.timeout_add).

Ideally we also let blueman-manager handle things without exitfunc but this requires a bit more work in ``check_bluetooth_status`` and proper testing.

[...]

Issue #2161, opened 2023-10-09
Devices window closes when BT adapter is enabled


MartinGebert:
blueman: 2.3.5
BlueZ: 5.68
Distribution: Gentoo
Desktop environment: LXQt

* [X] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

This is a minor usability issue and for me a tiny but frequent nuisance, and I wonder if it can be solved:

I've configured Bluetooth to start with the adapter disabled; this also applies when it's waking up from hibernating. In case I want to use BT I usually left click the tray icon, and the devices window appears, offering to enable BT in a dialog. So far this is exactly what I want.

When I click the "Enable Bluetooth" button the devices window is closed immediately, while BT has been enabled as expected. This is my issue; I would like to see the available devices **without** having to open the devices window again. OTOH closing the devices window probably makes sense when **disabling** the adapter, and I assume that's the reason it's implemented this way.

This can be worked around by using the context menu of the tray icon (on my system right instead of left button click) to enable the adapter, then open the devices window, but I haven't managed yet to convince my brain to adapt this procedure.

Is there a configuration or other way to customize this behaviour? If not, would you consider it as a feature request, please?

infirit:
It's really a bug. For some reason `check_bluetooth_status` thinks enabling/powering failed. The error `Failed to enable bluetooth`

MartinGebert:
> For some reason check_bluetooth_status thinks enabling/powering failed.

If this really were an error it should IMHO show an appropriate message, not just close the window and leave the user clueless.

infirit:
Tldr, we need to fix this in blueman :-).

It's a timing bug in blueman manager. What happens is that we ask Bluetooth to be powered/enabled but don't wait long enough for it to finish. We then ask too early and are told Bluetooth is not powered/enabled and erroneously close the window.

Sleeping for half a second between powering and asking if it succeeded fixes it. But I like to take the opportunity to change how this is handled in blueman manager. There is actually no reason to close the window as long as we have the (modal) dialog over the main window and prevent the user from interacting with it.

infirit:
@MartinGebert if you're able feel free to test #2163.

MartinGebert:
[...]

Issue #1707, opened 2022-05-02
Device List doesn't get cleared on adapter turns off


kuraga:
blueman: 2.2.3
BlueZ: 5.64
Distribution: Calculate Linux (Gentoo)
Desktop environment: XFCE 4.16

* [x] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

If Bluetooth is disconnected, Bluetooth Devices window is closed after an appropriate dialog.

But if Bluetooth Devices window is opened and then Bluetooth disconnects, window is nor closed nor cleared. Some inconsistency...

Thanks.

cschramm:
Not really sure what you mean.

I guess it's about disabled Bluetooth, i.e. existing adapters are not powered. In that case, blueman-manager shows a hint and a way to power the adapter(s) on start. If the window is already open when adapters get unpowered, it does not close, but it does get cleared.

Is that what you mean? I guess it would be a bit harsh and does not add value if we'd just close it.

kuraga:
@cschramm , thanks for reply!

> blueman-manager shows a hint and a way to power the adapter(s) on start.

...and close the window.

> If the window is already open when adapters get unpowered, it does not close, but it does get cleared.

On my system, it doesn't get cleared.

P.S. It's not so important, just for knowledge.

cschramm:
That could be a bug then. blueman-manager should only show devices bound to powered adapters. Please provide the basic information we ask for in the issue template and the output of `blueman-manager --loglevel debug` of when you disable Bluetooth.

kuraga:
```
blueman-manager 19.04.13 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0 {'Class': 0}
blueman-manager 19.04.13 INFO     ManagerDeviceList:369 row_update_event: ServicesResolved False
blueman-manager 19.04.13 INFO     ManagerDeviceList:369 row_update_event: Connected False
blueman-manager 19.04.13 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0/dev_1C_6E_4C_80_0D_6E {'ServicesResolved': False, 'Connected': False}
blueman-manager 19.04.13 DEBUG    Base:60 do_g_properties_changed: /org/bluez/hci0 {'Powered': False, 'Discovering': False}
blueman-manager 19.04.13 INFO     PulseAudioUtils:324 __event_callback: 41 162
[...]

PR #2163, opened 2023-10-10
blueman-manager: Rework bluetooth disabled dialog

infirit:
I want to cleanup all the nested function in `do_activate`, having some is ok but imo this is too much. Putting it out here for quick testing.

### DIFF
--- a/blueman/main/Manager.py
+++ b/blueman/main/Manager.py
@@ -7,7 +7,7 @@ from blueman.bluez.Adapter import Adapter
 from blueman.bluez.Device import Device
 from blueman.bluez.Manager import Manager
 from blueman.bluez.errors import DBusNoSuchAdapterError
-from blueman.Functions import *
+from blueman.Functions import setup_icon_path, bmexit, launch, e_
 from blueman.gui.manager.ManagerDeviceList import ManagerDeviceList
 from blueman.gui.manager.ManagerToolbar import ManagerToolbar
 from blueman.gui.manager.ManagerMenu import ManagerMenu
@@ -40,6 +40,8 @@ class Blueman(Gtk.Application):
         s.attach()
 
     window: Optional[Gtk.ApplicationWindow]
+
+    __bt_status_dialog: Gtk.MessageDialog
 
     def do_startup(self) -> None:
         def doquit(_a: Gio.SimpleAction, _param: None) -> None:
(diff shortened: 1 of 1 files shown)


infirit:
lol, just realized a new menu entry was added to handle this and now it's unreachable when bluetooth is off :fearful: . I can just drop the check/dialog completely but how do we prevent users interacting with the list?

cschramm:
It makes sense to me that the dialog shows up on start but not if Bluetooth gets switched off while blueman-manager is open, but we can also consider dropping the dialog completely in favor of the menu entry. Users should be curious enough to find it in the all-empty, non-functional manager window.

infirit:
I'm mostly concerned with users interacting with the list as most actions will result in errors.

cschramm:
Ah, shoot, I keep forgetting that I'm kind of blind there as my internal adapter gets removed when rfkill blocked (i.e. my list is empty then).

Now I'm immediately wondering: Don't you run into the very same "interact with unavailable devices in the list" problem

a) if the adapter gets powered off or rfkill blocked (without getting removed) while blueman-manager is already running and
b) if you have multiple adapters and *some* are powered off or rfkill blocked.

And yes, while I cannot test rfkill and do not have an external adapter at hand, I can e.g. `bluetoothctl power off` while blueman-manager is running and of course run into errors when trying to connect to devices (both in main and 2.3.5) and with #2132 I can even do that from within blueman-manager now (when disabling the `Killswitch` plugin so that it actually only gets powered off and not removed in my case).

My conclusion is: Yes, there definitely are rough edges (that we should care about), but even without #2132 and with the Bluetooth disabled dialog they're nothing new.

cschramm:
Quick draft for hiding connect items: https://github.com/blueman-project/blueman/commit/a751ad1c8062d2b58fba9fb7620ad45567a26f12

Ideally, some items in `DeviceMenuItem.Group.ACTIONS` should also get hidden, but not all.

Also "Send a file" in the toolbar needs to get disabled and I just realized that its search button is always available. :see_no_evil:

infirit:
The simplest option is to clear/not populate when unpowered. Then show a label that states what the state of things are in. Like 

* Bluetooth not running
* adapter Killswitch enabled
* blueman applet not running
* something I missed

With instructions on how to change the state perhaps...

cschramm:
That would be another quick fix, yes, but I do not really like the idea of hiding the devices as you can still interact with them in terms of removing, trusting, unpairing, renaming etc.

cschramm:
#2183 tries to fix the manager for an unpowered adapter. No matter if we go that way or clear the list, I still fancy to just drop the dialog to fix #1119 #1707 #2161. Anything else will keep things inconsistent as the dialog tries to avoid *starting* blueman-manager with disabled Bluetooth but it is very well possible to have it running with an unpowered adapter.

PR #2176, opened 2023-10-23
RFC: BluezAgent: Use 5 digit pin for request_pin/passkey

infirit:
Is there a reason why we don't make a pin for the user instead of having them type one in a dialog?

ps: completely untested


cschramm:
That's just not how it works. Generally with legacy pairing or when one device has only input or only display capability, one side displays a code and the other side requests to enter *that* code. There are deviations, e.g. where a code is hardcoded and you have to enter *that* code on the other end.

I probably totally do not get the intention, but I'm not aware of a single case where the user may choose the requested code.

infirit:
You're not missing anything. I was confused by discussion where blueberry apparently was showing a pin.

PR #2183, opened 2023-11-04
Respect adapter power state in manager

cschramm:
Toolbar: Disable search and send buttons if the adapter is not powered (unpaired devices should not be present anyway).

Device menu: Do not add the following if not powered:
* Connect items
* Send file
* Send note

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -5,6 +5,7 @@ import logging
 import cairo
 import os
 
+from blueman.bluez.Adapter import Adapter
 from blueman.bluez.Battery import Battery
 from blueman.bluez.Device import Device
 from blueman.bluez.Manager import Manager
@@ -242,7 +243,7 @@ class ManagerDeviceList(DeviceList):
             if self.menu.show_generic_connect_calc(row["device"]['UUIDs']):
                 if row["connected"]:
                     self.menu.disconnect_service(row["device"])
-                else:
+                elif Adapter(obj_path=row["device"]["Adapter"])["Powered"]:
                     self.menu.connect_service(row["device"])
 
         if event.type == Gdk.EventType.BUTTON_PRESS and cast(Gdk.EventButton, event).button == 3:
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -5,6 +5,7 @@ from operator import attrgetter
 from typing import Dict, List, Tuple, Optional, TYPE_CHECKING, Union, Iterable
 
 from blueman.Functions import create_menuitem, e_
+from blueman.bluez.Adapter import Adapter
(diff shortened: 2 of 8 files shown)


infirit:
I'll test this over the weekend. I just realized we should probably include blocked devices, the rules apply.

The error we display when unpowered is actually not that bad in the manager. I'm not at my machine but it's something including it being unpowered.

infirit (commented):
The `ManagerDeviceMenu` in the menubar is not regenerated like the popup. I think it will be enough to call `self.device_menu.generate()` in here https://github.com/blueman-project/blueman/blob/40449c0fc350cd62806da3407d1792c47b45adef/blueman/gui/manager/ManagerMenu.py#L174C24-L174C24

cschramm:
Good catch. I also just remembered that I wanted to care about the double click to connect.

> I just realized we should probably include blocked devices, the same rules apply.

:thinking: You mean you run into similar errors when trying to connect to a blocked device? Hm, true, makes sense.

> The error we display when unpowered is actually not that bad in the manager. I'm not at my machine but it's something including it being unpowered.
> 
> edit: This is the full error without this PR (`Verbinding mislukt: br-connection-adapter-not-powered`)

If we expect br-connection-adapter-not-powered / le-connection-adapter-not-powered (and we actually should have before this) it should get added to `ManagerDeviceMenu._BLUEZ_ERROR_MAP`, but I think it's better to just avoid it completely. On the other hand, br-connection-aborted-by-local / le-connection-abort-by-local is expected when unpowering the adapter during a connection attempt. :unamused:

infirit (approved):

infirit:
> > I just realized we should probably include blocked devices, the same rules apply.
> 
> :thinking: You mean you run into similar errors when trying to connect to a blocked device? Hm, true, makes sense.

The error is that the device does not respond after waiting for a long time. I'll do some testing now this landed.

> > edit: This is the full error without this PR (`Verbinding mislukt: br-connection-adapter-not-powered`)
> 
> If we expect br-connection-adapter-not-powered / le-connection-adapter-not-powered (and we actually should have before this) it should get added to `ManagerDeviceMenu._BLUEZ_ERROR_MAP`, but I think it's better to just avoid it completely. On the other hand, br-connection-aborted-by-local / le-connection-abort-by-local is expected when unpowering the adapter during a connection attempt. :unamused:

I'll play around a bit and see if I can get those errors and add them to the map.

cschramm:
I just wanted to write that it's super easy: Just start connecting to a device that is not online and then turn off Bluetooth. Funny enough, I did not get abort(ed)-by-local this time, but by-remote... The problem with these is that I think they can actually mean a range of things. I even explicitly excluded them from #1363 (not knowing that you can easily reach them by playing with the power state).

cschramm:
Oh, I did not know of https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/errors.txt yet. At least it's pretty clear on ECONNRESET. Not so much ECONNABORTED.

PR #2194, opened 2023-11-13
Fix comment in ISSUE_TEMPLATE

cschramm:
I just realized why so many insufficient issues tend to contain hidden text. (Not that making it visible makes them much better... :sweat_smile:)


infirit (approved):
Great, now I don't have to edit these manually 👍

cschramm:
I was a little bit confused as I was sure that the closing tag worked at some point but it definitely does not anymore.

PR #2198, opened 2023-11-18
Remove check_bluetooth_status from blueman-manager

cschramm:
The function is completely unreliable regarding its option to turn Bluetooth on and regularly leads to closing the application while doing so (#1119 #2161). Also, the whole system is incosistent as while it avoids starting blueman-manager with Bluetooth powered off, you can absolutely run blueman-manager with Bluetooth powered off e.g. simply by turning it off while it's open (#1707). There is even a menu item for turning Bluetooth off (and back on!) now.

Fun fact: Turning Bluetooth off while blueman-manager is running is actually meant to lead to the same dialog (and should have in 2.0), but #543 broke the signal handling: Since 2.1 `status` is set to a 1-tuple returned by `unpack` instead of a `bool` and the tuple is always `True`ish.

Note: blueman-adapters suffers from the same glitches (unreliable status check when turning Bluetooth on via the dialog, keeps running when Bluetooth gets turned off). That is not handled here.

@infirit: I'm not quite sure if you're in the same boat or rather want to to keep (and fix) the status check, so take this as an RFC.

### DIFF
--- a/blueman/gui/manager/ManagerStats.py
+++ b/blueman/gui/manager/ManagerStats.py
@@ -25,8 +25,11 @@ class ManagerStats:
 
         blueman.List.connect("adapter-changed", self.on_adapter_changed)
 
-        assert blueman.List.Adapter is not None
-        self.hci = adapter_path_to_name(blueman.List.Adapter.get_object_path())
+        self.hci = (
+            None
+            if blueman.List.Adapter is None
+            else adapter_path_to_name(blueman.List.Adapter.get_object_path())
+        )
 
         self.time = None
 
--- a/blueman/main/Manager.py
+++ b/blueman/main/Manager.py
@@ -6,7 +6,6 @@ from typing import Optional, Any, Tuple
 from blueman.bluez.Adapter import Adapter
 from blueman.bluez.Device import Device
 from blueman.bluez.Manager import Manager
-from blueman.bluez.errors import DBusNoSuchAdapterError
 from blueman.Functions import *
 from blueman.gui.manager.ManagerDeviceList import ManagerDeviceList
 from blueman.gui.manager.ManagerToolbar import ManagerToolbar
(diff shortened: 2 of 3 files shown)


infirit:
I'll check this out asap, work has been busy the last couple of weeks.

Generally I agree that blueman apps should not close when not powered or even bluetoothd not running. 

My biggest worry is the UI kinda assumes they are in a lot of places. With the dialog it is quite obvious something is off/not running. We can hide certain menu entries but how do we communicate why they are hidden.

infirit:
I have devices that are only Trusted not Paired (dualshock 3 in this case) so this should also be handled in the menu/toolbar. Other than that it works as advertised.

But there is pretty no feedback to the user why and they can't connect/pair/etc their devices.  I don't really have a good solution for this either.

Also ran into this while testing.

```python
blueman-manager 13.58.31 INFO     ManagerMenu:191 on_adapter_selected: selected /org/bluez/hci0
  File "/usr/lib/python3.11/site-packages/blueman/gui/manager/ManagerToolbar.py", line 70, in on_device_selected
    self._update_buttons(Adapter(obj_path=device["Adapter"]))
                                          ~~~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

Fix below. Feel free to claim it as your own or I can send it as a PR later.

```
From 211f3b6f69c3e14208269e172400e38ed051dece Mon Sep 17 00:00:00 2001
From: Sander Sweers <infirit@gmail.com>
Date: Sat, 16 Dec 2023 14:04:52 +0100
Subject: [PATCH] ManagerToolbar: device is optional

---
 blueman/gui/manager/ManagerToolbar.py | 6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)

diff --git a/blueman/gui/manager/ManagerToolbar.py b/blueman/gui/manager/ManagerToolbar.py
index c042b8d61..e5a11a6d5 100644
--- a/blueman/gui/manager/ManagerToolbar.py
+++ b/blueman/gui/manager/ManagerToolbar.py
@@ -66,7 +66,11 @@ class ManagerToolbar:
         logging.debug(f"toolbar adapter {adapter_path}")
         self._update_buttons(None if adapter_path is None else Adapter(obj_path=adapter_path))

-    def on_device_selected(self, dev_list: ManagerDeviceList, device: Device, _tree_iter: Gtk.TreeIter) -> None:
+    def on_device_selected(
+            self, _dev_list: ManagerDeviceList, device: Optional[Device], _tree_iter: Gtk.TreeIter
+    ) -> None:
+        if device is None:
+            return
         self._update_buttons(Adapter(obj_path=device["Adapter"]))

     def _update_buttons(self, adapter: Optional[Adapter]) -> None:
--
2.43.0
```

cschramm:
Already fixed in #2199.

I'll look into "trusted, not paired".

infirit:
I thought about this and I think we should add two new icons to the statusbar to indicate,
1. Whether bluetoothd is running
2. if the adapter is powered

Both with tool-tip explaining what to do.

Then we can drop the dialog from blueman manager, does this sound right to you?

cschramm:
Absolutely, yes. I think the status bar can take the burden of more content. Are you going to add those?

infirit:
Here is an example of what I am thinking off for the UI. I haven't put much thought into the wording of the tooltip or the status name.
image

infirit (approved):
Let's merge this and fix any issues afterwards.

PR #2224, opened 2023-12-16
Show hci device in tab label

infirit:
It get's really confusing which adapter a user is modifying when it just shows what BlueZ initially set. BlueZ uses the hostname as name and subsequent adapters appends `#N`. 
image


infirit:
Removed the left over debug log and dropped the commit that used adapter_path_to_name for hcidev. That latter made mypy very unhappy which I don't want to deal with right now.

cschramm (approved):

PR #2231, opened 2023-12-29
Use small toolbar size and text beside icons

axel358:
Screenshot from 2023-12-29 09-49-01
Text can help new users identify actions easily and since there are few toolbar items i don't think space will be a problem with languages like Russian

### DIFF
--- a/blueman/gui/manager/ManagerToolbar.py
+++ b/blueman/gui/manager/ManagerToolbar.py
@@ -91,8 +91,7 @@ class ManagerToolbar:
             self.b_remove.props.sensitive = True
             self.b_send.props.sensitive = powered and row["objpush"]
 
-            icon_name = "blueman-untrust-symbolic" if row["trusted"] else "blueman-trust-symbolic"
-            self.b_trust.props.icon_widget = Gtk.Image(icon_name=icon_name, pixel_size=24, visible=True)
+            self.b_trust.props.icon_name = "blueman-untrust-symbolic" if row["trusted"] else "blueman-trust-symbolic"
             self.b_trust.props.label = _("Untrust") if row["trusted"] else _("Trust")
 
     def on_device_propery_changed(self, dev_list: ManagerDeviceList, device: Device, tree_iter: Gtk.TreeIter,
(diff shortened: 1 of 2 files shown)


infirit:
Just to be sure as I see no device address, you removed them from the screenshot?

Changes look OK to me but I will have a better look later this week and test it on various themes.

axel358:
Yes I only changed the toolbar UI file and replaced the icon_widget prop of the Trust Tool button with an icon_name instead on the python side

infirit:
Had a better look and the toolbar looks a bit better, but... It exposes an alignment and spacing issue with the Trust button label due to how we force a specific size to avoid buttons moving within the toolbar when the label text is updated (see screenshot 1).

Fixing this is easy if we have direct access to the label widget by setting the `xalign` property to `0`. But we don't and I would prefer to avoid fetching it from the container. But even with setting `xalign` we have a lot of ugly empty space in the button.

I did a quick experiment setting `xalign` and a different way of forcing the size but it looks worse (see 2nd screenshot). I get the feeling this is why blueman only shows an icon :sweat_smile: .

The reason it looks like this for me as in Dutch there is no short translation for "Untrust", "Vertrouwen beëindigen". I suspect it's similar in other languages.

The buttons jumping around in the toolbar is distracting and annoying so that's not an option. Not sure how to move forward with this at the moment...

image

image

axel358:
What about making that button/group the last one on the toolbar?

axel358:
Hwy there, it's been a while. Any chances of  at least getting the small toolbar icon size changes in? The icons look a bit too big compared to the text next to them

cschramm:
Hey @axel358, the smaller icons seem to make sense in any case, yes. Would you like to prepare that?

infirit:
> Hey there, it's been a while. Any chances of at least getting the small toolbar icon size changes in? The icons look a bit too big compared to the text next to them

Sorry, been extremely busy lately and now picking things back up on blueman. I agree with @cschramm, the smaller toolbar is fine. The buttons I really want to leave as is for now.

I hopefully have a reworked devicelist (Gtk.ListBox based) ready at the beginning of next year for blueman manager. Initially it will just be replicating what we already have. But then I want make several of these buttons/actions part of the Gtk.Box that will be in the row.

PR #2265, opened 2024-02-16
Add bluetooth status icon in the statusbar

infirit:
I would have liked to show the state of the adapter and the killswitch. While getting the adapter state is easy, the state of the killswitch is not. So for now get the global state from the applet.

I'll rebase when the PR for the dialog is merged.

### DIFF
--- a/blueman/main/Manager.py
+++ b/blueman/main/Manager.py
@@ -84,6 +84,23 @@ class Blueman(Gtk.Application):
                 margin_right = statusbar.get_margin_right()
                 statusbar.set_margin_right(margin_right + 10)
 
+            def update_bt_status(state: bool) -> None:
+                if state:
+                    icon_name = "bluetooth"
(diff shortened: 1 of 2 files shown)


cschramm (approved):

cschramm (approved):

infirit:
What kind of interactivity are you looking for? Make it a button like  the one from the Adapter menu? Or some kind of popover to show the status changed?

cschramm:
It just bugs me that the tooltip sends me to a menu in the opposite corner instead of a possibility to just toggle the status from the icon somehow. Part of it might be that it has similarities with the tray icon (also bottom right on my desktop). I don't really have a good idea for the implementation, though. Toggling by just clicking the icon could be surprising. Maybe a click could open a 1-item popup menu or a dialog asking for confirmation.

cschramm:
Makes perfect sense to me, yes. :+1:

PR #2267, opened 2024-02-19
Copy address to clipboard from manager device list

cschramm:
#2235


cschramm:
@HubKing: Could you try this?

infirit (approved):
Quick test and it does what should.

PR #2283, opened 2024-02-28
Manager: Use Gtk.Switch to toggle bluetooth status

infirit:
Here is what I have so far. Still need to:
* cleanup
* drop menu entry
* perhaps change color of the button

The idea here is that the toggle button is "active" when bluetooth is enabled and showing label and icon what would happen if the button is pressed, disable bluetooth.

image

image

### DIFF
--- a/blueman/gui/manager/ManagerMenu.py
+++ b/blueman/gui/manager/ManagerMenu.py
@@ -8,7 +8,6 @@ from blueman.bluez.Manager import Manager
 from blueman.gui.manager.ManagerDeviceList import ManagerDeviceList
 from blueman.gui.manager.ManagerDeviceMenu import ManagerDeviceMenu
 from blueman.gui.CommonUi import show_about_dialog
-from blueman.main.DBusProxies import AppletService
 from blueman.Constants import WEBSITE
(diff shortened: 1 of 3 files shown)


cschramm:
I think that makes a lot of sense. :ok_hand:

Not sure about the "active" visualization, though. It feels wrong that the button is highlighted (in sense of a primary action) when it allows to turn Bluetooth off while it's not when it allows to turn it on. From that perspective, I'd expect the contrary.  Not sure, but it might make sense to just not highlight any of the states (I see that the state might be less obvious then).

infirit:
Perhaps a Gtk.Switch is more appropriate as it can have proper on/off labels. Let me are how that looks.

infirit:
Here is how I can look with a switch. I didn't add a label as I don't see what that would add, I can so let me know if you want it.
image
image

cschramm:
I think that's perfect!

infirit:
Cool, I'll commit that and push it tonight.

infirit:
Updated with the switch

cschramm (approved):

cschramm (approved):

PR #2362, opened 2024-05-09
Only show a singly battery notifcation

### LINKED ISSUE #2317: Annoying battery notifications
blueman: 2.4
BlueZ: 5.73
Distribution: Arch Linux
Desktop environment: Sway

* [X] I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

Hi,

I'm getting a notification every time my earbuds' batteries drop by 1%. It is getting pretty annoying, how do I disable that?

Regards,
Jules

cschramm:
From feedback in #2317 it became pretty clear that battery data is a mess with devices providing levels like full, medium, low and devices providing fine-grained percent values, flapping values etc. and we do not want to go into power management applet business, so we only show a single battery notification on connection (when the battery interface shows up or - if the interface is already present when the device connects - when the existing interface sends its first property update) and drop the reference to the connection notification to update so that further values do not lead to a notification. (In case people will ask for those notifications we might add them as a separate plugin that's disabled by default.)

Due to the great demand, I plan to backport this for 2.4.2, even though it's technically a regression.

Closes #2317

### DIFF
--- a/blueman/plugins/applet/ConnectionNotifier.py
+++ b/blueman/plugins/applet/ConnectionNotifier.py
@@ -36,7 +36,7 @@ class ConnectionNotifier(AppletPlugin):
                 Notification(device.display_name, _('Disconnected'), icon_name=device["Icon"]).show()
 
     def _on_battery_update(self, path: str, value: int) -> None:
-        notification = self._notifications.get(path, None)
+        notification = self._notifications.pop(path, None)
         if notification:
             try:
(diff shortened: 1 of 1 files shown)


ouuan (approved):
This works as expected for me (with Galaxy Buds 2).

infirit (approved):

PR #2364, opened 2024-05-10
Rename recent connections header

### LINKED ISSUE #2328: Regression 2.4: "Recent Connections" is empty
blueman: 2.4-1
BlueZ: 5.73-4
Distribution: arch
Desktop environment: wayland

Updating from 2.3 to 2.4 introduced a regression having an empty list in the "Recent Connection" tab in the applet. Downgrading to 2.3 showed them again:

2.3:
bm-2 3 5

2.4:
bm-2 4 0

Either this or the idea was to remove "Recent Connections" completely as all audio profiles are now inlined?

cschramm:
I hope this makes things clearer and avoids confusion due to the change from 2.3 to 2.4.

Closes #2328

### DIFF
--- a/blueman/plugins/applet/RecentConns.py
+++ b/blueman/plugins/applet/RecentConns.py
@@ -185,7 +185,7 @@ class RecentConns(AppletPlugin, PowerStateListener):
         menu: "Menu" = self.parent.Plugins.Menu
         self._mitems: List[MenuItem] = []
         menu.unregister(self)
-        menu.add(self, 52, text=_("Recent _Connections"), icon_name="document-open-recent-symbolic",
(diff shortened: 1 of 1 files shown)


infirit:
You dropped the accelerator in the string not sure if you intended this. If not let's do `Reconne_ct to…`

cschramm:
I think an accelerator is pointless for the insensitive item, isn't it?

infirit:
It doesn't which make me wonder why it was there in the first place lol.

cschramm:
It used to make sense when the item had a submenu and we just didn't remove the accelerator (which also made sense to not mess up translations).

PR #2376, opened 2024-05-19
Also extradist testhelpers

infirit:
This should be the remaining bits for all the tests. Anything else that got missed?


cschramm (changes requested):
`test/services/meta/Makefile.am` seems broken. I could not spot anything else.

infirit:
Fixed.

cschramm (approved):

PR #2386, opened 2024-05-23
DBusProxies: Make method calls asynchronous

infirit:
Non testing version of #2384.

This is a good idea regardless if it solved the problem or not.


cschramm (dismissed):

infirit:
Sorry, I messed up the typing because I returned from `self.call` which is always `None`.  And now mypy is yelling at me for issue in PluginManager :sob:

cschramm:
Unrelated to this, related to mypy 1.11, fixed with :crossed_fingers: in #2437

infirit:
Re-based against main and now all is well :balloon:

cschramm (approved):

PR #2424, opened 2024-07-02
Add appstream metadata

androidnisse:
https://www.freedesktop.org/wiki/Distributions/AppStream/

Needed for this application to show up in various Linux software centers such as gnome-software and discover.

### DIFF
--- a/data/io.github.blueman_project.blueman.metainfo.xml
+++ b/data/io.github.blueman_project.blueman.metainfo.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>io.github.blueman_project.blueman</id>
+  
+  <name>blueman</name>
+  <summary>GTK+ Bluetooth manager</summary>
+  
+  <metadata_license>FSFAP</metadata_license>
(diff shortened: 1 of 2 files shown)


infirit:
Personal initial opinion; I'm not sure I want blueman to show up. 

One reason is that blueman, bluedevil and gnome-bluetooth tend to fight for control (default agent for example).

I know in a couple of distributions you can't have blueman and gnome-bluetooth installed at the same time. At least that is what I remember from a couple years ago so this may have changed. How would that work if people try to install blueman, will gnome-bluetooth be uninstalled? Or is it then impossible to install blueman.

I also don't see blueman as a typical end user application as it manages the lower level bluetooth stack.

I'm not against adding it but I want to make sure we think this through properly.
 
@androidnisse I appriciate the PR so please don't take the above as a personal critique :-).

androidnisse:
Yes no problem. You are having valid points, I am not sure how this program works with others or if you want it to show up. I have added this to the package in Solus so there it will show up. So if you want to close this I will not take it personal. :)

cschramm:
I'm not sure about this either. That metainfo seems like something distributions could add if it fits them, but not something for upstream. On the other hand, distributions might end up adding the same file over and over again, of course.

PR #2435, opened 2024-07-18
meson.build: use internal cython

MarkusVolk:
This prevents the buildpath from being built into _blueman.so and thus fixes reproducibility

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,8 +1,8 @@
 project(
-    'blueman', 'c',
+    'blueman', 'c', 'cython',
     version: '2.4',
     license: 'GPL3',
-    meson_version: '>=0.56.0',
+    meson_version: '>=0.63.0',
     default_options: 'b_lundef=false'
 )
 
--- a/module/meson.build
+++ b/module/meson.build
@@ -1,10 +1,8 @@
 cython = find_program('cython', 'cython3' ,required: true)
 
-blueman_c = custom_target(
+blueman_c = pyinstall.extension_module(
     'blueman_c',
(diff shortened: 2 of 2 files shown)


MarkusVolk:
Sorry. It's not working like this. The lib is not getting properly built. I'll close the pr for now

MarkusVolk:
> Why did you bump the min required version of meson to 0.63.0?

I also read that this function is available since 0.59.0, but meson said at do_configure that I use functions from 0.63.0. 

This is also all new to me and I have yet to figure out exactly how this is supposed to work. I'll have another look at it, but unfortunately it doesn't help with my current problem. Unfortunately the buildpath is still in _blueman.so

static const char __pyx_k_home_flk_poky_build_hypr_tmp_wo[] = "/home/flk/poky/build-hypr/tmp/work/corei7-64-poky-linux/blueman/2.4.2/git/module/_blueman.pyx";

infirit:
> I also read that this function is available since 0.59.0, but meson said at do_configure that I use functions from 0.63.0.

I'll have a look this weekend.

> This is also all new to me and I have yet to figure out exactly how this is supposed to work. I'll have another look at it, but unfortunately it doesn't help with my current problem. Unfortunately the buildpath is still in _blueman.so
> 
> static const char __pyx_k_home_flk_poky_build_hypr_tmp_wo[] = "/home/flk/poky/build-hypr/tmp/work/corei7-64-poky-linux/blueman/2.4.2/git/module/_blueman.pyx";

Yeah, I have no idea how to fix that.

PR #2444, opened 2024-07-31
Initial meson devenv

infirit:
If the build dir is `build` use `meson devenv -C build`. Seems to work ok from the command line. 

Not sure how this can be integrated into pycharm. The env can be dumped with `meson devenv -C build --dump` in several formats, maybe it can be used in pycharm somehow.

edit: When ubuntu latest moves to >=0.62 start making use of the method keyword.


infirit:
Nope, broken I just didn't notice.

PR #2449, opened 2024-08-05
Set Toolbar buttons half the opacity when not sensitive

infirit:
This is especially noticeable when people try to use the Pair button with only keyboard as the button is skipped/not selectable when insensitive. PR is mainly to get feedback if this is what we want. One downside is that when text is slightly less readable as can be seen with the Send File button. I can also only do this for the Pair button.

This is how it looks like now:
image

Before:
image


cschramm:
Makes perfect sense to me. I find the difference for the "Send File" text in the images hard to even notice.

infirit:
Rebased and ready for merge.

cschramm (approved):
Looks good, but has a conflict now

cschramm (approved):

PR #2461, opened 2024-08-15
meson: Do not emit absolute path when S != B

kraj:
build systems like OE build outside sourcetree in such cases it works ok but cython resolves the input file to absolute path and that gets emitted into genetate _blueman.c as module name, renders the build non-reproducible, wish cython had a better way to handle this but there is not, therefore tweak the meson build rule to account for specifying workdir to cython which will search the inputs correctly, and use meson's build_root to emit the output into build dir. This ensures that it becomes independent of source or build directories and cython does not generate the absolute paths into generate C code.

See cython discussion on [1]

[1] https://github.com/cython/cython/issues/5949

### DIFF
--- a/module/meson.build
+++ b/module/meson.build
@@ -1,10 +1,12 @@
 cython = find_program('cython', 'cython3' ,required: true)
 
+cython_outfile = join_paths(meson.project_build_root(), '@OUTPUT@')
+cython_infile = join_paths(meson.project_build_root(), '@INPUT@')
 blueman_c = custom_target(
     'blueman_c',
     output: '_blueman.c',
     input: '_blueman.pyx',
-    command: [cython, '-w', meson.source_root(), '--output-file', join_paths(meson.build_root(), '@OUTPUT@'), join_paths(meson.build_root(), '@INPUT@')])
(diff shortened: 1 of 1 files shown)


infirit:
This broke the meson build, cannyou check. 

Regardless, please use `join_paths()` to build paths in meson.

https://mesonbuild.com/Reference-manual_functions.html#join_paths

infirit:
Also ideally we use mesons native cython support 1, but when I tried I ran into weird compilation errors and put it on hold

1 https://mesonbuild.com/Cython.html

kraj:
> This broke the meson build, cannyou check.
> 
> Regardless, please use `join_paths()` to build paths in meson.
> 
> https://mesonbuild.com/Reference-manual_functions.html#join_paths

done.

kraj:
> This broke the meson build, cannyou check.
> 

yeah I needed to make input relative to buildroot as well. Fixed now.

> Regardless, please use `join_paths()` to build paths in meson.
> 
> https://mesonbuild.com/Reference-manual_functions.html#join_paths

infirit (approved):
Doesn't break anything for me on Arch and CI is happy.

If this helps with reproducible builds ok with me.

PR #2472, opened 2024-08-25
Fix Autoconnect and annotate object paths and bluetooth addresses

infirit:
Like the annotate commit mentions, I was hoping to catch the issue where we passed the wrong string to `Manager.find_device`. I can move this commit to a separate PR if you like.

We have been storing the object path to GSettings for the AutoConnect plugin since the latest release so let's not break it for those users by now storing the bluetooth address instead.

### DIFF
--- a/blueman/bluemantyping.py
+++ b/blueman/bluemantyping.py
@@ -1,4 +1,4 @@
-from typing import Dict, Tuple, Union, TYPE_CHECKING
+from typing import Dict, Tuple, Union, TYPE_CHECKING, NewType
 from gi.repository import GObject
 
 if TYPE_CHECKING:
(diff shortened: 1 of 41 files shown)


cschramm:
Now I'm wondering if it even makes sense to store object paths. :thinking: Are they that stable? If I run a system with adapter A, shut it down exchange it with adapter B, wouldn't B become `hci0` earlier assigned to A? And if I now additionally plug in A, what used to be `hci0` would now be `hci1`, no?

infirit:
I don't think the kernel makes guarantees on detection order on this so let's store the bluetooth address. Updated the PR.

cschramm (approved):
Thinking out loud: 2.3 releases used to store addresses, while current 2.4 releases store object paths due to https://github.com/blueman-project/blueman/commit/9d9af921e5c1ae965512a2a4e32daf057536ff38. This reverts back to addresses. *All* versions treat entries as addresses, meaning that those from 2.4 do no work. This change thus fixes the feature for old entries from 2.3 (users could be surprised, but that's fine, I guess) and newly created entries while those from existing 2.4 releases are just doomed like before but not even visible anymore, so users can set new entries now.

infirit:
It's fairly simple to check for paths and then convert and update the value in GSettings. I can add this in a follow up PR.

Also, wondering. I was planning to only cherry-pick the fix and not typing change for 2-4-stable as it's quite large. Let me know if you like to have it in 2-4-stable anyway :-).

PR #2489, opened 2024-09-12
Rfcomm.py: convert port_id gdouble/float to int

epilys:
When connecting to a serial port from blueman the following exception is raised because `port_id` in `OpenRFCOMM` is passed as `Glib.Variant` `gdouble` `d`:

```
  Connection Failed: File "/usr/lib/python3/dist-packages/blueman/main/DbusService.py", line 154, in _handle_method_call
      method(*(args + (ok, lambda exception: self._return_dbus_error(invocation, exception))))
    File "/usr/lib/python3/dist-packages/blueman/plugins/applet/DBusService.py", line 93, in connect_service
      service.connect(reply_handler=lambda port: ok(), error_handler=err)
    File "/usr/lib/python3/dist-packages/blueman/services/meta/SerialService.py", line 104, in connect
      Mechanism().OpenRFCOMM('(d)', port_id)
    File "/usr/lib/python3/dist-packages/gi/overrides/Gio.py", line 349, in __call__
      result = self.dbus_proxy.call_sync(self.method_name, arg_variant,
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.Failed: Traceback (most recent call last):
    File "/usr/lib/python3/dist-packages/blueman/main/DbusService.py", line 156, in _handle_method_call
      ok(method(*args))
         ^^^^^^^^^^^^^
    File "/usr/lib/python3/dist-packages/blueman/plugins/mechanism/Rfcomm.py", line 14, in _open_rfcomm
      subprocess.Popen([RFCOMM_WATCHER_PATH, f"/dev/rfcomm{port_id:d}"])
                                                          ^^^^^^^^^^^
  ValueError: Unknown format code 'd' for object of type 'float'
   (0)


```

Similar thing happens to `CloseRFCOMM`.

Change the typing annotation to allow for `Union` of `float+int` and explicitly convert to `int` by calling `int(port_id)`.

PR #2492, opened 2024-09-13
Rfcomm: Change dbus type signature to an int

infirit:
Fixes: #2489

### DIFF
--- a/blueman/plugins/mechanism/Rfcomm.py
+++ b/blueman/plugins/mechanism/Rfcomm.py
@@ -7,8 +7,8 @@ from blueman.plugins.MechanismPlugin import MechanismPlugin
 
 class Rfcomm(MechanismPlugin):
     def on_load(self) -> None:
-        self.parent.add_method("OpenRFCOMM", ("d",), "", self._open_rfcomm)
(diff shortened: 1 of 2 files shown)


cschramm (approved):
I'm still wondering why this did not come up yet. Are there any specific conditions that cause the error? :thinking:

infirit:
Not sure. I usually don't use serial devices/rfcomm so I probably never noticed. I completely relied on the the type annotation while I should have done at least a quick test.

cschramm:
Oh, I get it. This only became an issue with https://github.com/blueman-project/blueman/commit/5ea654259c36fd51d43a354fd1af4190e402b376, thus the 2.4 release. I expected an issue that we'd have had for much longer, but it's only due to the f-string.

infirit:
Yeah, I assumed the type hint was correct. Maybe we can change the add_method function to require python types. We might then be able to change the annotation and be more specific than `Callable(..., Any)`. So mypy can help us (mostly me lol) spot these issues.

Maybe a TypeVar if we pass a proper type or maybe even simpler NewType of str for each dbus typestring 🤔 

I may experiment with that when I'm done with the for main PR.

PR #2499, opened 2024-09-24
Applet: Handle UnknownObject DBus error.

tommie:
Every 10 s, I see this error, on Ubuntu 24.04.1, which uses Blueman 2.3.5:

```
 During handling of the above exception, another exception occurred:
   File "/usr/lib/python3/dist-packages/blueman/main/Applet.py", line 97, in on_device_created
     plugin.on_device_created(path)
   File "/usr/lib/python3/dist-packages/blueman/plugins/applet/RecentConns.py", line 99, in on_device_created
     self._rebuild()
   File "/usr/lib/python3/dist-packages/blueman/plugins/applet/RecentConns.py", line 80, in _rebuild
     self._items = self._get_items()
                   ^^^^^^^^^^^^^^^^^
   File "/usr/lib/python3/dist-packages/blueman/plugins/applet/RecentConns.py", line 199, in _get_items
     return sorted(
            ^^^^^^^
   File "/usr/lib/python3/dist-packages/blueman/plugins/applet/RecentConns.py", line 213, in <genexpr>
     if i["adapter"] not in adapter_addresses or self._get_device_path(i["adapter"], i["address"])),
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/lib/python3/dist-packages/blueman/plugins/applet/RecentConns.py", line 193, in _get_device_path
     device = self.parent.Manager.find_device(address, adapter.get_object_path())
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/lib/python3/dist-packages/blueman/bluez/Manager.py", line 146, in find_device
     if device['Address'] == address:
        ~~~~~~^^^^^^^^^^^
   File "/usr/lib/python3/dist-packages/blueman/bluez/Base.py", line 135, in __getitem__
     return self.get(key)
            ^^^^^^^^^^^^^
   File "/usr/lib/python3/dist-packages/blueman/bluez/Base.py", line 106, in get
     raise parse_dbus_error(e)
 blueman.bluez.errors.BluezDBusException: org.freedesktop.DBus.Error.UnknownObject Method "Get" with signature "ss" on interface "org.freedesktop.DBus.Properties" doesn't exist
```

Since there is already exception handling for the adaptor going away (https://github.com/blueman-project/blueman/commit/c67ae920c5bfd3932ec6715cc0336139b436f88b), it seems reasonable to add this, in case of devices that "flicker". I can't say for sure why my device is listed by the adapter, but don't actually exist.

(I've also had the applet pinned at 100% CPU, but I think that's a separate issue.)

### DIFF
--- a/blueman/bluez/errors.py
+++ b/blueman/bluez/errors.py
@@ -85,6 +85,10 @@ class DBusUnsupportedMajorClassError(BluezDBusException):
     pass
 
 
+class DBusUnknownObjectError(BluezDBusException):
+    pass
+
+
 class DBusServiceUnknownError(BluezDBusException):
     pass
 
@@ -121,6 +125,7 @@ __DICT_ERROR__ = {'org.bluez.Error.Failed': DBusFailedError,
                   'org.bluez.Error.AuthenticationCanceled': DBusAuthenticationCanceledError,
                   'org.bluez.serial.Error.NotSupported': DBusNotSupportedError,
                   'org.bluez.Error.UnsupportedMajorClass': DBusUnsupportedMajorClassError,
+                  'org.freedesktop.DBus.Error.UnknownObject': DBusUnknownObjectError,
                   'org.freedesktop.DBus.Error.ServiceUnknown': DBusServiceUnknownError}
 
 
--- a/blueman/plugins/applet/RecentConns.py
+++ b/blueman/plugins/applet/RecentConns.py
@@ -7,7 +7,7 @@ from typing import List, TYPE_CHECKING, Optional, Callable, cast, Union
 from blueman.bluemantyping import ObjectPath, BtAddress
 
 from blueman.bluez.Device import Device
-from blueman.bluez.errors import DBusNoSuchAdapterError
+from blueman.bluez.errors import DBusNoSuchAdapterError, DBusUnknownObjectError
 from blueman.gui.Notification import Notification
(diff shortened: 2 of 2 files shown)


infirit:
We should handle this error but I want to know why/how you got into this loop. Recentconns is notified of a new device which it may or may not know and tries to rebuild it's menu. It fails because the just created device doesn't exist anymore, which I find a little unusual.

Can you kill blueman-applet and run it from a terminal with `blueman-applet --debug` and let it log for a couple of minutes while this is going on and post/link it here?

tommie:
I'll try to do that. This triggers very rarely, probably around suspend/resume. When it happens, it gets stuck repeating the same error every 10 s.

infirit:
If it happens with suspend/resume it might be the adapter isn't restoring it's state properly on resume. Kernel log may also provide some clues then.

infirit (approved):
The change is fine. If you do find the time to get some logs I would still like to see them.

tommie:
Thanks. :)

I've actually been running an unpatched Blueman since we last spoke, but haven't been able to trigger it. I only suspend 1-2 times a day, though. I'll get back to you when it happens.

PR #2506, opened 2024-10-04
Dbus method added to set a flag for the paring button

astcri:
**Environment**
Ubuntu22.04

**Blueman version**
Commit hash: 52f0f03

**Problem**
Every time a headset is paired with the paring button, the notification of "Connected" and "Disconnected" pops up even though the headset is neither connected nor disconnected. 

**Cause**
This problem seems to be caused by that it receives "Connected" and "Disconnected" signal even when it is just paring.

**Modification**
Dbus method is added to set a flag to judge if a device is paring using the paring button. The method is accessed from bond(), which is a callback method set for the paring button. Also, since it seems bond() is never used as a static method, the decorator was removed.

### DIFF
--- a/blueman/main/Manager.py
+++ b/blueman/main/Manager.py
@@ -254,13 +254,13 @@ class Blueman(Gtk.Application):
             dialog.connect("close", lambda d: d.destroy())
             dialog.show()
 
-    @staticmethod
-    def bond(device: Device) -> None:
+    def bond(self, device: Device) -> None:
         def error_handler(e: Exception) -> None:
             logging.exception(e)
             message = f"Pairing failed for:\n{device.display_name} ({device['Address']})"
             Notification('Bluetooth', message, icon_name="blueman").show()
-
+        path = device.get_object_path()
+        self.Applet.SetParingState("(sb)", path, True)
         device.pair(error_handler=error_handler)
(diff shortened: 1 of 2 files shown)


infirit:
> This problem seems to be caused by that it receives "Connected" and "Disconnected" signal even when it is just paring.

Well, it's not just pairing though, the device is literally being connected to for the pair/bond. And when explicitly asking only to pair (which is what blueman does with the pair option) the connection is dropped afterwards. 

What you probably want to use instead is the connect option and let BlueZ's autopair plugin deal with the bond/pair. With that you only get the connection notification and the device remains connected.

astcri:
"I'm aware that the connect option keeps devices connected. However, it would be better to prevent notifications from appearing when it is just paring, since there is an option to do so. Or would you recommend replacing the pairing button with the connect button?

infirit:
Sorry wanted to reply earlier. When blueman shows the Connect option in the menu it's recommended users use that over the Pair entry. It has to do with how BlueZ can handle pairing (autopair plugin mostly) for many devices on it's own without the need for frontends to explicitly Pair.

Our UI definitely can use some work to communicate this better to the users. Like adding a connect button to the toolbar.

PR #2513, opened 2024-10-17
UI Fix: Make search button available after device list becomes empty

astcri:
**Problem**
The search button becomes unavailable after the device list is empty.

**Cause**
The _update_button() function receives None when the device list becomes empty, causing the sensitivity of the search button to be disabled if the argument (adapter) is None.

**Solution**
Pass the instance variable of ManagerDeviceList, Adapter, to the _update_button() function.


infirit (approved):
@cschramm we should include this in the next stable, I'll open a PR.

@astcri good catch.

Issue #2510, opened 2024-10-15
Blueman is suddenly not working


vmxt:
blueman: 2.4.3-1
BlueZ: 5.78-1
Distribution: Linux archlinux 6.11.3-zen1-1-zen
Desktop environment: xfce4

* ✅ I have consulted the [Troubleshooting page](https://github.com/blueman-project/blueman/wiki/Troubleshooting) and done my best effort to follow.

I went through the troubleshooting steps, but nothing worked. I'm using a bluetooth dongle because my board lacks built-in Bluetooth. It was working earlier, but now Bluetooth is not working. The blueman app cannot click search, despite the fact that I have started, enabled, and restarted the bluetooth services. I tried the manual `rfkill unblock blutooth` and `modprobe btusb` and etc; it works perfectly on status and actively running, but there is no trace of working bluetooth.

bluetooth status
![image](https://github.com/user-attachments/assets/f9a51662-f1da-44e7-8c18-80e3a2f7032b)

blueman
![image](https://github.com/user-attachments/assets/16229974-0bc2-4939-93dc-9ec7a77c83c0)

What went missing? Any hints?

cschramm:
What does `bluetoothctl list` show?

vmxt:
Hi @cschramm , as requested to answer your ff. questions

> What does `bluetoothctl list` show?

Nothing shows

> Also please provide the output of `blueman-applet --loglevel debug` with blueman-applet initially not running.

```
blueman-applet 15.15.32 INFO     PluginManager:86 load_plugin: ['PPPSupport', 'DiscvManager', 'AuthAgent', 'DhcpClient', 'StatusNotifierItem', 'StandardItems', 'ExitItem', 'AutoConnect', 'GameControllerWakelock', 'DisconnectItems', 'PulseAudioProfile', 'DBusService', 'StatusIcon', 'Menu', 'SerialManager', 'TransferService', 'ConnectionNotifier', 'NMDUNSupport', 'Networking', 'KillSwitch', 'RecentConns', 'NetUsage', 'NMPANSupport', 'PowerManager', 'ShowConnected']
blueman-applet 15.15.32 INFO     PluginManager:159 __load_plugin: loading <class 'blueman.plugins.applet.DBusService.DBusService'>
blueman-applet 15.15.32 INFO     PluginManager:159 __load_plugin: loading <class 'blueman.plugins.applet.Menu.Menu'>
[...]

PR #2514, opened 2024-10-17
ManagerToolbar: Include bluetooth switch when updating sensitive

infirit:


cschramm (approved):

cschramm:
Now I'm wondering if that's correct. I have to check with my old Thinkpad. Using the killswitch removes the internal adapter there. If I'm not mistaken, this change makes the toggle insensitive then, while I could previously use it to unblock and get the device back.

infirit:
It should be insensitive on your thinkpad, let me know if you like to revert it.

The bigger picture is that PowerManager just switches internal power state as if bluetooth should be working when in reality there is no adapter. On your thinkpad that is a good thing as the KillSwitch will be flipped and the new adapter is updated with the existing adapter power state, which is on.

Fixing that is probably not trivial 😅.

cschramm:
What was the reason for this change? Is it in context of #2510?

Issue #2102, opened 2023-08-02
Disable blueman-applet in Gnome


ispanos:
blueman: blueman.x86_64 -- 1:2.3.5-3.fc38
BlueZ: bluez.x86_64 -- 5.68-1.fc38
Distribution: Fedora 38
Desktop environment: Gnome / Swaywm

I use Blueman in Sway and I start it in the config file. However, when using Gnome, blueman applet auto-starts and I can't find a way to disable it. In case it is relevant, here is the output of `systemctl --user list-units | grep blue`

```
  sys-devices-pci0000:00-0000:00:14.0-usb3-3\x2d10-3\x2d10:1.0-bluetooth-hci0-hci0:3585.device loaded active plugged /sys/devices/pci0000:00/0000:00:14.0/usb3/3-10/3-10:1.0/bluetooth/hci0/hci0:3585
  sys-devices-pci0000:00-0000:00:14.0-usb3-3\x2d10-3\x2d10:1.0-bluetooth-hci0.device           loaded active plugged /sys/devices/pci0000:00/0000:00:14.0/usb3/3-10/3-10:1.0/bluetooth/hci0
  sys-subsystem-bluetooth-devices-hci0.device                                                  loaded active plugged /sys/subsystem/bluetooth/devices/hci0
  sys-subsystem-bluetooth-devices-hci0:3585.device                                             loaded active plugged /sys/subsystem/bluetooth/devices/hci0:3585
  app-gnome-blueman-17533.scope                                                                loaded active running Application launched by gnome-session-binary
  bluetooth.target                                                                             loaded active active  Bluetooth
```

Thank you for your time.

P.S. I have [disabled bluetooth on boot](https://github.com/blueman-project/blueman/wiki/Troubleshooting#turn-on-bluetooth-device-on-boot) and I can't figure out how to enable using a key-bind. `bluetoothctl power on` works only after I first enable bluetooth using blueman or gnome settings. I would appreciate it if you could point me to the right direction.

cschramm:
Gnome seems to use an application called Tweaks to manage "Startup Applications". You should find and have an option to disable blueman there.

Could you elaborate on "`bluetoothctl power on` does not work"?

ispanos:
> You should find and have an option to disable blueman there.

It is not in tweaks nor in any of the directories for autostart (~/.config/autostart $XDG_CONFIG_HOME/autostart   $XDG_CONFIG_DIRS/autostart   /usr/share/gnome/autostart )

```
[...]

PR #2549, opened 2024-12-09
Hide blueman-applet in KDE and GNOME

mxaddict:
New versions of GNOME and KDE have dedicated UI for this

### DIFF
--- a/data/blueman.desktop.in
+++ b/data/blueman.desktop.in
@@ -5,4 +5,6 @@ Icon=blueman
 Exec=blueman-applet
 Terminal=false
 Type=Application
+NoDisplay=true
+NotShowIn=KDE;GNOME;
 Categories=


infirit:
That's a no. Many people use blueman on plasma and GNOME.

mxaddict:
I'm pretty sure if they are using blueman on KDE ang GNOME they are fully capable of starting the applet yes?

Would it not make more since to have this as opt-in rather than opt out?

Scenario:
1. Have KDE/GNOME installed alongside a window manager
1. Use `blueman` in window manager
1. Want to use Default KDE/GNOME UI for bluetooth
1. User has to disable blueman via /etc/xdg config

As a side note nm-applet already has this option set to make it opt in.

cschramm:
> New versions of GNOME and KDE have dedicated UI for this

For configuring autostart? Please elaborate. In #2102 we concluded that current Gnome does not have any proper mechanism for that anymore (at that time). That discussion was exactly around users switching between Gnome and another desktop and preferring Gnome Bluetooth in Gnome. We did not find a proper solution. The best idea was to implement our own autostart switch (creating and removing a link) and leave it to the user to somehow set things up for the desktops he wants to use blueman with. Not sure if that's actually any better than requiring him to override the autostart with his preferences.

PR #2552, opened 2024-12-11
Rewrite the SendTo dialog to use a Gtk.Listbox

infirit:
Rewrite the sendto device list to use the much simpler and flexible Gtk.ListBox.

The list has only one job now which is present a list of devices to the user. It doesn't manage any state for devices or adapters.

The list now also contains devices from all Adapters and filters based on the selection in the combobox. No more rebuilding the whole list when a different adapter is selected.

In the process I also changed the following:
* Remove the paired and trused icons as they imo serve little purpose.
* Add a warning icon when the device doesn't advertise it can handle ObjectPush.
* Changed the adapter selection combobox to also allow it to show all devices from all adapters.

<details>
<summary>Old description</summary>
This is a rough, mostly working, proof of concept to replace the TreeView in blueman-sendto we currently have with a much simpler and flexible Gtk.ListBox. Before I go further I'd like some feedback on the overall design of it. I do want to move the bluez.Manager and bluez.AnyDevice signal handling further up into apps/blueman-sendto as we already have them there anyway.

I do want to make the Gtk.ListBox and Gtk.ListBoxRow only be responsible for presenting the data. Not have it handle state on adapters/devices internally, this will be done through methods. Using properties for the row in blueman-manager may not be ideal so maybe I should use some sort of properties_update method and pass in all the changes in one go.

I like to treat this as a blueprint for when I start working on the Gtk.ListBox for blueman-manager so we can finally implement some of the designs done by users some time ago.

ps: I'm sure mypy will be unhappy with me but it's just a proof of concept for now.
</details>


cschramm:
I did not review this in detail, but it makes a lot of sense to me :ok_hand:

infirit:
Decided to remove the trusted and paired icons from the list as afaics they really serve no purpose when sending files over bluetooth. I added a new icon to indicate if the device advertises obex object push. Also added a togglebutton but not sure if this is useful (not happy with how it looks regardless).

image

infirit:
I'm mostly done with it and will undraft it soon. Here is a quick recording of the list.
[Screencast_20241231_020140.webm](https://github.com/user-attachments/assets/690227aa-38e1-4fa7-bc19-96190f1ed31d)

infirit:
I'm happy with this now. So ready for review and testing :-)

blueman/gui/DeviceSelectorDialog.py
```diff
@@ -1,56 +1,192 @@
...
+        if name is None:
+            raise ValueError("Invalid adapter")
+
+        for key in self._rows:
+            if self._rows[key].adapter_path == object_path:
+                row = self._rows.pop(key)
+                self._listbox.remove(row)
+
+        if name == self._adapter_combo.get_active_id():
+            self._adapter_combo.set_active_id(name)
```
cschramm, line 111: What's the intention here? :thinking:

cschramm (approved):

blueman/main/Sendto.py
```diff
@@ -96,6 +164,8 @@ def on_result(_sender: Sender, _res: bool) -> None:
 
         sender.connect("result", on_result)
 
+        return False
```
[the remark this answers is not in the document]
infirit: No idea :sweat_smile: . Changed it back to return `None`.

blueman/gui/DeviceSelectorDialog.py
infirit: making sure the combobox has something valid selected. I changed it to select `any` if the active adapter is removed.

infirit:
Fixed the comments. Also added a bug fix where discovery is not started for new adapter(s).

PR #2555, opened 2024-12-14
apps: parse log level with logging.getLevelName()

stokito:
The logging.getLevelName() function returns the numeric representation of logging level.

The args.LEVEL is "warning" by default.
Just to be in line with the Python let's make it in upper case "WARNING". But if previously an incorrect level name used e.g. CRIT then the WARNING fallback value used instead, but now it will fail. This shouldn't be a big problem.

### DIFF
--- a/apps/blueman-adapters.in
+++ b/apps/blueman-adapters.in
@@ -22,20 +22,7 @@ if __name__ == '__main__':
     parser.add_argument("--socket-id", dest="socket_id", action="store", type=int, metavar="ID")
     parser.add_argument("adapter", nargs="?", metavar="ADAPTER NAME")
     args = parser.parse_args()
-
-    if args.LEVEL.upper() == "DEBUG":
-        log_level = logging.DEBUG
-    elif args.LEVEL.upper() == "INFO":
-        log_level = logging.INFO
-    elif args.LEVEL.upper() == "WARNING":
-        log_level = logging.WARNING
-    elif args.LEVEL.upper() == "ERROR":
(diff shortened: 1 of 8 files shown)


infirit:
Unhandled Exceptions are not an option for me. They tend to get caught by reporting tools like abrt. And it only shows the user the application crashed as they don't read the backtrace.

This would be fine as well and informs the user properly what options they have.

```python
diff --git a/blueman/Functions.py b/blueman/Functions.py
index d8810de09..7100738b9 100644
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -269,8 +269,9 @@ def create_parser(
     if parser is None:
         parser = argparse.ArgumentParser()
 
+    debug_choices = ["warning", "debug", "info", "error", "critical"]
     if loglevel:
-        parser.add_argument("--loglevel", dest="LEVEL", default="warning")
+        parser.add_argument("--loglevel", dest="LEVEL", default="warning", choices=debug_choices)
 
     if syslog:
         parser.add_argument("--syslog", dest="syslog", action="store_true")
```

```
$ python apps/blueman-services --loglevel bad
usage: blueman-services [-h] [--loglevel {warning,debug,info,error,critical}] [--syslog]
blueman-services: error: argument --loglevel: invalid choice: 'bad' (choose from 'warning', 'debug', 'info', 'error', 'critical')
```

infirit:
> looks good 👍

Can you include the argparse choices in this PR then I'll approve this.

PR #2556, opened 2024-12-15
Improve error reporting when connecting to device

stokito:
I was confused when saw an `br-connection-unknown` error and didn't know where to start from.
At first I tried to enable debugging `blueman-manager --loglevel=debug` but this didn't helped a lot. For some reason the BlueZ doesn't return any details in the DBus call. Can you work with the devs and add more details to a message?
To make it easier for a user I added the message "Check bluetoothd logs." so that I will faster figure out to make the `journalctl -u bluetooth -f -n 400` and see _the real error_ there.
I made the logging message same as the message to a user, but not translated.


infirit:
Haven't looked too closely at the changes but, don't commit .po changes pls. We send an updated blueman.pot to Weblate for the translators. Then Weblate will take care of sending PR to update the .po files :-).

infirit:
It's all automated so just ignore everything related to translations.

PR #2566, opened 2024-12-17
Fix fatal LoadException

cschramm:
It looks like this rarely ever manifests, but we erroneously stopped catching the exception in https://github.com/blueman-project/blueman/commit/c71294c1cc1246e55da263a9d0014c94977716df.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+### Bugs fixed
+
+* Fatal LoadException
+
 ### Changes
 
 * Terminate applet on manager termination if it was started by manager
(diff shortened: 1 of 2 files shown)


infirit (approved):
It's not a common error but as it's fatal, should we also include it in 2.4.4?

cschramm:
Makes sense, yes.

I think this did not bite us yet because it is not reachable with affected versions of blueman alone. You either have to modify `/org/blueman/general/plugin-list`, e.g. to `['DhcpClient']`, or adopt such a value from blueman 2.0 which seems to be the case in [Debian #1082668](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1082668). Enabling a plugin in the plugin manager will automatically disable all conflicting ones, so that e.g. `DhcpClient` will not be set without `!NMPANSupport`. We do not reach the point where `NMPANSupport` would have to get unloaded as it never gets loaded then.

PR #2573, opened 2024-12-20
Make connection notifications transient

merlinz01:
This PR sets a timeout for connection notifications and sets the transient hint so that these notifications do not remain in the notification list.

This has been a bother for me, with my Bluetooth mouse connecting and disconnecting and piling up a list of irrelevant notifications.

### DIFF
--- a/blueman/gui/Notification.py
+++ b/blueman/gui/Notification.py
@@ -162,6 +162,12 @@ class _NotificationBubble(Gio.DBusProxy):
         self._timeout = timeout
         self._return_id = None
 
+        if self._timeout > 0:
+            try:
+                self.set_hint('transient', True)
+            except ValueError:
+                pass
+
         if icon_name:
             self._app_icon = icon_name
         elif image_data:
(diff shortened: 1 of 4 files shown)


infirit:
Which notification daemon is storing all these? for me on Plasma none of these show up on the list so I like a way to replicate the behavior you are seeing. (why plasma doesn't do this I'll have to investigate :-)).

A general note, we usually don't set a timeout so each daemon can use whatever their default is. I think making the connection notification transient is fine but I prefer this to be passes as an argument to `Notification` and not rely on the timeout.

merlinz01:
I'm on Linux Mint with the Cinnamon desktop. I don't know what notification daemon it uses.

I'm fine with not relying on the timeout. I'll add a `transient` parameter that's a no-op for the manually created notification windows.

infirit:
The eternal battle with the CI lol. Just some whitespace now, mypy is happy.

infirit (approved):

PR #2575, opened 2024-12-20
Revert "TransferService: small code optimization for on_transfer_comp…

### LINKED ISSUE #2574: File transfer: existing files are silently overwritten, file renaming failure
blueman: 2.3.5
BlueZ: 5.66
Distribution: Debian 12.8
Desktop environment: none

Steps to reproduce:
- Transfer a file from an external device
- Repeat the previous step with the same file (or with another file having the same file name)
- In the first notification popup message (`"Incoming file over Bluetooth"`) the original file name is displayed, e.g. `bluetooth_content_share.html`
- In the final popup message (`"File Received"`) the file name has a timestamp string prepended, e.g. `20241220140923_bluetooth_content_share.html`
- The blueman log says: `TransferService:294 _on_transfer_completed: Destination file exists, renaming to: 20241220140923_bluetooth_content_share.html`

Expected result: the existing file should be left alone and the new copy should be saved under the modified file name (timestamp prepended)

Actual result: the existing file is silently overwritten.
It looks like the safety renaming mentioned in the log doesn't take place actually

infirit:
This completely ignored the new filename we generated. fixes #2574

### DIFF
--- a/blueman/plugins/applet/TransferService.py
+++ b/blueman/plugins/applet/TransferService.py
@@ -288,12 +288,12 @@ class TransferService(AppletPlugin):
         dest_dir, ignored = self._make_share_path()
         filename = os.path.basename(src)
 
-        dest = os.path.join(dest_dir, filename)
-        if os.path.exists(dest):
+        if os.path.exists(os.path.join(dest_dir, filename)):
             now = datetime.now()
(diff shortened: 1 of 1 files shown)


cschramm (approved):
:facepalm: Stared at it a while before I even got the difference :sweat_smile:

PR #2582, opened 2024-12-25
Update AutoConnectConfig.py

UmairShafiqu3:
# Key Changes:
# 1. Corrected the import path for BtAddress to `blueman.bluez.BtAddress`. # 2. Converted the "services" key values to a list when setting them in `switch` to match Gio.Settings expectations. # 3. Grouped imports from `gi.repository` to a single line for simplicity. # 4. Ensured logical flow and type consistency for `data` and `services`. Use `list` instead of `set` when assigning to Gio.Settings values

### DIFF
--- a/blueman/config/AutoConnectConfig.py
+++ b/blueman/config/AutoConnectConfig.py
@@ -3,24 +3,27 @@ import gi
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk, Gio
+from typing import Tuple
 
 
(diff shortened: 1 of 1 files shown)


infirit:
I little confused with all of this.

1. blueman.bluez.BtAddress does not exist
2. The use of a set is intentional as duplicates should stored. Also, any iterable (list like) will be fine to store in GSettings.
3. It's a matter of taste and preference imo.
4. is identical to your point in 2.

If this was generated from AI please don't post this to blueman.

PR #2685, opened 2025-03-27
Manager: Hide bt status switch when PowerManager is not available

infirit:
This will hide the switch if the PowerManager plugin isn't loaded.


infirit:
> ../meson.build:31:9: ERROR: Dependency 'girepository-2.0' is required but not found.

This is new for the tests 🤔 . It's 1 in the morning so I'll look at that later this week.

cschramm:
[pygobject](https://pypi.org/project/PyGObject/) 3.52 switched to girepository-2.0, so we'd need to install `libgirepository-2.0-dev` instead of `libgirepository[-]1.0-dev`. Unfortunately, it is not available on 22.04. I'll give 24.04 a shot.

cschramm (approved):

PR #2691, opened 2025-04-04
Use GActions where possible

infirit:
This supersedes #1927. As mentioned in there, RadioMenuItems don't work with GAction in Gtk3.

The next step is to move the `Blueman`/`ManagerDeviceList` object out of `ToolBar` and `ManagerMenu` as both connect to the same `ManagerDeviceList` signals to do mostly the same thing, update the UI.


infirit:
* Converted GSettings binds to use actions.
* Partially converted sort-by to use an action.

infirit:
Re-based due to conflicts.

Issue #2722, opened 2025-05-05
Applet Dbus service unavailable when turning off multiple plugins at the same time


infirit:
This happens (for me) when multiple plugins are disabled in quick succession due to one being a dependency. I can trigger below when disabling PowerManager which also disables KillSwitch.

Not sure what the best approach is but we need to fix eventually.

```python
Traceback (most recent call last):
  File "/home/sander/repos/blueman/blueman/main/Manager.py", line 107, in on_applet_signal
    if "PowerManager" in self.Applet.QueryPlugins():
                         ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/gi/overrides/Gio.py", line 380, in __call__
    result = self.dbus_proxy.call_sync(self.method_name, arg_variant,
                                       flags, timeout, None)
gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.blueman.Applet” on object at path /org/blueman/Applet (19)
```

cschramm:
I think this is eventually due to a GDBus limitation: To remove methods and signals (those provided by `PowerManager` in that case), we remove the object and re-register it with the new (reduced) `DBusInterfaceInfo`. Unfortunately, there does not seem to be a way to modify that data for introspection without that step, even though it should just be internal state purely used for handling introspection calls.

infirit:
I was thinking to add the plugin name and it's state in the signal signature like below. Then anything interested can check and act accordingly to the state of the plugin.

```patch
diff --git a/blueman/plugins/applet/DBusService.py b/blueman/plugins/applet/DBusService.py
index e3ed45899..2bba74b2c 100644
--- a/blueman/plugins/applet/DBusService.py
+++ b/blueman/plugins/applet/DBusService.py
@@ -10,6 +10,7 @@ from blueman.Service import Service
 from blueman.bluez.errors import BluezDBusException
 if TYPE_CHECKING:
     from blueman.main.NetworkManager import NMConnectionError
+    from blueman.main.PluginManager import PluginManager
 from blueman.plugins.AppletPlugin import AppletPlugin
 from blueman.bluez.Device import Device
 from blueman.services.Functions import get_service
@@ -56,12 +57,12 @@ class DBusService(AppletPlugin):
         self._add_dbus_method("DisconnectService", ("o", "s", "d"), "", self._disconnect_service, is_async=True)
[...]

PR #2756, opened 2025-06-15
Appletplugin: Plugins provide dbus interfaces

infirit:
This moves the handling of plugin dbus services to the individual plugins. `DBusService` provides the basic services to mange the applet and it's plugins under `org.blueman.Applet`. This service should never be unloaded to avoid #2722.

Instead of tearing down `org.blueman.Applet` when a plugin is (un)loaded each plugin now provides it's own interface. Added convenience functions to the proxies to make this more manageable.

### DIFF
--- a/blueman/Functions.py
+++ b/blueman/Functions.py
@@ -39,7 +39,7 @@ import time
 import subprocess
 import cairo
 
-from blueman.main.DBusProxies import AppletService, DBusProxyFailed
+from blueman.main.DBusProxies import AppletService, AppletPowerManagerService, DBusProxyFailed
 from blueman.Constants import BIN_DIR, ICON_PATH, BLUETOOTHD_PATH
 
 import gi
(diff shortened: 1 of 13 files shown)


cschramm:
I might be missing something, but this seems incomplete. Just from looking at the changes, I wonder if:

* `DhcpClient` lacks D-Bus setup
* `check_bluetooth_status` tries to use `SetBluetoothStatus` and `GetBluetoothStatus` on `org.blueman.Applet`
* `BluetoothStatusChanged` is now on `org.blueman.Applet.PowerManager`, but expected on `org.blueman.Applet`

infirit:
You're not missing things, I completely overlooked DhcpClient :sweat:. I'll fix the other bits as well.

infirit:
Fixed all the issues, mypy and sonarcloud are happy as well.

cschramm (approved):

PR #2766, opened 2025-06-21
Use file monitor to handle changes in KillSwitch

infirit:
Technically mechanism plugin RfKill could also not be allowed to open the device but I'm uncertain how likely this is.

Also used an enum to represent the operation and some housekeeping.

### DIFF
--- a/blueman/bluez/errors.py
+++ b/blueman/bluez/errors.py
@@ -130,8 +130,6 @@ __DICT_ERROR__ = {'org.bluez.Error.Failed': DBusFailedError,
 
 
 def parse_dbus_error(exception: GLib.Error) -> BluezDBusException:
-    global __DICT_ERROR__
-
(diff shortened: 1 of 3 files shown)


infirit:
Oh, I don't have a device with a hardware killwitch and while I don't think it needs special handling it would be nice to confirm this doesn't break that.

cschramm (approved):

cschramm:
I'll test the hardware killswitch. Regarding flake8, you can simply remove that `global` statement. It's pointless.

cschramm:
:facepalm: I got confused there. My old Thinkpad doesn't have a hardware killswitch (in terms of something physical) either. What it does is to somehow disconnect the adapter when the switch is soft-blocked. At least that does not cause any issues.

cschramm (approved):

PR #2817, opened 2025-07-31
NetworkService: return not avaiable when device is blocked

infirit:
I am thinking we should extend this check to all services when a device is blocked.

### DIFF
--- a/blueman/services/meta/NetworkService.py
+++ b/blueman/services/meta/NetworkService.py
@@ -16,9 +16,9 @@ class NetworkService(Service):
 
     @property
     def available(self) -> bool:
-        # This interface is only available after pairing
-        paired: bool = self.device["Paired"]
-        return paired
+        # This interface is only available after pairing and blocked
+        available: bool = self.device["Paired"] and not self.device["Blocked"]
+        return available
 
     @property
     def connectable(self) -> bool:


infirit:
Alternative patch is below. Reasoning is that most services will not work if not paired and definitely not when blocked. But still keep it abstract to force sub-classes to handle either the default or do their own.


``` patch
From 913674531490495feff204b79b7114d6e0f17535 Mon Sep 17 00:00:00 2001
From: Sander Sweers <infirit@gmail.com>
Date: Thu, 31 Jul 2025 23:37:25 +0200
Subject: [PATCH] Service: return not avaiable when device is blocked

BlueZ removes the relevant dbus interface.

fixes: #2815
---
 blueman/Service.py                      | 4 +++-
 blueman/services/meta/NetworkService.py | 4 +---
 blueman/services/meta/SerialService.py  | 4 +---
 3 files changed, 5 insertions(+), 7 deletions(-)

diff --git a/blueman/Service.py b/blueman/Service.py
index c5f2c7602..3b24e96fb 100644
--- a/blueman/Service.py
+++ b/blueman/Service.py
@@ -66,7 +66,9 @@ class Service(ABC):
     @property
     @abstractmethod
     def available(self) -> bool:
-        ...
+        # Most services will not work if not paired or if they are blocked.
+        available: bool = self.device["Paired"] and not self.device["Blocked"]
+        return available
 
     @property
     @abstractmethod
diff --git a/blueman/services/meta/NetworkService.py b/blueman/services/meta/NetworkService.py
index e681f1434..6b1b5280d 100644
--- a/blueman/services/meta/NetworkService.py
+++ b/blueman/services/meta/NetworkService.py
@@ -16,9 +16,7 @@ class NetworkService(Service):
 
     @property
     def available(self) -> bool:
-        # This interface is only available after pairing
-        paired: bool = self.device["Paired"]
-        return paired
+        return super().available
 
     @property
     def connectable(self) -> bool:
diff --git a/blueman/services/meta/SerialService.py b/blueman/services/meta/SerialService.py
index 87d54f0f0..8dfe7d4d1 100644
--- a/blueman/services/meta/SerialService.py
+++ b/blueman/services/meta/SerialService.py
@@ -21,9 +21,7 @@ class SerialService(Service):
 
     @property
     def available(self) -> bool:
-        # It will ask to pair anyway so not make it available
-        paired: bool = self.device["Paired"]
-        return paired
+        return super().available
 
     @property
     def connectable(self) -> bool:
-- 
2.50.1
```

PR #2821, opened 2025-08-02
Handle blocked devices better (plus cleanups)

infirit:
This also contains the cleanups from #2820

I decided on having a default `available` but to keep it abstract to force each subclass to think about how to handle it.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -405,7 +405,7 @@ class ManagerDeviceMenu(Gtk.Menu):
             builder = Builder("rename-device.ui")
             dialog = builder.get_widget("dialog", Gtk.Dialog)
             dialog.set_transient_for(self.Blueman.window)
-            dialog.props.icon_name = "blueman"
+
             alias_entry = builder.get_widget("alias_entry", Gtk.Entry)
             alias_entry.set_text(device['Alias'])
             dialog.connect("response", on_response)
--- a/blueman/gui/manager/ManagerDeviceMenu.py
+++ b/blueman/gui/manager/ManagerDeviceMenu.py
@@ -303,12 +303,19 @@ class ManagerDeviceMenu(Gtk.Menu):
         show_generic_connect = self.show_generic_connect_calc(self.SelectedDevice['UUIDs'])
 
         powered = Adapter(obj_path=self.SelectedDevice["Adapter"])["Powered"]
+        blocked = self.SelectedDevice["Blocked"]
 
         if not row["connected"] and show_generic_connect and powered:
+            if blocked:
+                tooltip = _("Not available (Blocked)")
+            else:
+                tooltip = _("Connects auto connect profiles A2DP source, A2DP sink, and HID")
+
             connect_item = self._add_menu_item(
                 label=_("<b>_Connect</b>"),
                 icon_name="bluetooth-symbolic",
-                tooltip=_("Connects auto connect profiles A2DP source, A2DP sink, and HID")
+                tooltip=tooltip,
+                sensitive=not blocked
             )
             connect_item.connect("activate", lambda _item: self.connect_service(self.SelectedDevice))
         elif row["connected"] and show_generic_connect:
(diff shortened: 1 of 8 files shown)


infirit:
Rebased against main to solve merge conflicts.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -331,112 +346,90 @@ def generate(self) -> None:
...
+            send_item.connect("activate", lambda _: self.Blueman.send(self.SelectedDevice))
+
+            self._add_seperator()
+
+        self._add_menu_item(
+            label=_("_Pair"),
+            icon_name="blueman-pair-symbolic",
+            tooltip=_("Create pairing with the device") if not blocked else _("Not available (Blocked)"),
+            sensitive=False if blocked else not row["paired"]
+        )
```
cschramm, line 387: This lacks the action

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -256,15 +256,26 @@ def show_generic_connect_calc(self, device_uuids: Iterable[str]) -> bool:
...
+        self.append(sep_item)
+
+    def _add_menu_item(self, label: str, icon_name: str | None = None, sensitive: bool = True,
+                       tooltip: str | None = None) -> Gtk.ImageMenuItem:
+        item = create_menuitem(text=label, icon_name=icon_name)
+        item.set_sensitive(sensitive)
+        if tooltip is not None:
+            item.set_tooltip_text(tooltip)
+        self.append(item)
+        return item
```
cschramm, line 270: Seems to lack `item.show()` or `visible=true`

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -256,15 +256,26 @@ def show_generic_connect_calc(self, device_uuids: Iterable[str]) -> bool:
         # LE devices do not appear to expose certain properties like uuids until connect to at least once.
         return not device_uuids
 
+    def _add_seperator(self) -> None:
```
cschramm, line 259: Typo

cschramm (commented):
I don't think all those actions should be unavailable if a device is marked blocked. Many can still work, can't they?

blueman/gui/manager/ManagerDeviceMenu.py
infirit: Messed this up in the rebase, will fix.

infirit:
> I don't think all those actions should be unavailable if a device is marked blocked. Many can still work, can't they?

Unfortunately they fail. Connecting just ends up timing out so there is no point trying. Instead of not showing them I can make them not sensitive and give them an updated name/tooltip like I did for Connect.

Things like renaming and querying data that BlueZ stored on disk are fine.

NetworkService is particularly bad and breaks the ManagerDeviceMenu as interface org.bluez.Network1 is removed by bluez (Mr Edgy is my Motorola Edge phone).


```
blueman-manager 23.07.05 DEBUG    ManagerDeviceMenu:314 generate  : Mr Edgy
  File "/home/sander/repos/blueman/blueman/bluez/Base.py", line 100, in get
    prop = self.__proxy.call_sync(
        'org.freedesktop.DBus.Properties.Get',
    ...<2 lines>...
        GLib.MAXINT,
        None)
gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: No such interface 'org.bluez.Network1' (16)
```

cschramm:
I think it varies. What is rejected for a blocked device is when it tries to open an L2CAP channel. That is the case, e.g., if it acts as an A2DP source or hands-free device, but it should not be the case, e.g., if it's an A2SP sink or LE GATT is used.

Pairing a blocked device should always work, I guess, except that it might trigger an immediate, implicit connection attempt, which could fail, of course.

I am a bit worried that this interferes with valid use cases like blocking an overeager device in a "don't call me, I'll call you" manner. Your hypothesis seems to be that connecting will not be possible then, so that this case simply is not supported by BlueZ. If that holds (which I doubt), we also need to avoid auto-connect attempts (refs #3197 #3283)

PR #2832, opened 2025-08-09
Reformat with Cambalache

infirit:
Cambalache is the replacement for Glade which has been broken for me for a while now. It also improved a lot that now I would say it is very usable so let it reformat the exiting files.


data/ui/services-window.ui
```diff
@@ -1,43 +1,26 @@
...
-            </child>
-            <child type="tab">
-              <placeholder/>
-            </child>
-            <child>
-              <placeholder/>
-            </child>
-            <child type="tab">
-              <placeholder/>
-            </child>
```
cschramm, line 40: Is this fine?

cschramm (approved):

PR #2934, opened 2025-10-16
TransferService: Use Session Source to get correct adapter

infirit:
Ran into blueman using the first, but wrong in my case, adapter while investigating the TypeError issue I introduced.

Should I add the other properties [1] from the session while I'm at it?

[1] https://github.com/bluez/bluez/blob/master/doc/org.bluez.obex.Session.rst


cschramm:
> Should I add the other properties [1] from the session while I'm at it?

Probably not worth it, but I have no real opinion on that.

PR #2967, opened 2025-11-12
blueman-manager: When Adapter is not powered prevent device removal

infirit:
BlueZ reports the org.bluez.Error.NotReady


infirit:
Also add this to stable?

PR #3023, opened 2025-12-22
Rework UI for the Info plugin

infirit:
Draft for now but the major changes are:

- Properties and values are added, removed and updated dynamically
- It does not block the main window anymore
- Can have a window open for each device
- When including property names allow it to be copied to clipboard as json.

image

### DIFF
--- a/blueman/plugins/manager/Info.py
+++ b/blueman/plugins/manager/Info.py
@@ -1,130 +1,417 @@
+from collections.abc import Callable
 from gettext import gettext as _
 from typing import Any
-from collections.abc import Iterable, Callable
+from blueman.bluemantyping import ObjectPath
+from gi.repository import Gtk, Gdk, Gio, GObject
 
-from gi.repository import Gtk, Gdk
-
+import json
 import logging
-
-from gi.repository.GObject import GObject
 
 from blueman.Functions import create_menuitem
 from blueman.Sdp import ServiceUUID
-from blueman.bluez.Device import Device
+from blueman.bluez.Device import Device, AnyDevice
 from blueman.bluez.errors import BluezDBusException
 from blueman.gui.manager.ManagerDeviceMenu import MenuItemsProvider, ManagerDeviceMenu, DeviceMenuItem
+from blueman.main.Builder import Builder
 
 from blueman.plugins.ManagerPlugin import ManagerPlugin
 
+DevicePropertyTypes = str | int | bool | list[str]
+PropertyDataType = dict[str, str | list[str]]
 
-def show_info(device: Device, parent: Gtk.Window) -> None:
-    def format_boolean(x: bool) -> str:
-        return _('yes') if x else _('no')
 
-    def format_rssi(rssi: int) -> str:
-        if rssi in [0x99, 0x7f]:
-            return f'invalid (0x{rssi:02x})'
+def format_advflags(flags: list[int]) -> str:
+    return ", ".join([str(flag) for flag in flags])
+
+
+def format_boolean(x: bool) -> str:
+    return _('yes') if x else _('no')
+
+
+def format_rssi(rssi: int) -> str:
+    if rssi in (0x99, 0x7f):
+        return f'invalid (0x{rssi:02x})'
+    else:
+        return f'{rssi} dBm (0x{rssi:02x})'
+
+
+def format_class(cls: int) -> str:
+    return f"0x{cls:06x}"
+
+
+def format_appearance(appearance: int) -> str:
+    return f"0x{appearance:04x}"
+
+
+def data_to_plain_text(data: PropertyDataType, values_only: bool) -> str:
+    lines = []
+    for prop, value in data.items():
+        if isinstance(value, list):
+            line = ", ".join(value)
         else:
-            return f'{rssi} dBm (0x{rssi:02x})'
+            line = value
 
-    def format_uuids(uuids: Iterable[str]) -> str:
-        return "\n".join([uuid + ' ' + ServiceUUID(uuid).name for uuid in uuids])
+        if not values_only:
+            lines.append(f"{prop}, {line}")
+        else:
+            lines.append(line)
 
-    def format_advflags(flags: Iterable[bytes]) -> str:
-        return "\n".join([str(flag) for flag in flags])
+    return "\n".join(lines)
 
-    store = Gtk.ListStore(str, str)
-    view = Gtk.TreeView(model=store, headers_visible=False)
-    view_selection = view.get_selection()
-    view_selection.set_mode(Gtk.SelectionMode.MULTIPLE)
 
-    def on_accel_activated(_group: Gtk.AccelGroup, _dialog: GObject, key: int, _modifier: Gdk.ModifierType) -> bool:
-        if key != 99:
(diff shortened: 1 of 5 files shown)


cschramm:
@infirit: I think it would make sense to pick this up. Any specific reason this is still in draft?

infirit:
None anymore. No merge conflicts :partying_face:

blueman/plugins/manager/Info.py
```diff
@@ -1,130 +1,421 @@
...
+            info_window.add_item(item)
+        else:
+            prop_label, value_label = self.format_property(prop, value)
+            item.props.description = prop_label
+            item.props.value = value_label
+
+    def format_property(self, property_name: str, value: DevicePropertyTypes) -> tuple[str, str]:
+        _position, formatter = Properties[property_name]
+        if property_name == "UUIDs":
+            assert isinstance(value, str)
```
cschramm, line 362: Isn't this is a list?

blueman/plugins/manager/Info.py
```diff
@@ -1,130 +1,421 @@
...
+
+    def add_item(self, item: InfoItem) -> None:
+        """Insert InfoItem into the model"""
+        self.__items[item.props.description] = item
+        self.__model.insert_sorted(item, self.__model_sort_func)
+
+    def remove_item(self, item: InfoItem) -> None:
+        """Remove InfoItem from the model"""
+        found, position = self.__model.find(item)
+        if found is None:
```
cschramm: `found` should be `bool`, so `None` isn't possible and `false` should definitely be caught here, no?

data/ui/manager-info.ui
```diff
@@ -0,0 +1,102 @@
...
+          </packing>
+        </child>
+        <child>
+          <object class="GtkComboBoxText" id="cb_format">
+            <property name="active">0</property>
+            <property name="sensitive" bind-source="bt_values_only" bind-property="active" bind-flags="invert-boolean | sync-create"/>
+            <property name="visible">True</property>
+            <items>
+              <item id="plain" translatable="yes">Plain text</item>
+              <item translatable="yes"/>
```
cschramm: Is this intended?

data/ui/manager-info.ui
```diff
@@ -0,0 +1,102 @@
...
+<interface>
+  <!-- interface-name manager-info.ui -->
+  <requires lib="gtk+" version="3.24"/>
+  <object class="GtkGrid" id="info_grid">
+    <property name="border-width">3</property>
+    <property name="visible">True</property>
+    <child>
+      <object class="GtkLabel">
+        <property name="halign">start</property>
+        <property name="label">&lt;big&gt;Select the rows you want to copy&lt;/big&gt;</property>
```
cschramm:
```suggestion
        <property name="label" translatable="yes">&lt;big&gt;Select the rows you want to copy&lt;/big&gt;</property>
```

data/ui/manager-info.ui
```diff
@@ -0,0 +1,102 @@
...
+      <object class="GtkBox">
+        <property name="margin">5</property>
+        <property name="margin-end">3</property>
+        <property name="margin-start">3</property>
+        <property name="spacing">3</property>
+        <property name="valign">start</property>
+        <property name="visible">True</property>
+        <child>
+          <object class="GtkButton" id="bt_copy">
+            <property name="label">Copy </property>
```
cschramm:
```suggestion
            <property name="label" translatable="yes">Copy</property>
```

blueman/plugins/manager/Info.py
```diff
@@ -1,130 +1,421 @@
...
+    def __on_accel_activated(
+            self,
+            _group: Gtk.AccelGroup,
+            _gobject: GObject.Object,
+            key: int,
+            _modifier: Gdk.ModifierType
+    ) -> bool:
         if key != 99:
             logging.warning(f"Ignoring key {key}")
             return False
```
cschramm: Not possible

data/ui/manager-info.ui
```diff
@@ -0,0 +1,102 @@
...
+          </object>
+          <packing>
+            <property name="pack-type">end</property>
+            <property name="position">2</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkComboBoxText" id="cb_format">
+            <property name="active">0</property>
+            <property name="sensitive" bind-source="bt_values_only" bind-property="active" bind-flags="invert-boolean | sync-create"/>
```
cschramm, line 52: Enabling "Values only" while JSON is selected makes it impossible to switch the format. "Values only" shouldn't be available while JSON is active.

blueman/plugins/manager/Info.py
```diff
@@ -1,130 +1,421 @@
...
+    "LegacyPairing": (120, format_boolean),
+    "RSSI": (130, format_rssi),
+    "TxPower": (140, str),
+    "Connected": (150, format_boolean),
+    "UUIDs": (160, str),
+    "Modalias": (170, str),
+    "Adapter": (180, str),
+    "ManufacturerData": (190, str),
+    "ServiceData": (200, str),
+    "AdvertisingData": (210, format_advflags),
```
cschramm:
```suggestion
    "AdvertisingFlags": (210, format_advflags),
```
?

blueman/plugins/manager/Info.py
```diff
@@ -1,130 +1,421 @@
...
+        bt_values_only.connect("toggled", self.__on_values_only_toggled)
+
+        bt_copy = builder.get_widget("bt_copy", Gtk.Button)
+        bt_copy.connect("clicked", self.__on_copy_clicked)
+
+        accelgroup = Gtk.AccelGroup()
+        self.add_accel_group(accelgroup)
+
+        key, mod = Gtk.accelerator_parse("<Control>C")
+        accelgroup.connect(key, mod, Gtk.AccelFlags.MASK, self.__on_accel_activated)
```
cschramm: These flags do not make sense. I think it should just be 0. `MASK` is `VISIBLE`, `LOCKED` and a reserved bit. There is nothing to be visible here (no `AccelLabel`) and nothing to lock for  (no accel path, no `AccelMap`).

blueman/plugins/manager/Info.py
infirit: The new info window adds each UUID individually as an item to the list and that uses strings, see `_show_info` (line 395).

blueman/plugins/manager/Info.py
infirit: Good catch.

data/ui/manager-info.ui
infirit: Nope, will removed this emtpy item.

data/ui/manager-info.ui
infirit: With values only it doesn't make sense to send it as JSON. I will force the combobox to be plain if values only is toggled. Another good catch.

blueman/plugins/manager/Info.py
infirit: Not sure I understand. Is the issue it will always be 99 so pointless to check for it?

infirit: I removed the check.

cschramm (approved):

PR #3132, opened 2026-02-22
Rename help to about and move the menuitem in manager to adapter menu

infirit:
Should the about dialog go somewhere else?

Also, renamed help to about in the applet menu.


cschramm (approved):
Thanks!

> Should the about dialog go somewhere else?

It does not make much sense in the Adapter menu, but I don't have any better idea for it...

PR #3142, opened 2026-03-03
dbus: prefer closures2

### LINKED ISSUE #3101: Memory leak: 1.5GB/week
Over the course of about a week, I see the bluman-applet memory consumption grow to about 1.5GB.

I'd like to track down the cause, but I'm not really sure where to start. As far as I can tell, the Python memory profiling tools are intended to measure excessive allocations, not leaks. And I don't think there's anything like Java's VisualVM, that will all you to take memory snapshots and inspect them. Advice would be appreciated!

flaviut:
`register_object_with_closures` is deprecated in favor of `register_object_with_closures2`, since the original leaks memory or otherwise requires some extra work to avoid leaking memory, I don't fully understand:

> [Deprecated in favour of g_dbus_connection_register_object_with_closures2(), which has more binding-friendly reference counting semantics.](https://docs.gtk.org/gio/method.DBusConnection.register_object_with_closures.html)

Keep getattr fallback so older GLib runtimes still work.

Fixes #3101

Testing:

I've been running this patch locally for 8 days with no issues. I've also got [this minimal reproducer](https://github.com/flaviut/blueman/commit/d3fbcbc0c9b01a68672ed69f60044cdbbb9dc00c), which quickly shows the fix works:

closure_leak_comparison

### DIFF
--- a/blueman/main/DbusService.py
+++ b/blueman/main/DbusService.py
@@ -105,7 +105,10 @@ class DbusService:
 
         node_info = Gio.DBusNodeInfo.new_for_xml(node_xml)
 
-        regid = self._bus.register_object(
+        # register_object_with_closures2 requires GLib 2.84.0 (released 2025-03-06).
+        # Keep a fallback for older runtimes that only provide register_object().
+        register_object = getattr(self._bus, "register_object_with_closures2", self._bus.register_object)
+        regid = register_object(
             self._path,
             node_info.interfaces[0],
             self._handle_method_call,


infirit:
There is more information in the commit that introduced version 2. Preferably id like pygobject to add an override for this so it just works transparently for all projects.

But I think having this in blueman is OK as it's simple enough. Ill post some links to relevant discussion later.

Commit adding v2 with context: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4431
Reverted PR https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/219

infirit (approved):

Issue #3171, opened 2026-03-16
Blueman-applet fails to start with object has no attribute error.


lidistat67:
Blueman-applet fails to launch with the following error:

```python
Traceback (most recent call last):
  File "/usr/bin/blueman-applet", line 39, in <module>
    app = BluemanApplet()
  File "/usr/lib/python3.14/site-packages/blueman/main/Applet.py", line 34, in __init__
    s = GLib.unix_signal_source_new(signal.SIGINT)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/gi/overrides/__init__.py", line 31, in __getattr__
    return getattr(self._introspection_module, name)
  File "/usr/lib/python3.14/site-packages/gi/module.py", line 130, in __getattr__
    raise AttributeError("%r object has no attribute %r" % (
                         self.__name__, name))
AttributeError: 'gi.repository.GLib' object has no attribute 'unix_signal_source_new'
```

Blueman version: 2.4.6
Glib version: 2.84.4

infirit:
Please include the upstream bug report if there is one (https://bugs.gentoo.org/971271).

`unix_signal_source_new` is in `GlibUnix` so make sure you have the relevant introspection files. 

We should probably make sure it's there with `gi.require_version("GLibUnix", "2.0")`.

PR #3180, opened 2026-03-23
Use GLibUnix.signal_source_new

SuperHeron:
Fix to be compatible with newest GLib.

This completes https://github.com/blueman-project/blueman/pull/3173 and fully fixes https://github.com/blueman-project/blueman/issues/3171.

### DIFF
--- a/blueman/main/Adapter.py
+++ b/blueman/main/Adapter.py
@@ -13,8 +13,9 @@ from blueman.main.Builder import Builder
 import gi
 gi.require_version("Gtk", "3.0")
 gi.require_version("Gdk", "3.0")
+gi.require_version("GLibUnix", "2.0")
 gi.require_version("Pango", "1.0")
-from gi.repository import Gtk, Gio, Gdk, GLib
+from gi.repository import Gtk, Gio, Gdk, GLibUnix
 from gi.repository import Pango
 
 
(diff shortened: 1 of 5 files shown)


infirit:
Was there an APIs change in pygobject that disallows the use of the unix only functions like we are doing?

And for context, the user from the issue had build GLib without introspection support which is a hard runtime dependency for blueman.

SuperHeron:
I use:
- GLib 2.88.0 (with GObject-Introspection)
- GObject-Introspection 1.86.0
- PyGObject 3.50.1
- Python 3.13.12

All I know is I need this change to make blueman work on my system. Otherwise, I keep getting the same error messages.
If there is a breaking change, chances are it is in one of these packages (or some obscure combination).
Or maybe it is due to the Python version I use like the Gentoo bug implies.

infirit:
There is a change [1] in pygobject 3.55.3 but it adds an override to handle the split gracefully.

Your issue is that the version of pygobject you're using is too old to handle the split in GLib.

I think importing GLibUnix should work on older GLib versions. If so your changes will be fine but we will need to split the typehints. Ill poke at it over the weekend.

1 https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/492

infirit:
Thanks for the PR. I cherrypicked your commit and added it to #3183 which fixes the typehints.

PR #3183, opened 2026-03-25
Use GLibUnix directly

infirit:
Moved the stubs and applied @SuperHeron changes from #3180

Let's hope mypy is happy :pray: 

GLib merge that split things off https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4881

### DIFF
--- a/blueman/main/Adapter.py
+++ b/blueman/main/Adapter.py
@@ -13,8 +13,9 @@ from blueman.main.Builder import Builder
 import gi
 gi.require_version("Gtk", "3.0")
 gi.require_version("Gdk", "3.0")
+gi.require_version("GLibUnix", "2.0")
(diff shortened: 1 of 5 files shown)


infirit:
It shouldn't as GLibUnix was always there. But I'll check current debian and the version before to be sure.

infirit:
Good you questioning the min version as I was wrong. The situation is messy at best. https://bugs-devel.debian.org/cgi-bin/bugreport.cgi?bug=1115340

tldr this will not work on bookworm, in trixie and later (>=GLib 2.80) it will.

GLib 2.88 removed the duplication and PyGObject 3.55.3 adds an override to handle the transition.

Honestly I am inclined to revert all the GLibUnix bits and have distribution handle this.

PR #3204, opened 2026-04-14
AutoConnect: Log failed auto-connect attempts

gfrmin:
The error callback in `AutoConnect._run` currently swallows failures silently with `pass`, so users get no feedback when the 60-second auto-connect loop can't reach a configured device.

Log the failure at warning level with the device name and reason so it surfaces in the journal/console.


infirit:
@cschramm this looks fine to me but I don't remember if there was a reason this wasn't logged before. Maybe to spammy as it's every 60(ish) seconds?

PR #3205, opened 2026-04-14
ManagerDeviceList: Guard double-click against duplicate operations

gfrmin:
A quick second double-click on a device row while a connect/disconnect is still in flight currently fires a second `connect_service`/`disconnect_service` call. This can lead to duplicate BlueZ requests and confusing UI state.

Check `menu.get_op(device)` before initiating, and skip if an operation is already tracked for that device.

### DIFF
--- a/blueman/gui/manager/ManagerDeviceList.py
+++ b/blueman/gui/manager/ManagerDeviceList.py
@@ -242,6 +242,8 @@ class ManagerDeviceList(DeviceList):
 
         if event.type == Gdk.EventType._2BUTTON_PRESS and cast(Gdk.EventButton, event).button == 1:
             if self.menu.show_generic_connect_calc(row["device"]['UUIDs']):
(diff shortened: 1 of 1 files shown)


blueman/gui/manager/ManagerDeviceList.py
```diff
@@ -242,6 +242,8 @@ def _on_event_clicked(self, _widget: Gtk.Widget, event: Gdk.Event) -> bool:
 
         if event.type == Gdk.EventType._2BUTTON_PRESS and cast(Gdk.EventButton, event).button == 1:
             if self.menu.show_generic_connect_calc(row["device"]['UUIDs']):
+                if self.menu.get_op(row["device"]):
```
infirit: We can check this as soon as we have a valid row so please move this to earlier in the function. This will most likely also make SonarCloud happy.

infirit (changes requested):

PR #3206, opened 2026-04-14
errors: Handle malformed messages in parse_dbus_error

gfrmin:
`parse_dbus_error` assumes the exception message always has the form `gerror:dbus_error:message` and does an unchecked `split(':', 2)` unpack. If BlueZ (or a proxy/glib path) ever produces a message with fewer colons, this raises `ValueError` from inside the error handler itself, masking the original failure.

Wrap the split in try/except and fall back to returning a plain `BluezDBusException` with the raw message.


infirit (approved):
The mypy errors are unrelated so feel free to ignore them until we fixed them in main.

PR #3209, opened 2026-04-16
blueman-applet: serialize RegisterAgent and RequestDefaultAgent

jkuro-tii:
Fix race condition where RequestDefaultAgent is sent before RegisterAgent completes in async D-Bus environments (proxies/VMs), causing org.bluez.Error.DoesNotExist.

blueman-applet issues RegisterAgent and RequestDefaultAgent as two independent asynchronous D-Bus calls. In environments where D-Bus message processing or object export is not immediate (e.g. proxies, bridges, containers), this can lead to a race condition where RequestDefaultAgent is executed before the agent is fully registered, causing org.bluez.Error.DoesNotExist.

Fix this by ensuring RequestDefaultAgent is only called after completion of RegisterAgent, restoring correct ordering semantics required by BlueZ AgentManager1 API.

### DIFF
--- a/blueman/bluez/AgentManager.py
+++ b/blueman/bluez/AgentManager.py
@@ -13,7 +13,7 @@ class AgentManager(Base):
 
     def register_agent(self, agent_path: str, capability: str = "", default: bool = False) -> None:
         param = GLib.Variant('(os)', (agent_path, capability))
-        self._call('RegisterAgent', param)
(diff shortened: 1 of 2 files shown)


infirit:
Thanks for the report. I prefer to keep this async so instead in #3213 blueman request to be default agent when we get a dbus reply.

PR #3230, opened 2026-05-06
Fix typo in D-Bus activation error message

psypherium:
source string is read-only on weblate, so editing here.

Edit, there is some other master "English" file somewhere which needs to be changed instead of this one.


infirit:
The master language is American English (en_US). There is no translation file as this is the source code [here](https://github.com/blueman-project/blueman/blob/9cb3c1179c823d5fed719e221666b20d8d29bc69/blueman/main/DBusProxies.py#L164).

After the changes are merged weblate will get updated automatically and people can make the changes to the translations.

The change itself looks fine to me.

PR #3242, opened 2026-05-13
bluez.Device: Move fallbacks here from Base

infirit:
The fallbacks are for org.bluez.Device1 not everyting based on Base.

### DIFF
--- a/blueman/bluez/Base.py
+++ b/blueman/bluez/Base.py
@@ -37,6 +37,7 @@ class Base(GObject.Object, metaclass=BaseMeta):
     __instances__: dict[str, "Base"]
 
     _interface_name: str
+    _fallbacks: dict[str, Any] = {}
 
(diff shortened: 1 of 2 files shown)


infirit:
I disagree with sonarcloud here. This is an error and having the exception logged helps.

infirit:
Rebased and CI is clean.

blueman/bluez/Base.py
```diff
@@ -108,8 +107,8 @@ def get(self, name: str) -> Any:
             property = self.__proxy.get_cached_property(name)
             if property is not None:
                 return property.unpack()
-            elif name in self.__fallback:
-                return self.__fallback[name]
+            elif self._fallbacks is not None and name in self._fallbacks:
```
cschramm, line 110: `is not None` is redundant

```suggestion
            elif name in self._fallbacks:
```

PR #3244, opened 2026-05-14
Implement restoring last power state of bluetooth adapter(s)

infirit:
Off by default but I think this is reasonable. Bluedevil has a similar feature.

### DIFF
--- a/blueman/plugins/applet/PowerManager.py
+++ b/blueman/plugins/applet/PowerManager.py
@@ -75,10 +75,6 @@ class PowerManager(AppletPlugin, StatusIconProvider):
     def on_unload(self) -> None:
         self.parent.Plugins.Menu.unregister(self)
 
-    @property
-    def CurrentState(self) -> bool:
-        return self.current_state
-
     def on_manager_state_changed(self, state: bool) -> None:
         if state:
             def timeout() -> bool:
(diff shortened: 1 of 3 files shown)


infirit:
Sonar cloud was right to complain with the old error message but I disagree with it to use `logging.exception` here.

data/org.blueman.gschema.xml
```diff
@@ -200,4 +200,12 @@
...
     </key>
   </schema>
+  <schema id="org.blueman.plugins.powermanager" path="/org/blueman/plugins/powermanager/">
+    <key type="b" name="last-power-state">
+      <default>false</default>
+    </key>
+    <key type="mb" name="restore-power-state">
+      <default>nothing</default>
+    </key>
+</schema>
```
cschramm, line 210:
```suggestion
  </schema>
```

blueman/plugins/applet/PowerManager.py
```diff
@@ -35,6 +35,24 @@ class PowerManager(AppletPlugin, StatusIconProvider):
...
+        "path": None
+    }
+
+    __options__ = {
+        "last-power-state": {
+            "type": bool,
+            "default": False
+        },
+        "restore-power-state": {
+            "type": bool,
```
cschramm, line 49: This does not match `mb`. I think `get_option` might actually return `None` and `set_option` will just never write `None`. It probably makes sense to just use `b` with a `false` default.

cschramm (approved):
I am not really convinced if this should be blueman's business, but it's definitely a nice option.

PR #3252, opened 2026-05-17
Update passkey message instead of closing and recreating notification

coyotebush:
Closing and recreating the notification was particularly problematic in the context of [the Qubes OS notification proxy not supporting CloseNotification](https://github.com/QubesOS/qubes-issues/issues/10862), but even with just xfce4-notifyd or the fallback dialog, it causes flickering. By now the notification code now supports `set_message` for a better approach. But only update a previous passkey notification and not potentially a notification that has `actions`.

### DIFF
--- a/blueman/main/applet/BluezAgent.py
+++ b/blueman/main/applet/BluezAgent.py
@@ -49,6 +49,7 @@ class BluezAgent(DbusService):
         self._db: ElementTree.ElementTree | None = None
         self._devhandlerids: dict[str, int] = {}
         self._notification: _NotificationBubble | _NotificationDialog | None = None
+        self._passkey_notification: _NotificationBubble | _NotificationDialog | None = None
         self._service_notifications: list[_NotificationBubble | _NotificationDialog] = []
 
     def register_agent(self) -> None:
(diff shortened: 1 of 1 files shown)


infirit:
I don't think we need a dedicated notification for this. Does https://github.com/blueman-project/blueman/pull/3254/changes/447102b7ac7621c76db33ec705c6077cfba89d80 work for you? I tested this on plasma and it works fine there.

coyotebush:
I tried that first. The only issue would be that if `self._notification` was most recently assigned in `_on_request_confirmation`, it would retain action buttons. Presumably in practice bluez will invoke Cancel (or connect the device) in between, but relying on that feels a little fragile. The [original Qubes OS bug report](https://github.com/QubesOS/qubes-issues/issues/10862#issuecomment-4323297651) would have reached that state due to CloseNotification raising an exception, and I can certainly also reach that state with manual `dbus-send` calls.

infirit:
I see we don't actually check if the notification daemon supports actions in `_on_request_confirmation` which is a bug which we need to fix.

 I assume your notification daemon doesn't support actions. Do you advertise you support actions?

edit: Although I'm not sure if we can then use a notification for `_on_request_confirmation` if actions are not supported.

coyotebush:
The Qubes proxy supports actions and advertises such. And Notification.py would fall back to a dialog otherwise, which I suppose would appear even stranger if reused for a passkey.
https://github.com/blueman-project/blueman/blob/9cb3c1179c823d5fed719e221666b20d8d29bc69/blueman/gui/Notification.py#L296-L301

PR #3288, opened 2026-06-02
fix: route DHCP renew through AppletDhcpClientService proxy

geraldo-netto:
NetworkService.common_actions called AppletService().dchp_client(), but the dchp_client method lives on the org.blueman.Applet.DhcpClient interface, exposed by AppletDhcpClientService — AppletService has no such method, so the "Renew IP Address" action raised AttributeError at call time. This also gives AppletDhcpClientService a real production call site.

Add unit + fuzz coverage:
- AppletDhcpClientService.dchp_client dispatches the correct method name and object-path variant; valid paths forwarded verbatim, malformed object paths rejected by GLib before any D-Bus call.
- NetworkService renew action dispatches to AppletDhcpClientService with the device object path; regression guard that AppletService is not referenced.
- Guard that AppletStatusIconService remains a ProxyBase subclass: it is a signal-only proxy required in Tray.py to receive StatusIcon-interface signals, so it must not be removed.

Tests avoid dbusmock so they run without a live bus.

### DIFF
--- a/test/services/meta/test_network_service.py
+++ b/test/services/meta/test_network_service.py
@@ -3,7 +3,6 @@ from unittest import TestCase
 from unittest.mock import patch, MagicMock
 
 from blueman.services.meta.NetworkService import NetworkService
-from blueman.Service import Action
 
 # The package re-exports the NetworkService *class* under the name
 # ``blueman.services.meta.NetworkService``, shadowing the submodule. Grab the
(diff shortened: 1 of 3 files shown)


test/services/meta/test_network_service.py
```diff
@@ -0,0 +1,62 @@
...
+            action.callback()
+
+        proxy_cls.assert_called_once_with()
+        proxy_cls.return_value.dchp_client.assert_called_once_with(path)
+
+    def test_renew_does_not_use_bare_applet_service(self):
+        # Regression: must route through AppletDhcpClientService, which is the
+        # proxy that actually owns the dchp_client method.
+        self.assertTrue(hasattr(network_service_module, "AppletDhcpClientService"))
+        self.assertFalse(hasattr(network_service_module, "AppletService"))
```
cschramm: This tests the import lines.

cschramm (commented):
The tests do not seem really useful to me. I'd probably pick just `test_renew_dispatches_to_dhcp_client_proxy` and `test_dhcp_client_dispatches_to_dbus`. Please strip that down.

test/services/meta/test_network_service.py
geraldo-netto: @cschramm , I hope I fixed it, I was a bit aggressive and asked the agent to cover 100% both test and fuzz
if you have any specific guideline for test strategy and coverage, please, let me know, I'll improve in the upcoming batches

cschramm (approved):

PR #3289, opened 2026-06-02
perf: bound SpeedCalc sample log and guard zero elapsed time

geraldo-netto:
SpeedCalc.calc() only dropped a single stale sample per call, so under high or irregular sampling rates the log could retain far more samples than the moving-average window needs. Drop all out-of-window samples each call (keeping at least two), bounding the log to the window regardless of sample rate.

Also guard against a zero (or negative) elapsed time between the retained samples, which previously raised ZeroDivisionError when two samples landed on the same rounded timestamp.

Add unit + fuzz coverage: reference handling, two-sample speed, recent-window averaging, log-length bounding under high-rate and irregular-gap streams, zero-elapsed safety, reset, and a fuzz stream asserting finite, non-raising results. Tests mock time.time() and need no D-Bus or display.

### DIFF
--- a/blueman/main/SpeedCalc.py
+++ b/blueman/main/SpeedCalc.py
@@ -1,10 +1,16 @@
+import math
 import time
+from collections import deque
+
+
+SAMPLE_RESOLUTION = 0.01
 
 
 class SpeedCalc:
     def __init__(self, moving_avg: float = 3.0) -> None:
         self.moving_avg = moving_avg
-        self.log: list[tuple[float, float]] = []
+        max_log_length = max(2, math.ceil(self.moving_avg / SAMPLE_RESOLUTION) + 2)
+        self.log: deque[tuple[float, float]] = deque(maxlen=max_log_length)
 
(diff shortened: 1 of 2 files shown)


blueman/main/SpeedCalc.py
```diff
@@ -14,23 +20,25 @@ def calc(self, amount: float) -> float:
...
-                total_amount = self.log[-1][1] - self.log[0][1]
-                speed = total_amount / total_time
-                return speed
-
-        else:
+
+        # Drop samples that fall outside the moving-average window, always
+        # keeping at least two so a speed can still be computed. This bounds
+        # the log length to the window regardless of the sampling rate (the
+        # previous code only dropped a single sample per call).
```
cschramm: Especially the reference to "previous code" sounds like something that shouldn't end up in the source.

geraldo-netto: fixed, thanks

cschramm (approved):

PR #3290, opened 2026-06-02
security: validate network inputs before building shell/AT commands

geraldo-netto:
Harden three places where externally-influenced values were interpolated into command arguments without validation:

- NetConf: validate that the IPv4 address and netmask passed to apply_settings are plain IPv4 addresses before they reach ip/ifconfig/iptables, and build iptables rules from already-split argument lists instead of splitting a single string on spaces. A value containing a space can no longer inject extra iptables arguments.
- Rfcomm: the CloseRFCOMM handler parsed `ps` output with an unguarded split()/int(), so a malformed or unexpected line raised ValueError and aborted the mechanism. Skip rows that lack a numeric pid and a command.
- PPPConnection: validate the APN against the 3GPP label charset before embedding it in the quoted AT+CGDCONT command, preventing break-out of the quoted argument.

Add unit + fuzz coverage for each: IPv4 validation (valid/invalid/injection), iptables argument construction (no splitting, delete symmetry), ps-output parsing across malformed inputs, and APN acceptance/rejection with fuzzed payloads. Tests need no D-Bus or display.

### DIFF
--- a/blueman/main/PPPConnection.py
+++ b/blueman/main/PPPConnection.py
@@ -50,8 +50,15 @@ class PPPConnection(GObject.GObject):
         'error-occurred': (GObject.SignalFlags.NO_HOOKS, None, (str,))
     }
 
+    # 3GPP APN labels are alphanumerics separated by dots/dashes (TS 23.003).
+    # Anything else could break out of the quoted AT+CGDCONT argument below.
+    _APN_RE = re.compile(r"^[A-Za-z0-9.-]*$")
+
     def __init__(self, port: str, number: str = "*99#", apn: str = "", user: str = "", pwd: str = "") -> None:
         super().__init__()
+
+        if not self._APN_RE.match(apn):
+            raise PPPException(f"Invalid APN: {apn!r}")
 
         self.apn = apn
         self.number = number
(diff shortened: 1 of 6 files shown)


blueman/main/PPPConnection.py
```diff
@@ -50,9 +50,16 @@ class PPPConnection(GObject.GObject):
         'error-occurred': (GObject.SignalFlags.NO_HOOKS, None, (str,))
     }
 
+    # 3GPP APN labels are alphanumerics separated by dots/dashes (TS 23.003).
+    # Anything else could break out of the quoted AT+CGDCONT argument below.
+    _APN_RE = re.compile(r"^[A-Za-z0-9.-]*$")
```
cschramm:
```suggestion
    _APN_RE = re.compile(r"[A-Za-z0-9.-]*\Z")
```

Otherwise, a trailing line-break is a perfect break out, isn't it? (Not covered by the tests)

geraldo-netto: Hi @cschramm , I have applied your suggestion and added some more cases to the tests which hopefully would cover most if not all the regex cases

geraldo-netto: I just received the notification it couldn't build, I'll fix it and get back to you

cschramm (approved):

PR #3291, opened 2026-06-02
reliability: guard crash paths in DHCP/network teardown

geraldo-netto:
Several teardown and error paths could raise before doing their work:

- DhcpClient: self._client was undefined until run(), so _check_client() and _on_timeout() raised AttributeError if reached early. Initialise it to None and bail out when there is no client. Also fix the timeout check: poll() returning 0 (a clean exit) was treated like "still running" via `not poll()`; only terminate when poll() is None.
- mechanism Network: EnableNetwork indexed DHCPDHANDLERS with the caller's string, raising KeyError on an unknown handler. Validate the key and raise a clear ValueError before touching NetConf.
- NetConf: DHCPHandler.clean_up() used next() without a default, so an empty match raised StopIteration and aborted the rest of cleanup. Supply a default.
- Rfcomm: OpenRFCOMM now logs and re-raises if spawning the watcher fails instead of failing opaquely.

Add unit + fuzz coverage for each guard. Tests need no D-Bus or display.

### DIFF
--- a/blueman/main/DhcpClient.py
+++ b/blueman/main/DhcpClient.py
@@ -51,9 +51,6 @@ class DhcpClient(GObject.GObject):
         GLib.timeout_add(self._timeout * 1000, self._on_timeout)
 
     def _on_timeout(self) -> bool:
-        # poll() returns None only while the client is still running; an exit
-        # code of 0 previously also matched `not poll()` and wrongly triggered a
-        # terminate of an already-finished client.
         if self._client is not None and self._client.poll() is None:
             logging.warning("Timeout reached, terminating DHCP client")
             self._client.terminate()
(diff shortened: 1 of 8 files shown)


blueman/main/DhcpClient.py
```diff
@@ -50,12 +51,18 @@ def run(self) -> None:
         GLib.timeout_add(self._timeout * 1000, self._on_timeout)
 
     def _on_timeout(self) -> bool:
-        if not self._client.poll():
+        # poll() returns None only while the client is still running; an exit
+        # code of 0 previously also matched `not poll()` and wrongly triggered a
```
cschramm: Describes a change. Shouldn't end up in source

geraldo-netto:
Hi @cschramm , I have updated the PR, could you please review and share your thoughts?

PR #3297, opened 2026-06-10
Make the fallback dialog less annoying

infirit:
Whether we should close automatically is still an open question #3257.

* Add a close button (with accelerator) if no actions were provided, People can now easily close it with a mouse or keyboard shortcut.
* It drops the fade in and out (broken) effect as imo it doesn't really add much but makes it hard to see/read.
* Actions are reworked and are a NamedTuple which makes them much nicer to deal with.
* Actions are now handled (mostly) the same between the bubble and dialog.

### DIFF
--- a/blueman/gui/Notification.py
+++ b/blueman/gui/Notification.py
@@ -51,7 +51,6 @@ class _NotificationDialog(Gtk.MessageDialog):
         elif image_data:
             self.set_icon_from_pixbuf(image_data)
 
-        self.connect("response", self.dialog_response)
         self.props.icon_name = "blueman"
 
     def set_message(self, message: str) -> None:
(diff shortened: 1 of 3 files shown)


blueman/gui/Notification.py
```diff
@@ -1,57 +1,43 @@
...
-                action_name = a[1]
+        self.actions: dict[int, NotificationAction] = {}
 
-                self.actions[i] = action_id
-                self.add_button(action_name, i)
-                i += 1
+        if actions is None:
+            actions = [NotificationAction("close", _("_Close"), lambda s: self.close())]
 
-        self.actions[Gtk.ResponseType.DELETE_EVENT] = "close"
```
cschramm, line 48: This is fired by the window manager, e.g., in reaction to Esc or a request to close the window. I think those controls are broken now. Also, `actions_cb` fired for it (and any other event), and at least `BlueAgent.py`'s `on_auth_action` relies on that.

blueman/gui/Notification.py
```diff
@@ -1,57 +1,43 @@
...
-            for a in actions:
-                action_id = a[0]
-                action_name = a[1]
+        self.actions: dict[int, NotificationAction] = {}
 
-                self.actions[i] = action_id
-                self.add_button(action_name, i)
-                i += 1
+        if actions is None:
+            actions = [NotificationAction("close", _("_Close"), lambda s: self.close())]
```
cschramm, line 33: `self.close()` should be redundant here, as `do_response` already destroys.

PR #3306, opened 2026-06-15
perf: manager device list

geraldo-netto:
## Summary

- Batch manager device row setup/update property reads using `get_properties()` snapshots.
- Replace per-device power-level timers with one shared timer over monitored devices, driven by the cached connected state instead of a live D-Bus read per device per second.
- Cache device UUIDs, class, address and icon name in manager row data and use the cached state for drag motion, drag-and-drop send, double-click connect handling, Ctrl+C address copy, the visibility filter, tooltip reads and menu generation.
- Share a single `GetAll` snapshot when adding device rows (`DeviceList.add_device`, `DeviceSelectorList`) instead of three to four round trips per discovered device.
- Batch `GenericList.set` column writes into a single `row-changed` emission, so the sort machinery and visibility filter run once per update instead of once per column.
- Cache signal bar pixbufs per (bar, level, scale) instead of decoding the PNG on every level change, and skip the D-Bus battery read used only for a debug log line when debug logging is disabled.
- Add benchmark script, expand unit test coverage and add deterministic fuzz-style tests.

Measured (10000 iterations, DBUS_RTT=200us, /usr/bin/python3):

| Metric | [main](https://github.com/geraldo-netto/blueman/commit/fb4189074193a7bfe72f00a79e8c37761f2e6a2d) (`fb41890`) | [branch](https://github.com/geraldo-netto/blueman/commit/24cdca00c6055d207822f8addbd4fb4300f1bf20) (`24cdca0`) | Gain |
| --- | --- | --- | --- |
| D-Bus round-trips/iter | 57 | 1 | 98.2% fewer |
| Round-trips total | 570000 | 10000 | 98.2% fewer |
| cpu_seconds | 1.37 | 1.35 | 1.8% faster |
| Modeled wall time | 115.37s | 3.35s | 97.1% faster |
| Speedup factor | 1.0x | 34.4x | 34.4x |

Benchmarks the row_setup_event/row_update_event hot path by modeling each device property access as one synchronous D-Bus round-trip. run_compare.sh diffs the current branch against a base ref (default main) and asserts a minimum gain.

Why the gain: every `device["Key"]` resolves to one synchronous `org.freedesktop.DBus.Properties.Get` via Gio call_sync (`blueman/bluez/Base.py` `get()`). main reads each property individually and re-runs row_update_event four times inside row_setup_event, so a single row costs ~57 round-trips. The branch calls `get_properties()` once (one `Properties.GetAll`) per row lifecycle, caches uuids/class/address/icon in the liststore and keeps them fresh from property-changed signals, so setup costs exactly one round trip and property updates cost zero. D-Bus round-trips dominate wall time, so collapsing them is the gain; pure-Python CPU is near-flat between versions.

## Tests

- `/usr/bin/python3 -m unittest test.gui.manager.test_manager_device_list test.gui.test_device_lists` (36 tests)
- Deterministic fuzz-style tests: a 2000-event randomized property state machine over `row_update_event` asserting the cached row state always matches a reference model and the device proxy is never read, plus randomized sweeps over `_has_objpush` and `make_display_name`.
- Changed-line coverage vs main (stdlib `trace`): ManagerDeviceList 94%, GenericList / DeviceList / DeviceSelectorList 100%. `ManagerDeviceMenu.generate` remains untested (needs a real `Gtk.Menu` and installed GSettings schemas), as before this PR.
- `test/benchmarks/run_compare.sh 10000 main` passes (numbers above).


infirit:
How much of a performance impact are you seeing? It may be negative as getting all the properties instead of a couple will result in more dbus calls in the end.

The best performance increase will be using the cached properties on the proxy. However last time I tried I ran into issues with the singleton design and most likely stale proxies.

I haven't looked too closely but you don't need to cast `NewType`, just call the newtypes like `BTAddress` on the value and type checkers will pick the correct type.

geraldo-netto:
Hi @infirit , let me add some proper benchmarks, as I mentioned to @cschramm , I was vibe coding these things and some of these come as suggestions from claude

geraldo-netto:
@infirit @cschramm 
I'm adding the benchmarks now and the microbenchmark says:
but let me finish the microbenchmark, so, I can add it too to avoid performance regressions

image

geraldo-netto:
@infirit , could you please review it now?
I also have added to the PR body the numbers and why it works and fixed the flake8 issue. I think we are good to go

infirit:
Apologies I should have looked closer. I forgot that GetProperties is a single dbus call :sweat:. I'll do some testing over the weekend.

geraldo-netto:
@infirit @cschramm I managed to find some more optimization opportunities, I have updated the PR

PR #3318, opened 2026-06-19
Harden blueman/main/NetConf.py: reliability, concurrency, portability, tests

geraldo-netto:
Works through the open `blueman/main/NetConf.py` findings. One commit per fix; coverage of `NetConf.py` rises to ~93% via `test/main/test_netconf.py`.

## Reliability & consistency
- **depend-1**: poll the DHCP pid file with bounded retry and treat a missing pid as a failed start (tear down) instead of locking an unsupervised daemon that leaks on pan1.
- **depend-2**: omit the `dns-server` dhcp-option when there are no servers, instead of emitting a dnsmasq-rejected trailing comma.
- **sm-7**: make `DHCPHandler.clean_up` idempotent — clear the cached pid before signalling, default the running-binary lookup (fixes a latent `StopIteration`), and swallow `ProcessLookupError`.
- **wd-7**: bound the daemon-start `communicate()` calls with a timeout and kill on expiry.
- **dist-2**: tag iptables rules with a `blueman-pan1` comment and reconcile against the live ruleset (`iptables -S`) instead of trusting in-memory state across mechanism restarts.
- **dist-1**: hold an exclusive `fcntl.flock` across apply/clean_up so concurrent D-Bus calls are serialized.
- **dist-4**: roll back the partially-applied state (full clean_up) if `apply_settings` fails partway.

## Portability
- **plat-3**: resolve `iptables` via PATH instead of hardcoding `/sbin/iptables`.
- **plat-4**: fail with a clear error when the IPv4 forwarding sysctl tree is absent.
- **plat-9**: abstract `/proc` access in `_is_running` with an `os.kill` liveness fallback.
- **cfg-3**: prefer `/run` and derive the pid path from a single `_RUN_PATH`.
- **mem-3**: replace udhcpd's fixed `sleep(0.1)` with the bounded pid-file poll.

## Observability
- **obs-2**: log DHCP daemon termination instead of `print()`.
- **obs-5**: log the swallowed `BridgeException` on cleanup.

## Tests
- `test/main/test_netconf.py` extended to ~45 tests; `NetConf.py` coverage ~93%.
- `mypy -p blueman --strict` clean; flake8 (core codes) clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -20,12 +20,26 @@ class NetworkSetupError(Exception):
     pass
 
 
-def _is_running(name: str, pid: int) -> bool:
-    path = pathlib.Path(f"/proc/{pid}")
-    if not path.exists():
-        return False
+_PROC_PATH = pathlib.Path("/proc")
 
-    return name in path.joinpath("cmdline").read_text()
+
+def _is_running(name: str, pid: int) -> bool:
+    if _PROC_PATH.is_dir():
+        path = _PROC_PATH / str(pid)
+        if not path.exists():
+            return False
+        try:
+            return name in path.joinpath("cmdline").read_text()
+        except OSError:
+            return False
+
+    # No procfs (non-Linux): we cannot match the binary name, so fall back to
+    # a plain liveness check via signal 0.
+    try:
+        os.kill(pid, 0)
+    except OSError:
+        return False
+    return True
 
 
 def _read_pid_file(fname: pathlib.Path) -> int | None:
--- a/blueman/main/NetConf.py
+++ b/blueman/main/NetConf.py
@@ -356,8 +356,8 @@ class NetConf:
 
         try:
             destroy_bridge("pan1")
-        except BridgeException:
-            pass
+        except BridgeException as e:
+            logging.warning(f"Failed to destroy bridge pan1 (errno {e.errno})")
         cls.unlock("netconfig")
 
         cls._del_ipt_rules()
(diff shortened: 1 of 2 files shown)


blueman/main/NetConf.py
```diff
@@ -294,70 +394,112 @@ def _validate_ipv4(ip4_address: str, ip4_mask: str) -> None:
...
+            raise NetworkSetupError(
+                "Neither ifconfig or ip commands are found. Please install net-tools or iproute2")
+
+    @classmethod
+    def _apply_iptables(cls, ip4_address: str, ip4_mask: str) -> None:
+        cls._del_ipt_rules()
+        cls._add_ipt_rule("nat", "POSTROUTING", "-s", f"{ip4_address}/{ip4_mask}", "-j", "MASQUERADE")
+        cls._add_ipt_rule("filter", "FORWARD", "-i", "pan1", "-j", "ACCEPT")
+        cls._add_ipt_rule("filter", "FORWARD", "-o", "pan1", "-j", "ACCEPT")
+        cls._add_ipt_rule("filter", "FORWARD", "-i", "pan1", "-j", "ACCEPT")
```
geraldo-netto, line 465: @infirit @cschramm 
`cls._add_ipt_rule("filter", "FORWARD", "-i", "pan1", "-j", "ACCEPT")`
is duplicate also in the main branch, see line 343
do you think it is a bug?

PR #3378, opened 2026-08-25
ManagerDeviceList: Expand connect/disconnect to more input methods

ricmatsui:
Handle row-activated instead of two button press to support touchscreen double tap and Enter/Space in addition to double click.

https://docs.gtk.org/gtk3/signal.TreeView.row-activated.html


infirit:
You can ignore the mypy errors, they are unrelated to your changes. I'll have to do a quick test but this looks ok to me, just a small nit on variable naming.

We have several path types, ObjectPath (str newtype for the dbus object path) and `Gtk.TreePath` and they can be easily confused. So please use `tree_path` when we're dealing with `Gtk.TreePath`.

ricmatsui:
Thanks for reviewing, I have made the change.

rokokol:
I hit the same bug on a touchscreen laptop and made an equivalent patch before I found this PR. Some test results, in case they help the review:

- A small GTK3 tree view under `WAYLAND_DEBUG=1` shows the cause. Each tap arrives as a single `GDK_BUTTON_PRESS`. GDK does not make `_2BUTTON_PRESS` from touch, so the current double-click branch does not run for a finger. The double tap arrives only as `row-activated`.
- With the same change applied to blueman 2.4.6 (Wayland, Hyprland), a double tap on a device connects it.
- The tree view counts two taps as a double tap only if they are in `gtk-double-click-distance`. The default is 5 px. For a finger, I set it to 24 px in `settings.ini`. This is a user setting and needs no change in this PR.

PR #3388, opened 2026-08-29
StatusNotifierItem: fix right click doing nothing on the tray icon in Waybar

EinBaum:
I use Waybar to display the blueman icon in the tray.
Waybar checks whether the dbusmenu declares a Version, but it didn't.
Then the context menu never opens.

Declare Version, TextDirection and Status. (they are mandatory in the spec)


blueman/main/indicators/StatusNotifierItem.py
```diff
@@ -13,7 +13,12 @@
 
 class MenuService(DbusService):
     def __init__(self, on_activate_menu_item: MenuItemActivator) -> None:
-        super().__init__(None, "com.canonical.dbusmenu", "/org/blueman/sni/menu", Gio.BusType.SESSION)
+        super().__init__(None, "com.canonical.dbusmenu", "/org/blueman/sni/menu", Gio.BusType.SESSION,
+                         properties={"Version": "u", "TextDirection": "s", "Status": "s"})
+
+        self.Version = 3
+        self.TextDirection = ""
```
cschramm, line 20: This is out-of-spec and a no-op in the reference client

blueman/main/indicators/StatusNotifierItem.py
```diff
@@ -13,7 +13,12 @@
...
+        self.Status = "normal"
```
cschramm, line 21: Pointless as well, as it's the default

blueman/main/indicators/StatusNotifierItem.py
```diff
@@ -13,7 +13,12 @@
 
 class MenuService(DbusService):
     def __init__(self, on_activate_menu_item: MenuItemActivator) -> None:
-        super().__init__(None, "com.canonical.dbusmenu", "/org/blueman/sni/menu", Gio.BusType.SESSION)
+        super().__init__(None, "com.canonical.dbusmenu", "/org/blueman/sni/menu", Gio.BusType.SESSION,
+                         properties={"Version": "u", "TextDirection": "s", "Status": "s"})
+
+        self.Version = 3
```
cschramm, line 19: Did you actually try this? It should give you a menu that does not work.

It would work with versions 1 or 2. The choice is mostly arbitrary, with 2 being the slightly better fit, which is why I described in #2599 that I'm not keen to advertise any arbitrary, unspecified, messy-library-based version.

Anyway, if waybar is eager to randomly check for the presence of the `Version` property, we probably need to

cschramm (commented):
> Waybar checks whether the dbusmenu declares a Version, but it didn't. Then the context menu never opens.

*sigh*... Note that waybar only added that recently and it's not in a release, so there is some hope. :see_no_evil: 

> Declare Version, TextDirection and Status. (they are mandatory in the spec)

What spec?

What we have for dbusmenus is basically the D-Bus introspection XML, which doesn't declare anything mandatory or optional, and the reference client's behavior, which treats all properties as optional.

EinBaum:
https://github.com/blueman-project/blueman/pull/2599 would fix it too. 
So this PR is not needed.

Issue #2477, opened 2024-08-31
No UI is presented for requests for confirmation


pvonmoradi:
blueman: 2.3.5-2
BlueZ: 5.66-1
Distribution: Debian 12
Desktop environment: i3wm (no DE), dunst notification daemon

According to [Wiki](https://github.com/blueman-project/blueman/wiki/Integrations#desktop-notifications), blueman defers the yes/no confirmation for various bluetooth services to the notification daemon. But what happens when the daemon doesn't support "buttons"? Since blueman already uses gtk, shouldn't it present the yes/no confirmation UI with a normal gtk dialog too (why it defers this task to third-party notification daemons)?

For example I had trouble pairing with an Android device since I couldn't confirm the code on desktop side. I had to resort to `bluetoothctl` then run `scan on` and `pair MAC_ADDRESS`. 

Same issue when receiving a file from the paired device (I have to `trust` the device and enable `Accept files from trusted devices` option to be able to receive anything at all)

infirit:
> blueman: 2.3.5-2 BlueZ: 5.66-1 Distribution: Debian 12 Desktop environment: i3wm (no DE), dunst notification daemon
> 
> According to [Wiki](https://github.com/blueman-project/blueman/wiki/Integrations#desktop-notifications), blueman defers the yes/no confirmation for various bluetooth services to the notification daemon. But what happens when the daemon doesn't support "buttons"?

Dunst should support [actions](https://dunst-project.org/documentation/#ACTIONS).

> Since blueman already uses gtk, shouldn't it present the yes/no confirmation UI with a normal gtk dialog too (why it defers this task to third-party notification daemons)?

Blueman has a Gtk Dialog fallback if there is no notification daemon available.

> For example I had trouble pairing with an Android device since I couldn't confirm the code on desktop side. I had to resort to `bluetoothctl` then run `scan on` and `pair MAC_ADDRESS`.

Post the blueman-applet log when you attempt to pair (some some other action that requires a notification). See https://github.com/blueman-project/blueman/wiki/Troubleshooting#debugging-blueman how to.

Most of the issues with dunst have been due some misconfigured/broken system. When we see the blueman-applet log we can see what happened.

hrideshmg:
If you, like me, find dunst actions to be quite clunky for this purpose (why should i have to dedicate a keybind to access a dialog box) then it is possible to tell blueman to avoid using the notification daemon, use the following dconf configuration:

`gsettings set org.blueman.general notification-daemon false`

Issue #1525, opened 2021-04-27
do_g_signal: The notification was dismissed by the user | device_confirm_passkey: Operation not permitted


b3wii:
When in bluetooth setting on my android phone and I tap on the found entry(my pc) to establish a connection, {'Connected': True} appears in the log but as soon as i accept the auth dialog it shows {'Connected': False}.

The notification was NOT dismissed by the user but the log shows this message. And also this message: device_confirm_passkey: Operation not permitted.

LOGS:

BluezAgent:262 _on_request_confirmation: Agent.RequestConfirmation
blueman-applet 15.12.51 INFO     Notification:239 do_g_signal: 2
blueman-applet 15.12.51 DEBUG    Notification:245 do_g_signal: The notification was dismissed by the user.
blueman-applet 15.13.25 INFO     BluezAgent:207 _on_cancel: Agent.Cancel

blueman-manager version 2.1.2 starting
blueman-applet 15.40.29 DEBUG    Base:74 do_g_properties_changed: /org/bluez/hci0/dev_xXx {'Connected': False}
blueman-applet 15.40.41 DEBUG    Base:74 do_g_properties_changed: /org/bluez/hci0/dev_xXx {'Connected': True}
blueman-applet 15.40.45 DEBUG    Base:74 do_g_properties_changed: /org/bluez/hci0/dev_xXx {'Connected': False}
blueman-applet 15.41.02 DEBUG    Base:74 do_g_properties_changed: /org/bluez/hci0/dev_xXx {'Connected': True}
blueman-applet 15.41.13 DEBUG    Base:74 do_g_properties_changed: /org/bluez/hci0/dev_xXx {'Connected': False}

Apr 27 15:40:45 bluetoothd[78076]: src/adapter.c:bonding_attempt_complete() hci0 bdaddr xXx type 0 status 0xe
Apr 27 15:40:45 bluetoothd[78076]: src/device.c:device_bonding_complete() bonding (nil) status 0x0e
Apr 27 15:40:45 bluetoothd[78076]: src/device.c:device_bonding_failed() status 14
Apr 27 15:40:45 bluetoothd[78076]: src/adapter.c:resume_discovery()
Apr 27 15:41:02 bluetoothd[78076]: src/adapter.c:connected_callback() hci0 device xXx connected eir_len 21
Apr 27 15:41:02 bluetoothd[78076]: src/adapter.c:user_confirm_request_callback() hci0 xXx confirm_hint 0
[...]

Issue #669, opened 2017-03-14
"Timed out waiting for reply from agent" when pairing with an iPhone


hobarrera:
I'm trying to pair my laptop with my iPhone, but I'm getting a `Timed out waiting for reply from agent` error. Here's what I'm doing:

* Make both devices discoverable (just in case really; one ought to do).
* Run `blueman-manager`
* Search for devices.
* Select the iPhone.
* Right click, select "Trust" (again, just in case, though I've tried without this step before)
* Right click, select "Pair"
* These both happen at the same time:
  * iPhone prompts "Bluetooth pairing request", with "Cancel" and "Pair".
  * Laptop shows a notification: "Pairing request for Hugo's iPhone, bla bla, MAC, bla CODE".
* Tap pair on the iPhone's popup.

Attempting to initiate from the iPhone (eg: clicking my laptop in its list) is basically the same popup (iPhone) and notification (laptop), with the same result.

Pairing codes match on both ends.

Here's what little logging I have:
```
$ sudo journalctl --follow -u bluetooth
Mar 14 18:56:37 athena.barrera.io systemd[1]: Starting Bluetooth service...
Mar 14 18:56:37 athena.barrera.io bluetoothd[1056]: Bluetooth daemon 5.44
Mar 14 18:56:37 athena.barrera.io bluetoothd[1056]: Starting SDP server
Mar 14 18:56:37 athena.barrera.io bluetoothd[1056]: Bluetooth management interface 1.14 initialized
Mar 14 18:56:37 athena.barrera.io systemd[1]: Started Bluetooth service.
Mar 14 18:56:44 athena.barrera.io bluetoothd[1056]: Endpoint registered: sender=:1.22 path=/MediaEndpoint/A2DPSource
Mar 14 18:56:44 athena.barrera.io bluetoothd[1056]: Endpoint registered: sender=:1.22 path=/MediaEndpoint/A2DPSink
Mar 14 19:01:52 athena.barrera.io bluetoothd[1056]: Timed out waiting for reply from agent
Mar 14 19:06:34 athena.barrera.io bluetoothd[1056]: Timed out waiting for reply from agent
```

Pair attempts are the last two lines (each attempts give an extra line).

Both devices pair up with my headphones fine, so bluetooth itself is working.

hobarrera:
Bump?

infirit:
[...]

PR #3393, opened 2026-08-30
BluezAgent: Keep pairing confirmations actionable

Th0masL:
## Summary

- add an opt-in persistent dialog path to the notification factory
- use it for BlueZ `RequestConfirmation` callbacks, which block pairing until the user responds
- preserve notification-daemon behavior for all other notifications
- add regression tests for both the factory selection and BlueZ agent call site

## Problem

`RequestConfirmation` currently uses a desktop notification when the notification
server advertises action support. An action-capable server may still hide its
banner before BlueZ's authentication request expires, leaving no visible way to
confirm pairing.

This was reproduced with an Apple Magic Keyboard on:

- Linux Mint 22.3, X11 Cinnamon
- Blueman `2.4.4+mint2+xia`
- BlueZ `5.72-0ubuntu5.5`
- Cinnamon `6.6.9+zena`
- Cinnamon notification duration: 4 seconds

Cinnamon advertised `actions` and `persistence`. Blueman therefore used a
notification bubble. The preceding `Connected` notification and pairing
notification were submitted 0.428 seconds apart and displayed sequentially.
The pairing notification became visible about 4.1 seconds later, remained
visible for 4.323 seconds, and then disappeared. BlueZ eventually sent
`Agent.Cancel`; the keyboard remained unpaired and unbonded.

## Fix

Add a `force_dialog` option to the notification factory and enable it only for
`RequestConfirmation`. The GTK dialog remains actionable until the user
confirms, denies, closes it, or BlueZ cancels the request.

## Manual verification

The installed Blueman Python files were backed up, patched with the exact diff
from this branch, and identified by SHA-256 during the second capture. The
package version and all other environment details remained unchanged.

| Result | Stock | Patched |
|---|---:|---:|
| Confirmation remained visible for at least 10 seconds | No | Yes |
| BlueZ sent `Agent.Cancel` | Yes | No |
| Paired | No | Yes |
| Bonded | No | Yes |
| Connected at capture completion | No | Yes |

The patched confirmation dialog was deliberately left untouched for 10 seconds
before clicking **Confirm**. Pairing then completed successfully.

## Automated tests

```text
python3 -m unittest -v test.gui.test_notification test.main.applet.test_bluez_agent

Ran 3 tests in 0.004s
OK
```

The full local suite ran 297 tests. All 293 tests supported by the local test
environment passed; four unrelated tests could not run because `python-dbusmock`,
a writable `/var/run`, and a test `/dev/rfkill` device were unavailable. CI
provides those prerequisites.

Related reports include #669, #1525, and #2477.


infirit:
This is a design decision by cinnamon to move/hide notifications to the applet after a fixed time. It's still there and actionable if you click on it in the tray.

Th0masL:
Hmm I see, that was far from clear to me that I could still access the notification somewhere.
I guess that means that you will not want to merge that change (which I understand).
Do you suggest I propose a fix on the cinnamon's side ?

infirit:
I don't see this being merged and I also doubt Cinnamon will change this behaviour.

Pairing/adding new devices should be done in a dedicated wizard, like blueman did before. Then you can show all the relevant details, choices etc in a single window. I wanted to bring it back for a while but if someone want to work on this that would be great.

Th0masL:
After investigating Cinnamon’s notification handling, I think there may be a smaller Blueman-side solution than introducing a persistent GTK dialog.

Blueman first displays a **“Connected”** notification for the temporary pairing connection. The actionable confirmation is queued behind it:

```text
Connection → confirmation request: 0.754 seconds
Connection → “Connected” marker: 0.739 seconds
“Connected” marker → six-digit marker: 4.127 seconds
Six-digit popup visible: 4.323 seconds
Outcome: authentication-cancelled
```

The non-actionable banner delays the confirmation by about four seconds. Cinnamon then shows the confirmation for only about four seconds before moving it into the notification tray.

A focused solution could be:

- Suppress **“Connected”** notifications for devices that are not yet paired or bonded. This message is misleading during pairing and delays the actionable notification.
- Give `RequestConfirmation` notifications critical/action-required urgency so notification daemons can keep them visible appropriately.

This should address the main usability problem with a small, standards-based Blueman change. I can revise and test the PR using this approach if the maintainers prefer it.

infirit:
You can disable the ConnectionNotifier plugin in the plugin-dialog. Does this make things work better?

edit: you can set the urgency hint on the Notification class, `set_hint("urgency", 2)`, iirc 2 is urgent.

Th0masL:
Yeah, but my point is to try to make the user experience easier out of the box for everybody without having to change settings.

I'm an IT guy and it still took me couple of minutes to understand why I couldn't get my keyboard pair easily.
So I'm assuming other people might have similar frustrating experience.

cschramm:
> * Suppress **“Connected”** notifications for devices that are not yet paired or bonded. This message is misleading during pairing and delays the actionable notification.

`Connected && !Paired` is a legitimate steady state, not just a pairing transient. Pairing is only required for services that demand authentication. Peripherals frequently never pair at all. They sit at `Connected=True, Paired=False` for their whole lifetime. Classic profiles that don't require authentication include receiving a file via OBEX push from an unpaired phone or various SPP adapters (OBD dongles etc.).

Also, suppressing would mean losing the notification as of now, as there is no other trigger than the `Connected` transition

PR #3429, opened 2026-09-16
Add per-device connection notification toggle

woodpile72:
Bluetooth mice and other HID devices can reconnect frequently as they sleep and wake. Currently ConnectionNotifier shows a desktop notification for every connect and disconnect event.

This adds a per-device toggle in blueman-manager's device context menu:

Show notifications

The setting defaults to enabled for every device. Turning it off stores the device address in a new GSettings string-array key and ConnectionNotifier skips connect/disconnect notifications for that address. Other devices continue to notify normally.

Checks run:

python3 -m py_compile blueman/plugins/applet/ConnectionNotifier.py blueman/plugins/manager/ConnectionNotifier.py blueman/gui/manager/ManagerDeviceMenu.py
glib-compile-schemas --strict --targetdir=/tmp/blueman-schema-check data
flake8 on the touched Python files
mypy --strict on the touched Python files
CYTHONEXEC=cython3 ./autogen.sh
make -j$(nproc)
make -C module
meson --warnlevel 3 --buildtype debug -Druntime_deps_check=false builddebug-pr3429
ninja -v -C builddebug-pr3429/
PYTHONPATH=module/.libs python3 -m unittest
git diff --check

The unittest run completed 299 tests.


infirit:
Havent looked to closely on the logic but here are some quick comments.

I think this should be a `ManagerPlugin`. Menus can then be handled by leveraging `MenuItemsProvider`. And `BasePlugin` has the ability to handle `GSettings` for you.

On the `cast` usage, you can call `BtAddress` and it will do the cast for you.

You can ignore the mypy error, it's unrelated to your changes.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -406,6 +422,15 @@ def generate(self) -> None:
             item.show()
         item.props.tooltip_text = _("Block/Unblock this device")
 
+        item = Gtk.CheckMenuItem.new_with_mnemonic(_("Show _notifications"))
```
infirit: Construct the CheckMenuItem by passing the right properties as keyword arguments like this.

```python
item = Gtk.CheckMenuItem(
    label=_("Show _notifications"),
    tooltip_text = _("Show notifications when this device connects or disconnects"),
    visible=True,
    use_underline=True
)
```

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -336,7 +352,7 @@ def generate(self) -> None:
         config = AutoConnectConfig()
         generic_service = ServiceUUID("00000000-0000-0000-0000-000000000000")
         object_path = self.SelectedDevice.get_object_path()
-        btaddress: BtAddress = self.SelectedDevice["Address"]
+        btaddress = cast("BtAddress", self.SelectedDevice["Address"])
```
infirit: Per my earlier comment, BtAddress is a NewType for str, it's callable and mypy will understand what type it is.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -138,6 +140,20 @@ def _on_service_property_changed(self, _service: AnyNetwork | AnyDevice, key: st
         if key == "Connected":
             self.generate()
 
+    def _connection_notifications_enabled(self, device: Device) -> bool:
+        return device["Address"] not in self._config["connection-notification-disabled-devices"]
+
+    def _set_connection_notifications_enabled(self, device: Device, enabled: bool) -> None:
+        address = device["Address"]
```
infirit: This will make mypy aware `address` is a `BtAddress` type.

```suggestion
        address = BtAddress(device["Address"])
```

data/org.blueman.gschema.xml
```diff
@@ -73,6 +73,11 @@
       <summary>Use notification daemon</summary>
       <description>If this is set to false blueman always uses its internal fallback notification dialog and does not invoke a notification daemon. Otherwise the fallback dialog will only be used if actions need to be displayed and the notification daemon does not report to support them.</description>
     </key>
+    <key type="as" name="connection-notification-disabled-devices">
```
infirit: I personally would prefer to put this in it's own schema, especially when it's made a `ManagerPlugin`. Then the setting can be given a much shorter name like `disabled-devices`.

blueman/services/Functions.py
```diff
@@ -1,5 +1,6 @@
 import inspect
```
infirit: pls drop the changes in this file . I have a PR open with a fix for this.

blueman/gui/manager/ManagerDeviceMenu.py
```diff
@@ -138,6 +140,20 @@ def _on_service_property_changed(self, _service: AnyNetwork | AnyDevice, key: st
...
+        disabled_devices = self._config["connection-notification-disabled-devices"]
```
infirit: When converted to a `ManagerPlugin` you can use `self.get_option` and `self.set_option`.

woodpile72:
Thanks, that makes sense.

I reworked this so the menu item is provided by a manager plugin using MenuItemsProvider instead of adding it directly to ManagerDeviceMenu.

I also changed the applet and manager plugins to use BasePlugin settings handling, switched the address handling to BtAddress(...), and dropped the unrelated mypy workaround from services/Functions.py.

Local checks are updated in the PR description. The unittest run is now 299 tests.

woodpile72:
Just a quick note: I'll be offline for a bit, so I wanted to check whether anything else is needed from me right now.

I reworked this based on the feedback above and tested it locally with a reconnecting Bluetooth mouse. Disabling notifications for that device suppresses the connect/disconnect notifications as expected.

